PR #21, opened 2023-01-04
RSDK-792 - Robot client service wrappers

stuqdog:
Creates robot client and service wrappers, and creates initial testing framework. More work needs to be done on testing (will be done in subsequent PR) but wanted to get this out there for visibility/to allow folks to play with as desired.


stuqdog:
@viamrobotics/sdk-netcode I've just added y'all for visibility here, no need to look at this/review it unless you're interested (except @njooma, you have to review it 😆 )

PR #26, opened 2023-01-09
Add docker compatibility

njooma:
Creates a docker file and updates the readme to allow people to develop on docker (and not have to jump through linking libraries). 

Also adds a new `clean` method in the makefile, and limits the protobuf we generate to exclude things we don't need (like app)


Makefile
```diff
@@ -8,10 +8,14 @@ CXXFLAGS += -std=c++11 -I src/gen/ -I src/gen/proto/
 
 all: buf 
 
-buf:	buf-clean
+buf: buf-clean
 	@mkdir -p src/gen
 	buf generate buf.build/viamrobotics/goutils --template buf.gen.yaml
 	buf generate buf.build/googleapis/googleapis --template buf.gen.yaml --path google/rpc --path google/api 
-	buf generate buf.build/viamrobotics/api --template buf.gen.yaml
+	buf generate buf.build/viamrobotics/api --template buf.gen.yaml --path common,component,robot,service
```
stuqdog, line 15: I think we should add `module` to the `--path` flag as well, we'll be relying on it for the modular resources wrappers.

stuqdog (commented):
A couple nits, otherwise lgtm!

Makefile
njooma: Yup! I don't think `module` was in when I made this change, so I'll add it in

stuqdog (approved):
👍

PR #28, opened 2023-01-11
RSDK-1669 - Module wrapper

stuqdog:


stuqdog:
A few notes to reviewers:

- Testing at this point is not comprehensive. I will be working with @Fahmina to do so, but wanted to not block people getting eyes on this.
- There are for sure code smells, anti-patterns, various badbadnogoods in here. My understanding is this is a very high priority PR (@Fahmina please correct me if I misunderstood you), so please blame my rush to get this PR out and general C++ newbishness for all shortcomings (and thank you @acmorrow for all your help on making what improvements there are!). I've created a jira ticket to track general improvements that need to be made and which will be my next project. You can see [here](https://viam.atlassian.net/browse/RSDK-1742). If you see things that are wrong then obviously please flag them, but given the high priority of this project, please note that anything that _can_ wait until the next PR likely will.
- This is a large PR. Sorry about that! Folks shouldn't feel that they need to look closely at all the code, or that they need to evaluate for everything. in particular:

1. @gvaradarajan I'd love for you to sanity check the spatial math logic, but feel no need to look beyond that
2. @Otterverse I'd like for you to confirm that the general module logic seems sane/there's nothing obvious that I'm missing, but feel no need to look beyond that
3. @acmorrow Added you for general C++ review, don't feel any need to pore over the actual substance of the logic
4. @njooma added you for general review/awareness, feel free to keep in mind what others are looking at/take more of a general high-level view.

stuqdog:
@Fahmina @cheukt tagging y'all for visibility :)

src/module/service.cpp
```diff
@@ -0,0 +1,245 @@
+#define BOOST_LOG_DYN_LINK 1
```
Otterverse, line 1: The file above this, src/module/module.cpp is being added as an empty file. Is that necessary for compiling or something? Or just an accidental add?

src/module/manager.cpp
```diff
@@ -0,0 +1,68 @@
+#include <module/manager.hpp>
+#include <module/service.hpp>
+#include <resource/resource.hpp>
+
+void ModuleManager::reconfigure_resource(Component cfg, std::vector<std::string> dependencies) {
```
Otterverse, line 5: same, module manager not needed in SDK

Otterverse (commented):
So, glad to see this getting added! Before I go more in depth reviewing things, I kept coming across one big question... lots of stuff is included here for the parent-side of the module system. That shouldn't be needed in the SDK at all in my estimation. The SDK is only needed to build modules themselves, not to actually build a full standalone server that also loads other modules, right? Is there a scope expansion I'm not aware of?

Makefile
```diff
@@ -10,9 +10,9 @@ all: buf
 
 buf: buf-clean
 	@mkdir -p src/gen
-	buf generate buf.build/viamrobotics/goutils --template buf.gen.yaml
-	buf generate buf.build/googleapis/googleapis --template buf.gen.yaml --path google/rpc --path google/api
-	buf generate buf.build/viamrobotics/api --template buf.gen.yaml --path common,component,robot,service,module
+	buf generate buf.build/viamrobotics/goutils --template buf.gen.yaml 
+	buf generate buf.build/googleapis/googleapis --template buf.gen.yaml --path google/rpc --path google/api 
+	buf generate buf.build/viamrobotics/api --template buf.gen.yaml --path common,component,robot,service,module,tagger,app
```
njooma, line 15: Just tried this in python -- you should be able to do `app/v1` as the path, so you only pull in what you need from app and not all the additional fleet stuff

src/module/service.cpp
stuqdog: Accidental add, thanks! Will remove.

include/config/module.hpp
```diff
@@ -0,0 +1,20 @@
+#pragma once
+
+#include <string>
+
+class ModuleConfig {
```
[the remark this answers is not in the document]
stuqdog: 🤦 you're totally right, I created that and never used it and forgot to delete it!

stuqdog:
> So, glad to see this getting added! Before I go more in depth reviewing things, I kept coming across one big question... lots of stuff is included here for the parent-side of the module system. That shouldn't be needed in the SDK at all in my estimation. The SDK is only needed to build modules themselves, not to actually build a full standalone server that also loads other modules, right? Is there a scope expansion I'm not aware of?

@Otterverse so my general mental model here is that we want the C++ sdk to be usable not just as a client connecting to the RDK, but as a server to program against. See, e.g., the robot server code. I'm not positive that module server functionality is strictly necessary at this time given who we expect to be using this and how, but I generally believe that it's something we will want long-term. please @cheukt correct any misunderstandings I have so I don't continue doing unnecessary/non-valuable work!

src/module/client.cpp
```diff
@@ -0,0 +1,92 @@
...
+#include <config/resource.hpp>
+#include <module/client.hpp>
+#include <module/handler_map.hpp>
+#include <resource/resource.hpp>
+#include <rpc/dial.hpp>
+
+using viam::module::v1::ModuleService;
+using Viam::SDK::ViamChannel;
+
+ModuleClient::ModuleClient(ViamChannel channel) {
```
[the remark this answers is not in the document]
stuqdog: I'm surprised to hear you say the client code shouldn't be needed, which makes me think I'm operating a false understanding somewhere. Given that this just wraps the basic functionality of updating a module, it seems like a pretty straightforward "this is what the SDK is for" case? what am I missing?

stuqdog:
@dannenberg for visibility

examples/dial/example_dial.cpp
```diff
@@ -0,0 +1,63 @@
...
+
+int main() {
+    const char* uri = "<your robot URI here>";
+    Viam::SDK::DialOptions dial_options = Viam::SDK::DialOptions();
+    std::string payload = "<your payload here>";
+    Credentials credentials(payload);
+    dial_options.credentials = credentials;
+    boost::optional<Viam::SDK::DialOptions> opts(dial_options);
+    std::string address(uri);
+    Options options = Options(1, opts);
```
acmorrow, line 35: Lots to say about this block. C++ doesn't require you to make so many named objects to achieve what you are actually doing here. I expect the following will work:

```
    std::string address = "<your robot URI here>";
    std::string payload = "<your payload here>";

    DialOptions dial_options;
    dial_options.credentials.emplace(std::move(payload));

    auto robot =
        RobotClient::at_address(
            std::move(address),
            {
                1,
                std::move(dial_options)
            }
        );
```

examples/dial/example_dial.cpp
```diff
@@ -0,0 +1,63 @@
...
+    std::shared_ptr<RobotClient> robot = RobotClient::at_address(address, options);
+    robot->refresh();
+    std::vector<ResourceName>* resource_names = robot->resource_names();
```
acmorrow, line 38: Ownership unclear. Is the caller responsible for freeing `resource_names`?

examples/dial/example_dial.cpp
```diff
@@ -0,0 +1,63 @@
...
+    ResourceName the_one_we_care_about = resource_names->at(0);
+    for (ResourceName resource : *resource_names) {
```
acmorrow, line 40: No need to say `ResourceName` here: `const auto& resource`  will do. There isn't much question what the type is.

examples/dial/example_dial.cpp
```diff
@@ -0,0 +1,63 @@
...
+        std::cout << "Resource name: " << resource.name() << resource.type() << resource.subtype()
+                  << std::endl;
```
acmorrow, line 42: Do you mean `\n` here? Using `endl` includes a flush.

examples/dial/example_dial.cpp
```diff
@@ -0,0 +1,63 @@
...
+    }
+    std::vector<Status> status_plural = robot->get_status();
```
acmorrow, line 44: `const auto status_plural` here.

examples/dial/example_dial.cpp
```diff
@@ -0,0 +1,63 @@
...
+    std::cout << "Status plural len " << status_plural.size() << std::endl;
+    for (Status s : status_plural) {
```
acmorrow, line 46: `const auto& s :` here too.

examples/dial/example_dial.cpp
```diff
@@ -0,0 +1,63 @@
...
+        std::cout << " Status! " << s.name().subtype() << std::endl;
+    }
+
+    std::vector<ResourceName> just_one = std::vector<ResourceName>();
```
acmorrow, line 50: You don't need to write so much to make a new empty vector. You could write this either as:

```
std::vector<ResourceName> just_one;
```

Or

```
auto just_one = std::vector<ResourceName>();
```

Thought the former is probably more standard.

include/common/proto_type.hpp
```diff
@@ -1,11 +1,10 @@
-#ifndef PROTOTYPE_H
-#define PROTOTYPE_H
+#pragma once
```
acmorrow, line 1: I think going to `pragma once` is fine for now, to eliminate the current include guards which lack sufficient uniqueness, but I'm in favor of considering moving to uuid-style guards:

```
#ifndef VIAM_CPP_SDK_HEADER_b8b6bfd7-5b15-4865-9061-d54a570de6ae_INCLUDED
#define VIAM_CPP_SDK_HEADER_b8b6bfd7-5b15-4865-9061-d54a570de6ae_INCLUDED
#endif // VIAM_CPP_SDK_HEADER_b8b6bfd7-5b15-4865-9061-d54a570de6ae_INCLUDED
```

Where each file gets a unique uuid. It would be pretty easy to script the conversion from `#pragma once` to that, so I'm fairly happy with unifying around `#pragma once` for now.

stuqdog:
@Otterverse @njooma @gvaradarajan @acmorrow sorry y'all, I think I have made some mistakes here 😬 changing this PR back to draft, will hopefully be re-open for review soon

examples/dial/example_dial.cpp
```diff
@@ -0,0 +1,63 @@
...
+    just_one.push_back(the_one_we_care_about);
+    std::vector<Status> status_singular = robot->get_status(just_one);
```
acmorrow, line 52: Similar comments re `status_singular`, `auto`, and the loop.

acmorrow (commented):
Holding off on the rest, per request, but sending the comments I had in flight.

stuqdog:
Hi folks sorry for all the delays! This is ready again for review. There are some shortcomings in the implementation; I'm deferring somewhat on fixing them until we add more wrappers and have a better sense of how to generically set everything up.

@Otterverse you are right I ended up throwing in a lot of parent-side code that is totally unnecessary. I'm leaving it in here under a philosophy of "it might be useful later, might as well keep it around." I think there's a valid argument for tossing it though, so happy to hear thoughts to the contrary.

@acmorrow thanks for all the C++ general comments. They're super helpful. I think I'd like to save any changes that aren't safety-critical for [RSDK-1742](https://viam.atlassian.net/browse/RSDK-1742). To the extent that it's not a hassle, can you mark which suggestions you think should be blocking on release here?

@gvaradarajan again, please only look at the spatialmath code. Also none of it is actually being used currently so it's not urgent at all :)

include/module/service.hpp
```diff
@@ -0,0 +1,45 @@
+#pragma once
+
+#include <component/generic/v1/generic.grpc.pb.h>
+#include <module/v1/module.grpc.pb.h>
+
+#include <components/service_base.hpp>
+// #include <generic/generic.hpp>
```
Otterverse, line 7: [nit] just remove if unused

Otterverse (approved):
So looking just at the module stuff, it looks okay to me, and I think it's okay to leave in the unused bits, now that you've put comments around them. I'll leave the rest of the reviewing to people with more c++ expertise.

acmorrow (commented):
@stuqdog - Comments from our discussion

acmorrow:
@stuqdog - Per our discussion this morning, please fold the `include` directory contents back in under the `src` directory (at least for now), and remove everything not required to make the new example module work. I think that will make it much easier for reviewers to understand the changes.

src/spatialmath/geometry.cpp
```diff
@@ -0,0 +1,91 @@
+#include <common/v1/common.pb.h>
```
stuqdog, line 1: @acmorrow not particularly important for you to review

src/spatialmath/geometry.hpp
```diff
@@ -0,0 +1,37 @@
+#pragma once
```
stuqdog, line 1: @acmorrow not particularly important for you to review

src/spatialmath/orientation.cpp
```diff
@@ -0,0 +1,136 @@
+#include <app/v1/robot.pb.h>
```
stuqdog, line 1: @acmorrow not particularly important for you to review

src/spatialmath/orientation.hpp
```diff
@@ -0,0 +1,26 @@
+#pragma once
```
stuqdog, line 1: @acmorrow not particularly important for you to review

src/spatialmath/orientation_types.cpp
```diff
@@ -0,0 +1,11 @@
+#include <app/v1/robot.pb.h>
```
stuqdog, line 1: @acmorrow not particularly important for you to review

src/spatialmath/orientation_types.hpp
```diff
@@ -0,0 +1,46 @@
+#pragma once
```
stuqdog, line 1: @acmorrow not particularly important for you to review

tests/test_robot.cpp
```diff
@@ -0,0 +1,313 @@
+#include <common/v1/common.pb.h>
```
stuqdog, line 1: just a code move; nothing new to review

src/referenceframe/frame.hpp
```diff
@@ -0,0 +1,20 @@
+#pragma once
```
stuqdog, line 1: @acmorrow not particularly important for you to review

src/common/proto_type.cpp
```diff
@@ -1,87 +1,47 @@
+#include <google/protobuf/struct.pb.h>
```
stuqdog, line 1: just enforcing cpp/hpp distinction and making a very small refactor. review not necessary.

src/common/utils.cpp
```diff
@@ -1,56 +1,41 @@
+#include <common/v1/common.pb.h>
```
stuqdog, line 1: cpp/hpp distinction enforcement, signature change. review not necessary.

src/components/component_base.hpp
```diff
@@ -1,33 +1,18 @@
-#ifndef COMPONENT_BASE_H
```
stuqdog, line 1: just removing `component_type`, which is now in a separate file. no meaningful changes for review.

src/robot/client.cpp
```diff
@@ -1,5 +1,7 @@
 #include <algorithm>
```
stuqdog, line 1: largely just enforcing cpp/hpp distinction and defining a couple trivial functions. not particularly important for review.

src/robot/service.cpp
```diff
@@ -1,48 +1,26 @@
+#include <common/v1/common.pb.h>
```
stuqdog, line 1: mostly just enforcing cpp/hpp distinction, plus adding a couple small `service` functions that already had a `component` analog. Not particularly important for review.

src/rpc/dial.hpp
```diff
@@ -1,16 +1,21 @@
-#ifndef DIAL_H
```
stuqdog, line 1: nothing meaningful for review here

src/module/service.cpp
```diff
@@ -0,0 +1,290 @@
...
+}
+
+void ModuleService_::add_api_from_registry(Subtype api) {
+    if (module->services.find(api) != module->services.end()) {
+        return;
+    }
+    module->lock.lock();
+    std::shared_ptr<SubtypeService> new_svc = SubtypeService::of_subtype(api.resource_subtype);
+
+    std::shared_ptr<ResourceSubtype> rs = Registry::lookup_subtype(api);
```
stuqdog, line 263: Note to reviewers: this isn't being used here (whoops!) but is being used in a subsequent ticket, so I'm leaving it here for now.

acmorrow (approved):
I'm approving this, despite the large number of comments.

Overall, this review is far too large, and isn't the right vehicle to address the systemic issues in the codebase.

I've tried to categorize my comments in ways that should make clear whether something is just guidance or style notes, or something that looks like a real problem. I'd be happy to do another round of review on changes only for the things that I've called out as real problems. However, I'm not confident in that as a next step, because I suspect that some of those fixes will result in what amount to small refactorings which will make this review even bigger.

acmorrow (approved):
I'm approving this, despite the large number of comments.

Overall, this review is far too large, and isn't the right vehicle to address the systemic issues in the codebase.

I've tried to categorize my comments in ways that should make clear whether something is just guidance or style notes, or something that looks like a real problem. I'd be happy to do another round of review on changes only for the things that I've called out as real problems. However, I'm not confident in that as a next step, because I suspect that some of those fixes will result in what amount to small refactorings which will make this review even bigger.

src/module/service.cpp
```diff
@@ -0,0 +1,290 @@
...
+        return grpc::Status(grpc::UNKNOWN, "unknown resource type " + cfg.api.resource_type);
+    }
+
+    if (module->services.find(cfg.api) == module->services.end()) {
+        module->lock.unlock();
+        return grpc::Status(grpc::UNKNOWN, "module cannot service api " + cfg.api.to_string());
+    }
+
+    std::shared_ptr<SubtypeService> sub_svc = module->services.at(cfg.api);
+    if (cfg.api.is_component_type() && !(cfg.api == Generic::subtype())) {
```
cheukt, line 93: given the DoCommand project, would just omit this entirely

src/module/service.cpp
```diff
@@ -0,0 +1,290 @@
...
+        if (reg != nullptr) {
+            std::shared_ptr<ComponentBase> comp =
+                reg->create_rpc_client(module->name, module->channel);
+            if (!(cfg.api == Generic::subtype())) {
+                if (module->services.find(Generic::subtype()) == module->services.end()) {
+                    return grpc::Status(grpc::UNKNOWN, "no generic service");
+                }
+                std::shared_ptr<SubtypeService> generic_service =
+                    module->services.at(Generic::subtype());
+                generic_service->replace_one(cfg.resource_name(), res);
```
cheukt, line 153: probably update this logic as well because we don't need generic service to serve everything now

cheukt (changes requested):
depending on scale of changes needed, I can see adding a few followup tickets instead of just fixing them all right now.

A significant one is that we need to use the config we send in through grpc - a good way to test this would be to update the example module to use it. Same for reconfigure.

Another note is expanding our components to include services in registry and manager instead of making a separate thing for services in the different places.

src/module/service.cpp
```diff
@@ -0,0 +1,290 @@
...
+    viam::app::v1::ComponentConfig proto = request->config();
+    Component cfg = Component::from_proto(proto);
+    std::shared_ptr<Module> module = this->module;
+    module->lock.lock();
+
+    std::shared_ptr<ResourceBase> res;
+    if (cfg.api.is_component_type()) {
+        std::shared_ptr<ComponentRegistration> reg = Registry::lookup_component(cfg.api, cfg.model);
+        if (reg != nullptr) {
+            res = reg->create_rpc_client(module->name, module->channel);
```
cheukt, line 75: we most likely need to use a different method here

src/registry/registry.hpp
```diff
@@ -1,26 +1,99 @@
...
+          create_rpc_client(std::move(create_rpc_client)){};
+    Subtype subtype;
+    Model model;
     ComponentType component_type;
+    std::function<std::shared_ptr<ComponentBase>(std::string, std::shared_ptr<grpc::Channel>)>
+        create_rpc_client;
+    viam::robot::v1::Status create_status(std::shared_ptr<ComponentBase> component);
+};
+
+class ServiceRegistration {
```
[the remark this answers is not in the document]
stuqdog: This is being done in a subsequent ticket!

src/components/resource_manager.cpp
```diff
@@ -1,56 +1,70 @@
...
-    // 	If the name is not within the ResourceManager or the registered
-    // component's type is not the expected type, then register_component
-    // will throw an error.
-    ComponentBase get_component(std::string name, ComponentType of_type);
-    ResourceManager(std::vector<ComponentBase> components);
-    ResourceManager();
-};
-
 ResourceManager::ResourceManager() {}
+std::unordered_map<std::string, std::shared_ptr<ServiceBase>> ResourceManager::services;
```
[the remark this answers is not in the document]
stuqdog: Agreed, will resolve in subsequent ticket.

src/module/service.cpp
```diff
@@ -0,0 +1,290 @@
...
+    }
+
+    return parent->resource_by_name(name.to_proto());
+}
+
+::grpc::Status ModuleService_::AddResource(::grpc::ServerContext* context,
+                                           const ::viam::module::v1::AddResourceRequest* request,
+                                           ::viam::module::v1::AddResourceResponse* response) {
+    viam::app::v1::ComponentConfig proto = request->config();
+    Component cfg = Component::from_proto(proto);
```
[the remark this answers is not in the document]
stuqdog: We're using it to find the registration that defines how to create the resource. per your below comment, we could certainly use a different method that takes the config directly instead.

src/module/service.cpp
stuqdog: handling in subsequent pr.

src/module/service.cpp
stuqdog: handling in subsequent pr.

src/resource/resource_base.hpp
```diff
@@ -0,0 +1,17 @@
...
+#include <resource/resource.hpp>
+#include <unordered_map>
+
+class ResourceBase;
+using Dependencies = std::unordered_map<Name, std::shared_ptr<ResourceBase>>;
+class ResourceBase : public grpc::Service {
+   public:
+    virtual grpc::StatusCode stop(std::unordered_map<std::string, ProtoType*> extra);
+    virtual grpc::StatusCode stop();
+    std::function<void(Dependencies)> reconfigure;
```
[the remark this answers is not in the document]
stuqdog: Will address in subsequent PR.

src/module/service.cpp
stuqdog: per above comment, will update to explicitly take `config` as an argument.

PR #31, opened 2023-01-18
CPP Quality of Life Improvements

njooma:
* Add doxygen and clang-format requirements to docker env
* Rename all cc and h files to cpp and cpp (including but generated ones, and added the rename command to the buf make command)
* Reformatting code using clang-format


acmorrow (approved):
LGTM

PR #40, opened 2023-02-16
RSDK-1858 - Remove old build system in favor of CMake and Docker

acmorrow:


BUILDING.md
```diff
@@ -61,6 +61,7 @@ CMake by default, but it can be obtained from the Bullseye Backports
 repository:
 
 ``` shell
+apt-get install software-properties-common
```
acmorrow, line 64: Thanks @dannenberg for catching this, it is required for the `apt-add-respository` command below.

CMakeLists.txt
```diff
@@ -78,16 +78,16 @@ option(VIAMCPPSDK_USE_DYNAMIC_PROTOS "Regenerate protos as part of building (net
 # Enforce known toolchains and toolchain minima unless asked not to.
 if (VIAMCPPSDK_ENFORCE_COMPILER_MINIMA)
   if ("${CMAKE_CXX_COMPILER_ID}" STREQUAL "GNU")
-    if (CMAKE_CXX_COMPILER_VERSION VERSION_LESS 10.2)
-      message(FATAL_ERROR "Insufficient GCC version: 10.2+ required")
+    if (CMAKE_CXX_COMPILER_VERSION VERSION_LESS 9.4)
```
acmorrow, line 81: Downgrade to the clang/gcc compiler versions as found on focal and the XCode version associated with that clang version.

CMakeLists.txt
```diff
@@ -220,7 +220,7 @@ add_custom_target(
 
 
 # Time to find `BOOST`.
-find_package(Boost 1.74 REQUIRED COMPONENTS headers log)
+find_package(Boost 1.71 REQUIRED COMPONENTS headers log)
```
acmorrow, line 223: Downgrade to focal's boost version.

CMakeLists.txt
```diff
@@ -237,23 +237,30 @@ find_package(Boost 1.74 REQUIRED COMPONENTS headers log)
 # gRPC version of 1.30.2, but don't `REQUIRE` since we want to have
 # another try via pkgconfig.
 
-find_package(gRPC 1.30.2 CONFIG QUIET)
+include(FindPkgConfig)
+
+find_package(protobuf 3.12.4 CONFIG)
```
acmorrow, line 242: When doing the build against the hand-built version of gRPC in the focal Docker image, I had a little trouble where it could find the gRPC `.cmake` files but those in turn didn't correctly load the protobuf .cmake files. As a guess, I tried pre-finding `protobuf` with `find_package` and that fixed it. I actually think this is just better than it was before, because before we were keying off the presence or absence of `gRPC` via `find_package` and assuming the same would be true of `protobuf`, but that's not necessarily true.

README.md
```diff
@@ -2,8 +2,9 @@
 
 ## (In)stability Notice
 
-> **Warning**
-> This is a alpha release of the Viam C++ SDK. Stability is not guaranteed. Breaking changes are likely to occur, and occur often.
+> **Warning** This is a alpha release of the Viam C++ SDK. Stability
```
acmorrow, line 5: I line wrapped this file, but it should be pretty easy to see the real changes.

README.md
```diff
@@ -12,56 +13,53 @@
...
+CMake-based build system for this project.
+
+## Using Docker for Development
+
+A Docker description is included which will install the required
+packages.
+
+1. Create a new docker image from the provided docker file. From the
+   root directory of this project (where this README is located), run
+   `docker build -t cpp .`
```
acmorrow, line 26: This needs a fixup since I moved the files.

src/robot/client.cpp
```diff
@@ -5,7 +5,6 @@
 #include <grpcpp/support/status.h>
 #include <robot/v1/robot.grpc.pb.h>
 #include <robot/v1/robot.pb.h>
-#include <rpc/rpc.h>
```
acmorrow, line 8: I'm not sure why this include was here, as it seems to be part of the old Sun RPC system. Not all systems seem to offer it in the usual places or default packages, so if we can get away without it I think that would be best.

acmorrow (commented):
@stuqdog - This is a grab bag of changes related to work I've been doing with @dannenberg to get this working for him on Focal, which is an older baseline than I'd previously assumed (Bullseye). So there is some downgrading of compiler/package requirements.

It also rips out the old Makefile based system and updates the documentation, and adds Dockerfiles for the relevant platforms.

src/robot/client.cpp
stuqdog: Just tested to confirm, this is unnecessary. Most likely just an artifact of old testing. Please feel free to just delete it, or I can do so in a future PR!

acmorrow: I'll just drop it out in this PR since I already have the change.

README.md
acmorrow: Updated

acmorrow:
@stuqdog - OK, I'm not sure what I was having trouble with before, but I went back to basically the prior setup for finding the grpc/protobuf packages and now it seems to be working the way I wanted. I think I just confused myself somewhere along the way. Meanwhile, I tightened up some of the logic in there.

I think this is ready for another look and can hopefully merge soon if you are happy with finalizing the jump to cmake.

stuqdog (approved):

PR #42, opened 2023-02-23
(RSDK-2067) - fix robot_client noisy errors and thread closing

stuqdog:
Leaving in draft for now as this has the `module_wrappers` pr as a base, which isn't yet merged. once that feature is done, I'll open this for review, at which point the diff will drop from the very large number now to something very small and manageable.


src/robot/client.cpp
```diff
@@ -73,7 +77,7 @@ std::vector<Status> RobotClient::get_status(std::vector<ResourceName> components
     }
 
     grpc::Status response = stub_->GetStatus(&ctx, req, &resp);
-    if (!response.ok()) {
+    if (!response.ok() && response.error_message() != k_stream_removed) {
```
cheukt, line 80: would refactor out to a function and just use that everywhere

cheukt (approved):

PR #43, opened 2023-02-28
address module PR comments

stuqdog:
Addressing comments from @cheukt as seen here: https://github.com/viamrobotics/viam-cpp-sdk/pull/28


cheukt (changes requested):
would add a test to make sure reconfiguration works

cheukt (approved):

PR #44, opened 2023-02-28
Build with shared libraries by default

acmorrow:
@stuqdog - No rush on this. It flips the default to building shared libraries and adds some other SDK-ification stuff. The only remaining wrinkle that I'm aware of is that `viam-rust-utils.dylib` has an incorrect install name per https://viam.atlassian.net/browse/RSDK-1996. That only affects macOS builds.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -13,8 +13,7 @@
 # limitations under the License.
 
 
-# TODO: [required] libviam_rust_utils handling
-# TODO: [required] Runtime paths and make BUILD_SHARED_LIBS work
+# TODO: [required] fixup libviam_rust_utils install name after download
 # TODO: Sanitizer build configs
 # TODO: All warnings and warnings as errors
 # TODO: Runtime dependency sets?
@@ -43,6 +42,15 @@ project(viam-cpp-sdk
   HOMEPAGE_URL https://github.com/viamrobotics/viam-cpp-sdk
   LANGUAGES CXX
 )
+
+
+# Configure cmake-level options:
+#
+# - `BUILD_SHARED_LIBS`: Enabled by default so that we produce a
+# modern SDK, this option can be set to `OFF` to build a static
+# library. Note that this may result in non-functional examples.
+#
+option(BUILD_SHARED_LIBS "If enabled, build shared libraries" ON)
 
 
 # Configure project-global options:
@@ -109,11 +117,6 @@ set(CMAKE_CXX_STANDARD_REQUIRED True)
 set(CMAKE_CXX_EXTENSIONS OFF)
 
 
-# Everything needs threads, and prefer -pthread if available
-set(THREADS_PREFER_PTHREAD_FLAG ON)
-find_package(Threads REQUIRED)
-
-
 # Produce a compilation database when generating for a build
 # system that is able to make one.
 set(CMAKE_EXPORT_COMPILE_COMMANDS ON)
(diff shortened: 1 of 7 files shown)


src/CMakeLists.txt
```diff
@@ -212,14 +218,15 @@ target_include_directories(viamcpp
...
 
-# This shouldn't be necessary, but we need it so that the installation
-# path of the C++ SDK is added to the link line of consumers to that
-# `-lviam_rust_utils` can be meaningfully resolved.
-#
-# TODO: Remove this when we can import `viam_rust_utils` as a real
-# imported target with its own properties.
 target_link_directories(viamcpp
-  INTERFACE
+  PUBLIC
```
acmorrow, line 222: This was actually a build system bug! It took me quite a while to figure out why `libviamcpp.so` wasn't ending up with an RPATH that included the paths of its dependent runtime libraries. But I had marked those link diretories as 'interface', meaning that only things which depended on `libviamcpp.so` got them, but not `libviamcpp.so` itself. By switching to `public`, now `libviamcpp.so` gets them too, as it should.

src/CMakeLists.txt
```diff
@@ -246,16 +253,20 @@ target_link_libraries(viamcpp
 install(
   TARGETS viamcpp
   EXPORT LibTargets
-  LIBRARY COMPONENT runtime
-  FILE_SET viamcpp_public_includes COMPONENT dev
-  FILE_SET viamcpp_public_pb_includes COMPONENT dev
+  RUNTIME COMPONENT viam-cpp-sdk_runtime
```
acmorrow, line 256: Changes here are about tagging the new soname symlink and versioned library into the right components, since you don't need the dev symlink in the runtime package.

src/CMakeLists.txt
```diff
@@ -246,16 +253,20 @@ target_link_libraries(viamcpp
...
+  FILE_SET viamcpp_public_includes COMPONENT viam-cpp-sdk_dev
+  FILE_SET viamcpp_public_pb_includes COMPONENT viam-cpp-sdk_dev
 )
 
 install(
   EXPORT LibTargets
   FILE viam-cpp-sdkLibTargets.cmake
   NAMESPACE viam-cpp-sdk::
   DESTINATION ${CMAKE_INSTALL_LIBDIR}/cmake/viam-cpp-sdk
+  COMPONENT viam-cpp-sdk_dev
```
acmorrow, line 269: Drive-by, this should always have been part of the `dev` component.

src/CMakeLists.txt
```diff
@@ -277,5 +288,5 @@ configure_file(
 install(FILES
   ${CMAKE_CURRENT_BINARY_DIR}/config/viam-cpp-sdk-libviamcpp.pc
   DESTINATION ${CMAKE_INSTALL_LIBDIR}/pkgconfig
-  COMPONENT dev
+  COMPONENT viam-cpp-sdk_dev
```
acmorrow, line 291: Ditto

CMakeLists.txt
```diff
@@ -126,10 +129,38 @@ set(CMAKE_EXPORT_COMPILE_COMMANDS ON)
...
 if (CMAKE_INSTALL_PREFIX_INITIALIZED_TO_DEFAULT)
   set(CMAKE_INSTALL_PREFIX "${CMAKE_BINARY_DIR}/install" CACHE PATH "..." FORCE)
 endif()
+include(GNUInstallDirs)
+
+
+# Configure how the SDK will manage runtime paths. This is important
+# even for BUILD_SHARED_LIBS=OFF due to the always-shared
+# `libviam_rust_utils` shared library.
+set(CMAKE_BUILD_WITH_INSTALL_RPATH false)
```
acmorrow, line 141: These five flags basically set things up so that we have valid rpaths in both the build tree and the install tree, and that the installation tree is relocatable (e.g. uses $ORIGIN or @loader_path / @rpath).

CMakeLists.txt
```diff
@@ -126,10 +129,38 @@ set(CMAKE_EXPORT_COMPILE_COMMANDS ON)
...
+set(CMAKE_BUILD_RPATH_USE_ORIGIN true)
+set(CMAKE_INSTALL_RPATH_USE_LINK_PATH true)
+set(CMAKE_MACOSX_RPATH true)
+set(CMAKE_SKIP_BUILD_RPATH false)
+
+list(FIND CMAKE_PLATFORM_IMPLICIT_LINK_DIRECTORIES "${CMAKE_INSTALL_PREFIX}/${CMAKE_INSTALL_LIBDIR}" isSystemDir)
```
acmorrow, line 147: This section then actually sets the correct install name and RPATH values based on the layout of the installation directory.

CMakeLists.txt
```diff
@@ -126,10 +129,38 @@ set(CMAKE_EXPORT_COMPILE_COMMANDS ON)
...
+  if (CMAKE_SYSTEM_NAME STREQUAL "Darwin")
+    set(CMAKE_INSTALL_NAME_DIR "@rpath/${PREFIXTOLIBRELPATH}")
+    list(PREPEND CMAKE_INSTALL_RPATH "@loader_path/${BINTOPREFIXRELPATH}")
+  else()
+    list(PREPEND CMAKE_INSTALL_RPATH "$ORIGIN/${BINTOPREFIXRELPATH}/${PREFIXTOLIBRELPATH}")
+  endif()
+endif()
+
+
+# Everything needs threads, and prefer -pthread if available
```
acmorrow, line 161: I felt like we were doing this too early, before we had set up the shape of the environment fully. I doubt it makes a difference, but I like it down here better.

CMakeLists.txt
```diff
@@ -190,15 +221,15 @@ set_property(TARGET viam_rust_utils PROPERTY IMPORTED_LOCATION ${viam_rust_utils
 
 install(
   IMPORTED_RUNTIME_ARTIFACTS viam_rust_utils
-  LIBRARY COMPONENT runtime
+  LIBRARY COMPONENT viam-cpp-sdk_runtime
```
acmorrow, line 224: Made this change here and elsewhere based on some suggestions I saw online to namespace the components.

src/CMakeLists.txt
```diff
@@ -196,6 +196,12 @@ target_sources(viamcpp
       ${PROTO_GEN_DIR}/gen/robot/v1/robot.pb.h
 )
 
+set_target_properties(
```
acmorrow, line 199: This gets us a soname containing the ABI version and an on-disk runtime file containing the full version, then a dev symlink.

stuqdog (approved):

PR #45, opened 2023-03-01
RSDK-2131 - Refactor generic component, add generic wrappers

stuqdog:


src/module/service.cpp
```diff
@@ -24,7 +24,7 @@
 #include <boost/none.hpp>
 #include <common/utils.hpp>
 #include <components/component_base.hpp>
-#include <components/generic.hpp>
+#include <components/generic/generic.hpp>
```
cheukt, line 27: do we still need the generic import here?

stuqdog: nope! thanks for spotting

cheukt (approved):
lgtm, tho it does seem like some of the includes can be cleaned up

stuqdog:
> lgtm, tho it does seem like some of the includes can be cleaned up

yeah, for sure that's the case :) I'm kinda lumping that in with the general refactoring work that needs to come, but will be cleaning it all up soon!

PR #46, opened 2023-03-03
RSDK-2069 - update cpp proto gen workflow

stuqdog:


stuqdog:
@oliviamiller this should fix the issue with the outdated protos you were encountering!

.github/workflows/update_protos.yml
```diff
@@ -21,7 +21,7 @@ jobs:
         with:
           repo-token: ${{ secrets.GITHUB_TOKEN }}
       - name: Generate buf
-        run: make buf
+        run: mkdir build && cd build && cmake .. -G Ninja -DVIAMCPPSDK_USE_DYNAMIC_PROTOS=ON && ninja update-static-protos
```
maximpertsov, line 24: [nit] it feels like this should be a encapsulated in `make` (or `cmake` ?) command

stuqdog: Hmm... I don't think we can easily put all of this in `cmake` since `cmake` is just a way to generate build files that `ninja` is using. As for putting it in a `make` command, I'm a little hesitant just since it's a bit opinionated about the state of the `build` dir (which is part of our `.gitignore` and which we expect users to have). I'm admittedly not an expert on this side of things though, so happy to chat more and be convinced otherwise!

stuqdog: Ahh great, thanks! updated :)

acmorrow (approved):
LGTM!

.github/workflows/update_protos.yml
maximpertsov: makes sense - I’m okay leaving as you have it 👌

PR #48, opened 2023-03-10
RSDK-2299 - Add boost test

stuqdog:
Adds boost.test support. tests should automatically be built when compiling with cmake, and can be run by invoking `ctest` from the `build` directory.


tests/test_generic_component.cpp
```diff
@@ -0,0 +1,16 @@
+#define BOOST_TEST_MODULE test module test_generic_component
+#include <boost/test/included/unit_test.hpp>
+
+BOOST_AUTO_TEST_SUITE(generic_suite)
+
+BOOST_AUTO_TEST_CASE(test1) {
```
stuqdog, line 6: Note to reviewers: these are just dummy tests to verify that our testing works. Actual, meaningful generic component tests will be added in a follow-up PR.

CMakeLists.txt
```diff
@@ -329,6 +329,7 @@ endif()
 # Pull in our subdirectories
 add_subdirectory(src)
 add_subdirectory(examples)
+add_subdirectory(tests)
```
acmorrow, line 332: I think you should push the `tests` directory under `src`. It makes some sense for the `examples` code to live at the root, since in some way those programs are clients of the library, not part of it. But I'd consider the tests to be "part of" the SDK. Fewer top level directories containing source code is generally a good idea.

CMakeLists.txt
```diff
@@ -376,3 +377,32 @@ add_custom_target(
   COMMAND ${CMAKE_COMMAND} -DCOMPONENT=viam-cpp-sdk_examples -P ${CMAKE_BINARY_DIR}/cmake_install.cmake
   DEPENDS all
 )
+
+add_custom_target(
+  install-tests
```
acmorrow, line 382: There is a decision to be made here. Should tests be installed / installable? If yes, should they be run from the installation directory or from the build directory?

Overall, I'm not sure it is particularly useful to install the tests. And I think it is harder to make the test execution targets make use of the installed tests, and easier to make it use the test binaries out of the build tree (I think that is what is happening now).

CMakeLists.txt
```diff
@@ -376,3 +377,32 @@ add_custom_target(
...
+  COMMAND ${CMAKE_COMMAND} -DCOMPONENT=viam-cpp-sdk_tests -P ${CMAKE_BINARY_DIR}/cmake_install.cmake
+  DEPENDS all
+)
+
+enable_testing()
```
acmorrow, line 387: By having `enable_testing` up here you are enabling it for the entire tree, not just the tests under the `test` directory, per the docs. I don't know what those effects are, but in general I'd recommend scoping effects as narrowly as possible. If you take my suggestion to move the `test` directory under `src`, then this `enable_testing` call would belong in `src/test/CMakeLists.txt`

CMakeLists.txt
```diff
@@ -376,3 +377,32 @@ add_custom_target(
...
+function(add_boost_test SOURCE_FILE_NAME)
```
acmorrow, line 388: I recommend to create a home for the inevitable accumulation of helper functions that this project will have over time rather than placing them in the top level. Make a new top-level directory called `cmake`, and add a file called `viamcpp_utils.cmake`. Then put this function in it. Then up in the top-level `CMakeLists.txt` you can add the `cmake` directory to the module search path:
```
set(CMAKE_MODULE_PATH
   ${CMAKE_MODULE_PATH}
   ${PROJECT_SOURCE_DIR}/cmake
)
```

Then you can `import(viamcppsdk_utils)` to pick up the function.

Also, rename the function to something like `viamcppsdk_add_boost_test` to namespace it.

CMakeLists.txt
```diff
@@ -376,3 +377,32 @@ add_custom_target(
...
+  string(
+    REGEX MATCHALL "BOOST_AUTO_TEST_CASE\\( *([A-Za-z_0-9]+) *\\)"
+    FOUND_TESTS ${SOURCE_FILE_CONTENTS})
+  
+  foreach(HIT ${FOUND_TESTS})
+    string(REGEX REPLACE ".*\\( *([A-Za-z_0-9]+) *\\).*" "\\1" TEST_NAME ${HIT})
+
+	add_test(
+      NAME "${TEST_EXECUTABLE_NAME}.${TEST_NAME}"
+	  COMMAND ${TEST_EXECUTABLE_NAME} --catch_system_error=yes)
```
acmorrow, line 404: Indentation looks off here, spaces vs tabs?

CMakeLists.txt
```diff
@@ -376,3 +377,32 @@ add_custom_target(
...
+  endforeach()
+endfunction()
+
+add_boost_test(tests/test_generic_component.cpp)
```
acmorrow, line 408: This belongs down in `tests/CMakeLists.txt`

tests/CMakeLists.txt
```diff
@@ -0,0 +1,26 @@
...
+
+add_executable(viam_cpp_sdk_tests
+  test_generic_component.cpp
+)
+
+target_link_libraries(viam_cpp_sdk_tests
+  viamcpp
+)
+
+install(
```
acmorrow, line 23: Only necessary if you think installing the tests is useful and/or you want to test the installed test binaries rather than the ones from the build tree.

tests/test_generic_component.cpp
acmorrow: I think that as a step before adding meaningful tests that the next step would be to start running these trivial tests in CI. Maybe in this PR?

acmorrow (changes requested):
This looks great! I have some suggested improvements.

tests/test_generic_component.cpp
stuqdog: I've got an update to the `fix proto generation` PR in the works that will have to be done before we can get tests running in CI. The short of it is: `boost` doesn't exist in CI by default, and when I added it via `MarkusJx/install-boost`, we still hit problems of CI not knowing where to find it. I'm very happy to defer merging of this PR until that is fixed though, and then make these tests run in CI here :)

acmorrow: Either way is fine, I'm happy to defer it to a later PR if there are obstacles to doing it here, and it sounds like there are.

stuqdog: Ack, given the existing hurdles I think I'll just merge this when ready, and work on getting CI fixed up in a subsequent PR.

src/components/generic/client.cpp
```diff
@@ -6,8 +6,8 @@
 #include "component/generic/v1/generic.grpc.pb.h"
 
 ProtoType GenericClient::do_command(std::unordered_map<std::string, ProtoType*> command) {
-    viam::component::generic::v1::DoCommandRequest req;
```
stuqdog, line 9: Note to reviewers: this change (and analogous changes) are unrelated to the core PR, but were necessary to match updates to our protos.

src/tests/CMakeLists.txt
```diff
@@ -0,0 +1,26 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+enable_testing()
+include(viamcppsdk_utils)
+viamcppsdk_add_boost_test(test_generic_component.cpp)
+
+add_executable(viam_cpp_sdk_tests
```
acmorrow, line 19: Do you still need this line? It looks to me like `viamcppsdk_add_boost_test` will call `add_executable` and `target_link_libararies` for you.

CMakeLists.txt
```diff
@@ -134,6 +134,15 @@ if (CMAKE_INSTALL_PREFIX_INITIALIZED_TO_DEFAULT)
...
 include(GNUInstallDirs)
 
+# Set the Cmake module path to include our cmake helper functions
+set(CMAKE_MODULE_PATH
+  ${CMAKE_MODULE_PATH}
+  ${PROJECT_SOURCE_DIR}/cmake
+)
+
+# Include our helper CMake functions
+include(viamcppsdk_utils)
```
acmorrow, line 144: I don't think you actually need this included here.

acmorrow (approved):
LGTM!

src/tests/CMakeLists.txt
stuqdog: quite right! fixed

PR #50, opened 2023-03-16
Random code review

stuqdog:


.github/CODEOWNERS
```diff
@@ -1 +1,2 @@
+* @viamrobotics/sdk-netcode
 * @stuqdog
```
maximpertsov, line 2: do you still want to keep yourself as a code-owner?

stuqdog: good question! probably not necessary, I'll update :)

maximpertsov: [opt] separately I was curious why you didn't automatically get added a code reviewer while you were an individual code owner - does the action randomize from the full pool of code-owners?

.github/workflows/set_reviewers.yml
```diff
@@ -0,0 +1,22 @@
...
+    container:
+      image: ghcr.io/viamrobotics/canon:amd64
+    steps:
+      - uses: actions/checkout@v3
+      - uses: pleo-io/assign-random-codeowners@v0.2.8
+        with:
+          reviewers-to-assign: 2
+          assign-individuals-from-teams: true
+        env:
+          GITHUB_TOKEN: ${{ secrets.GIT_ACCESS_TOKEN }}
```
maximpertsov, line 22: if you are referencing the automatically-created token that is created with each workflow run, then I think you need to use:

```suggestion
          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
```

.github/CODEOWNERS
stuqdog: I don't think I can be added as a reviewer to my own PR?

.github/workflows/set_reviewers.yml
stuqdog: this is a different token! you can see it in the settings under `Secrets and variables`/`actions` :) 

Although using `secrets.GIT_ACCESS_TOKEN` was sufficient to make the job run for `rust-sdk`, but doesn't seem to be working here... I wonder if it's because this isn't actually merged into main yet?

.github/workflows/set_reviewers.yml
```diff
@@ -0,0 +1,22 @@
+name: Set Reviewers
+
+on:
+  workflow_dispatch:
```
maximpertsov, line 4: RE: https://github.com/viamrobotics/viam-cpp-sdk/pull/50/files#r1139082301

does this change the situation at all?

```suggestion
  workflow_dispatch:
  workflow_call:
    secrets:
        GIT_ACCESS_TOKEN:
          required: true
```

stuqdog: does not, sadly :(

stuqdog:
@maximpertsov per convo with Eric in slack, I think this is just not going to work during this PR since it's trying to grab the secret from my account and necessarily failing, but it should work once it's merged.

PR #53, opened 2023-03-20
RSDK-2086 Camera Wrapper

oliviamiller:


cheukt:
seems like a lot of generated files got changed - is that expected?

Fahmina:
Just commenting so I can follow this PR. @shawnbmccarthy and I are ready to try to use this wrapper + modules to implement an OAKD camera

src/config/resource.hpp
```diff
@@ -9,7 +9,7 @@
 #include <string>
 #include <unordered_map>
 
-typedef std::unordered_map<std::string, ProtoType*> AttributeMap;
+typedef std::shared_ptr<std::unordered_map<std::string, std::shared_ptr<ProtoType>>> AttributeMap;
```
stuqdog, line 12: There are a lot of places where we use this exact signature, instead of the `typedef` here. Might make sense to try and replace those longer, annoying signatures with `AttributeMap`. 

Note that this _might_ get you into some circular dependency issues that might be pretty annoying. If that happens then feel free to just skip!

src/components/camera/camera.hpp
```diff
@@ -0,0 +1,145 @@
...
+            return mime_type == rhs.mime_type 
+            and bytes == rhs.bytes;
+        }
+    }; 
+
+    std::string lazy_suffix = "+lazy";
+
+    static std::shared_ptr<ResourceSubtype> resource_subtype();
+    static Subtype subtype();
+    virtual std::shared_ptr<std::unordered_map<std::string, std::shared_ptr<ProtoType>>> do_command(std::string name, std::shared_ptr<std::unordered_map<std::string, std::shared_ptr<ProtoType>>> command) = 0;
```
cheukt, line 135: the virtual methods should not have name as one of the args

cheukt (changes requested):
good work! can't really read C++ but added some general thoughts

src/components/resource_manager.hpp
stuqdog, line 1: I think this file got added back in during some of that github rebase stuff we were dealing with :) We can just delete it, unless you touched it at some point.

src/components/camera/server.hpp
```diff
@@ -0,0 +1,41 @@
...
+
+    void register_server() override;
+
+    CameraServer(){};
+    CameraServer(std::shared_ptr<SubtypeService> sub_svc) : sub_svc(sub_svc){};
+    std::shared_ptr<SubtypeService> sub_svc;
+
+
+   private:
+    static bool initialized;
```
stuqdog, line 37: looks like this is copied over from the `initialized` value in `generic/server.hpp`. That value is an artifact and doesn't do anything; it should be deleted here and there. Sorry about the confusion!

src/components/camera/server.hpp
```diff
@@ -0,0 +1,41 @@
...
+};
+
+  std::string bytes_to_string(std::vector<unsigned char> const &b);
```
stuqdog, line 40: similar to in the `client` code, there's nothing particularly `camera`-y about this function. It should probably live in `utils.cpp` if we need to expose it. Though again, since it's such a simple function, I'm not sure that adding it as a helper function adds much value.

src/components/camera/server.cpp
```diff
@@ -0,0 +1,162 @@
...
+    const ::viam::common::v1::DoCommandRequest* request,
+    ::viam::common::v1::DoCommandResponse* response) {
+
+    std::shared_ptr<ResourceBase> rb = sub_svc->resource(request->name());
+    if (rb == nullptr) {
+        return grpc::Status(grpc::UNKNOWN, "resource not found: " + request->name());
+    }
+
+    std::shared_ptr<Camera> camera = std::dynamic_pointer_cast<Camera>(rb);
+    std::shared_ptr<std::unordered_map<std::string, std::shared_ptr<ProtoType>>> command = struct_to_map(request->command());
```
stuqdog, line 17: (nit) I think we can get away with not defining this `command` here at all and in the below line just call `camera->do_command(request->name(), request->command());`

src/components/camera/server.cpp
```diff
@@ -0,0 +1,162 @@
...
+     std::shared_ptr<ResourceBase> rb = sub_svc->resource(request->name());
+    if (rb == nullptr) {
+        return grpc::Status(grpc::UNKNOWN, "resource not found: " + request->name());
+    } 
+
+   std::shared_ptr<Camera> camera =std::dynamic_pointer_cast<Camera>(rb);
+
+    std::string name = request->name();
+    std::string mime_type = request->mime_type();
+    Camera::raw_image image = camera->get_image(name, mime_type);
```
stuqdog, line 47: (nit) these three lines can be combines. the type information of `name` and `mime_type` aren't particularly interesting, so just calling `camera->get_image(request->name(), request->mime_type())` should be sufficient :)

src/components/camera/server.cpp
```diff
@@ -0,0 +1,162 @@
+#include <components/camera/server.hpp>
+#include <components/camera/camera.hpp>
+#include <rpc/server.hpp>
+#include <gen/google/api/http.pb.h>
+
+::grpc::Status CameraServer::DoCommand(
+    ::grpc::ServerContext* context,
+    const ::viam::common::v1::DoCommandRequest* request,
+    ::viam::common::v1::DoCommandResponse* response) {
```
stuqdog, line 9: I noticed that in other methods, we check that the `request` is non-null and return an error otherwise. Can we do the same here?

src/components/camera/server.cpp
```diff
@@ -0,0 +1,162 @@
...
+    if (rb == nullptr) {
+        return grpc::Status(grpc::UNKNOWN, "resource not found: " + request->name());
+    }
+
+    std::shared_ptr<Camera> camera = std::dynamic_pointer_cast<Camera>(rb);
+
+    std::string name = request->name();
+    std::string mime_type = request->mime_type();
+
+    Camera::raw_image image = camera->get_image(name, mime_type);
```
stuqdog, line 79: (nit) same nit about combining lines :)

src/components/camera/server.cpp
```diff
@@ -0,0 +1,162 @@
...
+
+    std::shared_ptr<ResourceBase> rb = sub_svc->resource(request->name());
+
+    if (rb == nullptr) {
+        return grpc::Status(grpc::UNKNOWN, "resource not found: " + request->name());
+    }
+
+    std::shared_ptr<Camera> camera = std::dynamic_pointer_cast<Camera>(rb);
+
+    std::string name = request->name();
```
stuqdog, line 106: (nit) same nit :)

src/components/camera/server.cpp
```diff
@@ -0,0 +1,162 @@
...
+    std::shared_ptr<ResourceBase> rb = sub_svc->resource(request->name());
+    if (rb == nullptr) {
+        return grpc::Status(grpc::UNKNOWN, "resource not found: " + request->name());
+    }
+
+    std::shared_ptr<Camera> camera =std::dynamic_pointer_cast<Camera>(rb);
+
+    std::string name = request->name();
+
+    Camera::properties properties = camera->get_properties(name);
```
stuqdog, line 137: (nit) same nit :)

src/components/resource_manager.hpp
```diff
@@ -0,0 +1,33 @@
+#pragma once
```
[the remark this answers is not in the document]
stuqdog: @cheukt sorry about this, this file got accidentally re-added when we were dealing with rebase stuff! We do have a more generalized `resource_manager` already, this is just an artifact :)

src/common/proto_type.cpp
```diff
@@ -46,32 +48,46 @@ ProtoType ProtoType::of_value(Value value) {
...
     }
 
-    return map;
+    return std::make_shared<std::unordered_map<std::string, std::shared_ptr<ProtoType>>>(map);
+    
 };
 
 Value ProtoType::proto_value() {
     Value v;
+       try {
```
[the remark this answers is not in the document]
oliviamiller: We already have a default that throws with the message `Invalid proto_value conversion type. This should never happen; please file a bug report.";`. Is it good if I remove the try catch and keep that error or should I change it to runtime_exception?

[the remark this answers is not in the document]
stuqdog: Oh, you're absolutely right, we do already have the default case! 🤦 thanks for catching that :)

In that case, I think we should get rid of this `try catch` here and change the `throw` at the bottom to a `runtime_exception` instead of a string.

src/common/proto_type.cpp
```diff
@@ -122,3 +143,110 @@ Value ProtoType::proto_value() {
...
+            std::shared_ptr<std::unordered_map<std::string, std::shared_ptr<ProtoType>>> lhs_map =
+                boost::get<std::shared_ptr<std::unordered_map<std::string, std::shared_ptr<ProtoType>>>>(lhs.proto_type);
+            std::shared_ptr<std::unordered_map<std::string, std::shared_ptr<ProtoType>>> rhs_map =
+                boost::get<std::shared_ptr<std::unordered_map<std::string, std::shared_ptr<ProtoType>>>>(rhs.proto_type);
+            
+            auto it1 = *lhs_map->begin();
+            auto it2 = *rhs_map->begin();
+
+
+            ProtoType lhs_type = *lhs_map->at(it1.first);
```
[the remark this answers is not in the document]
oliviamiller: The unordered map equals operator didn't work, not 100% sure why but I'm guessing because of the custom type in the map. However looking at this again pretty sure this would only test the equality for the first element in the map so fixed that, and added a unit test so I could confirm it worked :)

src/config/resource.hpp
oliviamiller: was causing errors in a few files but changed a lot of them to AttributeMap!

src/components/camera/camera.cpp
```diff
@@ -0,0 +1,150 @@
...
+}
+
+namespace {
+bool init() {
+    Registry::register_subtype(Camera::subtype(), Camera::resource_subtype());
+    return true;
+};
+
+bool inited = init();
+}  // namespace
```
stuqdog, line 150: see above re: terminating newline

src/components/camera/camera.hpp
```diff
@@ -0,0 +1,82 @@
...
+
+    static std::shared_ptr<ResourceSubtype> resource_subtype();
+    static Subtype subtype();
+    static raw_image from_proto(viam::component::camera::v1::GetImageResponse proto);
+    static point_cloud from_proto(viam::component::camera::v1::GetPointCloudResponse proto);
+    static intrinsic_parameters from_proto(viam::component::camera::v1::IntrinsicParameters proto);
+    static distortion_parameters from_proto(
+        viam::component::camera::v1::DistortionParameters proto);
+    static properties from_proto(viam::component::camera::v1::GetPropertiesResponse proto);
+    static std::vector<double> repeated_field_to_vector(
```
stuqdog, line 65: I don't think `repeated_field_to_vector` or `vector_to_repeated_field` are intrinsically `camera`-y and it doesn't look like they ever get called outside of `camera.cpp`. Can we just define these as regular functions within `camera.cpp` and not expose them to the outside world at all?

src/components/camera/client.cpp
```diff
@@ -0,0 +1,77 @@
...
+#include <components/camera/camera.hpp>
+#include <components/camera/client.hpp>
+#include <config/resource.hpp>
+#include <string>
+#include <utility>
+
+#include "common/v1/common.pb.h"
+#include "component/camera/v1/camera.grpc.pb.h"
+
+std::string normalize_mime_type(const std::string& str);
```
stuqdog, line 12: Can we just define `normalize_mime_type` up here so we don't have to forward declare it?

src/components/camera/client.cpp
```diff
@@ -0,0 +1,77 @@
...
+                    Camera::lazy_suffix) == 0) {
+        mime_type = mime_type.substr(0, mime_type.length() - Camera::lazy_suffix.length());
+    }
+    return mime_type;
+}
+
+CameraClient::CameraClient(std::string name, std::shared_ptr<grpc::Channel> channel_)
+    : channel_(channel_), stub_(viam::component::camera::v1::CameraService::NewStub(channel_)){};
+
+CameraClient::CameraClient(std::string name) : name_(name), channel_(nullptr), stub_(nullptr){};
```
stuqdog, line 77: see above re: terminating newline

src/components/camera/server.cpp
```diff
@@ -0,0 +1,137 @@
...
+    *response->mutable_intrinsic_parameters() = Camera::to_proto(properties.intrinsic_parameters);
+    response->set_supports_pcd(properties.supports_pcd);
+
+    return ::grpc::Status();
+}
+
+void CameraServer::register_server() {
+    viam::component::camera::v1::CameraService::Service* camera =
+        static_cast<viam::component::camera::v1::CameraService::Service*>(this);
+    try {
```
stuqdog, line 132: (nit) we don't need this `try/catch` statement, since it's just immediately throwing the exception it catches. just calling `Server::register_service(camera);` is fine

src/components/component_base.cpp
stuqdog, line 1: I'm a little confused as to why this file is empty... on `main` there is code here so at a minimum I'd expect to see a diff. It's strange! Anyway, is there a reason why this is now empty? And if not, can we restore the code that was there?

src/components/generic/client.hpp
```diff
@@ -5,7 +5,7 @@
 
 class GenericClient : public Generic {
    public:
-    ProtoType do_command(std::unordered_map<std::string, ProtoType*> command) override;
+    virtual AttributeMap do_command(AttributeMap command) override;
```
stuqdog, line 8: (nit) I don't think we need to mark this as `virtual`, since it's already marked as an `override`

src/common/proto_type.hpp
```diff
@@ -12,7 +12,7 @@ class ProtoType {
                    std::string,
                    int,
                    double,
-                   std::unordered_map<std::string, ProtoType*>,
+                   std::shared_ptr<std::unordered_map<std::string, std::shared_ptr<ProtoType>>>,
```
stuqdog, line 15: is there a reason we aren't using `AttributeMap` here?

stuqdog (changes requested):
Thanks for the quick turnaround on this! I have a couple minor comments, but otherwise this all looks great

src/common/proto_type.hpp
oliviamiller: I tried but I'm getting `use of undeclared identifier 'ProtoType'` and ` error: constructor cannot be redeclared`. might be a circular dependency?

src/components/component_base.cpp
oliviamiller: Im seeing the same code in this file on the main and on my branch, so I think it's good? its not showing as empty for me.

src/common/proto_type.cpp
```diff
@@ -115,10 +119,68 @@ Value ProtoType::proto_value() {
...
+            std::vector<ProtoType*> lhs_vec = boost::get<std::vector<ProtoType*>>(lhs.proto_type);
+            std::vector<ProtoType*> rhs_vec = boost::get<std::vector<ProtoType*>>(rhs.proto_type);
+            return lhs_vec == rhs_vec;
+        }
+        default: {
+            throw "Invalid proto_value conversion type. This should never happen;\
+						please file a bug report.";
+        }
+    }
+}
```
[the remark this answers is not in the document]
oliviamiller: oops sorry weird that the linter didn't do this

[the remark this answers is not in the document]
stuqdog: all good! it is a bit surprising, I agree :)

src/components/component_base.cpp
stuqdog: huh that's so strange! when I look at the review, it says `empty file`, but looking at the actual repo, I do see that all the code is there. Anyway, it's probably fine :)

src/tests/test_camera.cpp
```diff
@@ -0,0 +1,241 @@
...
+    AttributeMap command;
+    AttributeMap result_map = camera->do_command(command);
+
+    ProtoType expected_pt = *(expected->at(std::string("test")));
+    ProtoType result_pt = *(result_map->at(std::string("test")));
+
+    BOOST_CHECK(result_pt == expected_pt);
+}
+
+BOOST_AUTO_TEST_CASE(test_prototype_equality) {
```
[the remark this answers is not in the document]
oliviamiller: moved! Also realized that function still wasn't quite right so fixed and added another check to the test.

stuqdog (approved):
lgtm!

cheukt (approved):
looks great!

PR #56, opened 2023-03-24
RSDK-2510 - Fix builds in linux

stuqdog:
Major changes:

- Adds workflow to build and test in CI (note: this is currently broken due to github actions being unable to install versions of protoc past 3.20.3. however, this builds and tests pass locally for both linux and mac. Will update workflow in a subsequent PR).
- Refactors tests to remove extraneous code from mock stubs
- Refactors server tests to more directly call to server code

Minor changes:
- Remove C++17-specific syntax
- Stop auto sorting headers


src/tests/test_camera.cpp
```diff
@@ -114,11 +120,12 @@ BOOST_AUTO_TEST_CASE(test_render_frame_service) {
 }
 
 BOOST_AUTO_TEST_CASE(test_get_properties_service) {
-    grpc::ClientContext ctx;
+    auto server = get_camera_server();
```
oliviamiller, line 123: is there a reason we need to create the server in every test instead of just doing it once?

stuqdog: Good question! My thinking was, 1) avoiding global values is generally good, 2) it isn't particularly expensive to recreate it, especially given it's only a testing context, and 3) it wasn't immediately obvious how to insert into it (as we do in the `get_camera_server()` function) outside a function. There probably is a way but I didn't dig particularly because of reasons 1 and 2 above :)

Happy to chat/revert if you have a particular preference!

oliviamiller (approved):

PR #57, opened 2023-03-24
RSDK-2391 Unit testing for generic component 

oliviamiller:


src/components/generic/server.hpp
```diff
@@ -19,7 +19,5 @@ class GenericServer
 
   GenericServer(){};
   GenericServer(std::shared_ptr<SubtypeService> sub_svc) : sub_svc(sub_svc){};
-
-private:
   std::shared_ptr<SubtypeService> sub_svc;
```
stuqdog, line 22: this should probably stay private. If we need to access it from outside the class then we can always add an accessor function.

src/tests/mocks/generic_mocks.cpp
stuqdog, line 1: We should make a header file for this and include that in the test rather than the CPP itself. I think I missed this in the last PR (sorry!), but we should do the same for `camera`.

stuqdog (changes requested):
A couple small comments but overall this looks great! Thanks for putting it together.

oliviamiller:
builds for me!

PR #58, opened 2023-03-27
RSDK-2114 - Fix reflection library importing

stuqdog:
Was able to install and build CPP SDK on a fresh linux system with this PR.


CMakeLists.txt
```diff
@@ -329,9 +330,18 @@ else()
...
+  
+  # We need to set the reflection library, which unfortunately does not 
+  # have a `.pc` file. To do so, we duplicate the `VIAMCPPSDK_GRPCXX_LIBRARIES` 
+  # value, assume its first element is the base dependency for GRPC, and 
+  # prepend our new list with a new value, equal to that base dependency plus 
+  # the `::grpc++_reflection` suffix for finding the reflection library.
+  list(GET VIAMCPPSDK_GRPCXX_LIBRARIES 0 VIAMCPPSDK_GRPC_ROOT)
+  set(VIAMCPPSDK_GRPCXX_REFLECTION_LIBRARIES ${VIAMCPPSDK_GRPCXX_LIBRARIES})
+  set(VIAMCPPSDK_GRPC_REFLECTION_LIB ${VIAMCPPSDK_GRPC_ROOT}::grpc++_reflection)
+  list(PREPEND VIAMCPPSDK_GRPCXX_REFLECTION_LIBRARIES VIAMCPPSDK_GRPC_REFLECTION_LIB)
```
acmorrow, line 342: I think this is missing `${...}` around `VIAMCPPSDK_GRPC_REFLECTION_LIB`

CMakeLists.txt
```diff
@@ -329,9 +330,18 @@ else()
...
   set(VIAMCPPSDK_GRPCXX_LIBRARIES ${GRPCXX_LIBRARIES})
+  
+  # We need to set the reflection library, which unfortunately does not 
+  # have a `.pc` file. To do so, we duplicate the `VIAMCPPSDK_GRPCXX_LIBRARIES` 
+  # value, assume its first element is the base dependency for GRPC, and 
+  # prepend our new list with a new value, equal to that base dependency plus 
+  # the `::grpc++_reflection` suffix for finding the reflection library.
+  list(GET VIAMCPPSDK_GRPCXX_LIBRARIES 0 VIAMCPPSDK_GRPC_ROOT)
+  set(VIAMCPPSDK_GRPCXX_REFLECTION_LIBRARIES ${VIAMCPPSDK_GRPCXX_LIBRARIES})
+  set(VIAMCPPSDK_GRPC_REFLECTION_LIB ${VIAMCPPSDK_GRPC_ROOT}::grpc++_reflection)
```
acmorrow, line 341: It is a little confusing that some of these are `GRPC_` and others `GRPCXX_`.

acmorrow, line 341: The `gRPC::<x>` form of library is only something that happens on the happy path where we obtain the gRPC infra via `find_package`. The logic here is in the fallback case where we need to try to obtain it via `pkg_check_modules` (aka pkgconfig). In that model, the values in `VIAMCPPSDK_GRPCXX_LIBRARIES` are just going to be a list of library names, like `grpc++;grpc;gpr`. When I run your code right now (after correcting the issue I note below), I get:

```
XXX VIAMCPPSDK_GRPCXX_LIBRARIES grpc++;grpc;gpr
XXX VIAMCPPSDK_GRPCXX_REFLECTION_LIBRARIES grpc++::grpc++_reflection;grpc++;grpc;gpr
```

So I think you should just drop the `::grpc++` part:
```
set(VIAMCPPSDK_GRPC_REFLECTION_LIB ${VIAMCPPSDK_GRPC_ROOT}::_reflection)
```

If you use the `bullseye` Docker container you should be able to test this.

src/CMakeLists.txt
```diff
@@ -334,11 +334,11 @@ target_link_directories(viamcpp
 
 target_link_libraries(viamcpp
   PUBLIC ${VIAMCPPSDK_GRPCXX_LIBRARIES}
+  PUBLIC ${VIAMCPPSDK_GRPCXX_REFLECTION_LIBRARIES}
```
acmorrow, line 337: This was previously `PRIVATE`. Does it need to be promoted to `PUBLIC`?

acmorrow (changes requested):

CMakeLists.txt
acmorrow: Small error above: I forgot to take out `::` too. So:
```
set(VIAMCPPSDK_GRPC_REFLECTION_LIB ${VIAMCPPSDK_GRPC_ROOT}_reflection)
```

stuqdog: Agreed! adding `XX` to all for consistency.

CMakeLists.txt
```diff
@@ -329,9 +330,18 @@ else()
   set(VIAMCPPSDK_GRPCXX_INCLUDE_DIRS ${GRPCXX_INCLUDE_DIRS})
   set(VIAMCPPSDK_GRPCXX_LIBRARY_DIRS ${GRPCXX_LIBRARY_DIRS})
   set(VIAMCPPSDK_GRPCXX_LIBRARIES ${GRPCXX_LIBRARIES})
+  
+  # We need to set the reflection library, which unfortunately does not 
+  # have a `.pc` file. To do so, we duplicate the `VIAMCPPSDK_GRPCXX_LIBRARIES` 
+  # value, assume its first element is the base dependency for GRPC, and 
+  # prepend our new list with a new value, equal to that base dependency plus
```
acmorrow, line 337: It isn't easy to see here, but these four lines have trailing whitespace. Personally, I like to configure my editor to make trailing whitespace visible so I don't forget to clean it up.

benjirewis (approved):
I don't see any relevant error messages on Ubuntu 22.04 with these changes when trying to build on my machine.

PR #63, opened 2023-03-31
RSDK-2436 - testing in ci

stuqdog:
#### Major Changes ####
Fixes our CI testing for C++ (thanks to @cheukt and @maximpertsov for the help!)

#### Testing ####
It works and the CI test passes. When adding a failing test, the CI test fails.


cheukt (approved):

etc/docker/tests/Dockerfile.debian.bullseye
```diff
@@ -0,0 +1,43 @@
...
+
+
+RUN apt-add-repository 'deb http://deb.debian.org/debian bullseye-backports main'
+
+RUN apt-get update
+
+RUN apt-get -y --no-install-recommends install -t bullseye-backports \
+    cmake
+
+COPY . /tmp/
```
maximpertsov, line 41: non-blocking - if we find that this step is slow or we have container size issues, we could add a `.dockerignore` in our repo root to not copy over stuff that isn't necessary (e.g. `.git`, existing builds)

maximpertsov (approved):

etc/docker/tests/Dockerfile.debian.bullseye
stuqdog: Good call. I created a general "CI testing improvements" ticket in jira, added a comment to this effect there.

PR #65, opened 2023-04-03
RSDK-2352 - setup UB sanitizer, add -Wall and -Werror flags

stuqdog:
#### Major changes #### 

- Add flag for building with `-Wall` and `-Werror` (defaults to `on`)
- Add flag for building with UB sanitizer (defaults to `off`)

#### Minor changes ####

- Miscellaneous refactors to ensure successful building with `Wall/Werror` flags and no errors

#### Testing ####

- Builds successfully with flags passed
- Unit tests pass with instrumented build
- `dial` example runs successfully on mac with both ubsan and asan

#### Notes ####
- Currently we are not building with asan. This is because unit tests are failing with asan on linux. This is possibly due to `gRPC` not being similarly built with asan. We don't yet support doing this ourselves; this is a future goal however, see [here](https://viam.atlassian.net/browse/RSDK-2553).

#### Major changes #### 

- Add flag for building with `-Wall` and `-Werror` (defaults to `on`)
- Add flag for building with address sanitizer and UB sanitizer (defaults to `off`)

#### Minor changes ####

- Miscellaneous refactors to ensure successful building with `Wall/Werror` flags and no errors

#### Testing ####

- Builds successfully with flags passed
- Unit tests pass with instrumented build

#### Notes ####
- Currently integration testing/production use with an instrumented build is expected to fail unless `gRPC` is similarly built with sanitizers. We don't yet support doing this ourselves; this is a future goal however, see [here](https://viam.atlassian.net/browse/RSDK-2553).

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -81,6 +81,9 @@ option(BUILD_SHARED_LIBS "If enabled, build shared libraries" ON)
 option(VIAMCPPSDK_ENFORCE_COMPILER_MINIMA "Control whether we validate the selected compiler against known minima" ON)
 option(VIAMCPPSDK_OFFLINE_PROTO_GENERATION "Use local tools for proto generation" OFF)
 option(VIAMCPPSDK_USE_DYNAMIC_PROTOS "Regenerate protos as part of building (network access may be required)" OFF)
+option(VIAMCPPSDK_USE_WALL_WERROR "Build with -Wall and -Werror flags" ON)
+# N.b.: `VIAMCPPSDK_SANITIZED_BUILD` may cause runtime failures if gRPC is not also built with sanitizers.
+option(VIAMCPPSDK_SANITIZED_BUILD "Build with address and UB sanitizers (less performant)" OFF)
 
 
 # Enforce known toolchains and toolchain minima unless asked not to.
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -57,13 +57,15 @@ option(BUILD_SHARED_LIBS "If enabled, build shared libraries" ON)
 #
 # - `VIAMCPPSDK_ENFORCE_COMPILER_MINIMA`: The user can elect to
 # disable enforcement of compiler minima (at their peril).
-#
-#
-# # - `VIAMCPPSDK_LOCAL_PROTO_GENERATION`: Do not use buf.builds remote
+option(VIAMCPPSDK_ENFORCE_COMPILER_MINIMA "Control whether we validate the selected compiler against known minima" ON)
+
+
+# # - `VIAMCPPSDK_OFFLINE_PROTO_GENERATION`: Do not use buf.builds remote
 # definitions or services and use the local proto compiler plugins to
 # do generation.
-#
-#
+option(VIAMCPPSDK_OFFLINE_PROTO_GENERATION "Use local tools for proto generation" OFF)
+
+
 # - `VIAMCPPSDK_USE_DYNAMIC_PROTOS`: The user can select whether to
 # use the prebuilt proto files under the `gen` directory, or to
 # dynamically regenerate them as part of the build. In the latter
@@ -75,14 +77,18 @@ option(BUILD_SHARED_LIBS "If enabled, build shared libraries" ON)
 # cmake ... -DVIAMCPPSDK_USE_DYNAMIC_PROTOS=ON
 # [make|ninja|...] update-static-protos
 # ```
-#
 # Remember to check-in the results after regenerating!
-#
-option(VIAMCPPSDK_ENFORCE_COMPILER_MINIMA "Control whether we validate the selected compiler against known minima" ON)
-option(VIAMCPPSDK_OFFLINE_PROTO_GENERATION "Use local tools for proto generation" OFF)
 option(VIAMCPPSDK_USE_DYNAMIC_PROTOS "Regenerate protos as part of building (network access may be required)" OFF)
+
+
+# # - `VIAMCPPSDK_USE_WALL_WERROR`: This causes the SDK's internal
+# code to compile with `-Wall` and `-Werror` flags.
 option(VIAMCPPSDK_USE_WALL_WERROR "Build with -Wall and -Werror flags" ON)
-# N.b.: `VIAMCPPSDK_SANITIZED_BUILD` may cause runtime failures if gRPC is not also built with sanitizers.
+
+
+# # - `VIAMCPPSDK_SANITIZED_BUILD`: This causes the SDK to build with
+# UBSan, helping in the detection of undefined behavior. Note that builds
+# with this flag will be less performant.
 option(VIAMCPPSDK_SANITIZED_BUILD "Build with address and UB sanitizers (less performant)" OFF)
 
 
--- a/src/viam/sdk/CMakeLists.txt
+++ b/src/viam/sdk/CMakeLists.txt
@@ -166,8 +166,8 @@ if (VIAMCPPSDK_USE_WALL_WERROR)
   target_compile_options(viamsdk PRIVATE -Wall -Werror)
 endif()
 if (VIAMCPPSDK_SANITIZED_BUILD)
-  target_link_options(viamsdk PRIVATE -fsanitize=undefined -fsanitize=address -fno-omit-frame-pointer -fno-sanitize-recover)
-  target_compile_options(viamsdk PRIVATE -fsanitize=undefined -fsanitize=address -fno-omit-frame-pointer -fno-sanitize-recover)
+  target_link_options(viamsdk PRIVATE -fsanitize=undefined -fno-omit-frame-pointer -fno-sanitize-recover)
+  target_compile_options(viamsdk PRIVATE -fsanitize=undefined -fno-omit-frame-pointer -fno-sanitize-recover)
 endif()
 
 
--- a/src/viam/sdk/CMakeLists.txt
+++ b/src/viam/sdk/CMakeLists.txt
@@ -158,8 +158,6 @@ target_link_libraries(viamsdk
   PRIVATE Threads::Threads
 )
 
-# Build api as a system library so we Wall/Werror don't cause unsolvable compilation failures
-target_include_directories(viamapi SYSTEM INTERFACE)
 
 # Set compile and link options based on arguments
 if (VIAMCPPSDK_USE_WALL_WERROR)
--- a/src/viam/sdk/CMakeLists.txt
+++ b/src/viam/sdk/CMakeLists.txt
@@ -158,6 +158,18 @@ target_link_libraries(viamsdk
   PRIVATE Threads::Threads
 )
 
+# Build api as a system library so we Wall/Werror don't cause unsolvable compilation failures
+target_include_directories(viamapi SYSTEM INTERFACE)
+
+# Set compile and link options based on arguments
+if (VIAMCPPSDK_USE_WALL_WERROR)
+  target_compile_options(viamsdk PRIVATE -Wall -Werror)
+endif()
+if (VIAMCPPSDK_SANITIZED_BUILD)
+  target_link_options(viamsdk PRIVATE -fsanitize=undefined -fsanitize=address -fno-omit-frame-pointer -fno-sanitize-recover)
+  target_compile_options(viamsdk PRIVATE -fsanitize=undefined -fsanitize=address -fno-omit-frame-pointer -fno-sanitize-recover)
+endif()
+
 
 install(
   TARGETS viamsdk
--- a/src/viam/sdk/CMakeLists.txt
+++ b/src/viam/sdk/CMakeLists.txt
@@ -19,6 +19,17 @@ target_include_directories(viamsdk
     "$<BUILD_INTERFACE:$<TARGET_PROPERTY:viam-cpp-sdk::viamapi,INTERFACE_INCLUDE_DIRECTORIES>>"
 )
 
+
+# Set compile and link options based on arguments
+if (VIAMCPPSDK_USE_WALL_WERROR)
+  target_compile_options(viamsdk PRIVATE -Wall -Werror)
+endif()
+if (VIAMCPPSDK_SANITIZED_BUILD)
+  target_link_options(viamsdk PRIVATE -fsanitize=undefined -fno-omit-frame-pointer -fno-sanitize-recover)
+  target_compile_options(viamsdk PRIVATE -fsanitize=undefined -fno-omit-frame-pointer -fno-sanitize-recover)
+endif()
+
+
 target_sources(viamsdk
   PRIVATE
     # TODO(RSDK-1742): we have to put `registry` up top here because we need the
@@ -157,16 +168,6 @@ target_link_libraries(viamsdk
   PRIVATE viam_rust_utils
   PRIVATE Threads::Threads
 )
-
-
-# Set compile and link options based on arguments
-if (VIAMCPPSDK_USE_WALL_WERROR)
-  target_compile_options(viamsdk PRIVATE -Wall -Werror)
-endif()
-if (VIAMCPPSDK_SANITIZED_BUILD)
-  target_link_options(viamsdk PRIVATE -fsanitize=undefined -fno-omit-frame-pointer -fno-sanitize-recover)
-  target_compile_options(viamsdk PRIVATE -fsanitize=undefined -fno-omit-frame-pointer -fno-sanitize-recover)
-endif()
 
 
 install(
--- a/src/viam/sdk/components/camera/camera.hpp
+++ b/src/viam/sdk/components/camera/camera.hpp
@@ -25,6 +25,7 @@ namespace sdk {
 /// @ingroup Camera
 class CameraSubtype : public ResourceSubtype {
    public:
+    virtual ~CameraSubtype() = default;
     std::shared_ptr<ResourceServerBase> create_resource_server(
         std::shared_ptr<SubtypeService> svc) override;
     std::shared_ptr<ResourceBase> create_rpc_client(std::string name,
--- a/src/viam/sdk/components/camera/camera.hpp
+++ b/src/viam/sdk/components/camera/camera.hpp
@@ -25,7 +25,7 @@ namespace sdk {
 /// @ingroup Camera
 class CameraSubtype : public ResourceSubtype {
    public:
-    virtual ~CameraSubtype() = default;
+    virtual ~CameraSubtype();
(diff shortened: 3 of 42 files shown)


zaporter-work (approved):

src/tests/test_utils.cpp
```diff
@@ -6,8 +6,7 @@
 #include <config/resource.hpp>
 
 AttributeMap fake_map() {
-    std::shared_ptr<ProtoType> proto_ptr =
-        std::make_shared<ProtoType>(std::move(std::string("hello")));
+    std::shared_ptr<ProtoType> proto_ptr = std::make_shared<ProtoType>(std::string("hello"));
```
clintpurser, line 9: just wondering, why remove the `std::move()`? I think it does seem a bit excessive for a string so 👍.  Thanks!

CMakeLists.txt
```diff
@@ -81,6 +81,9 @@ option(BUILD_SHARED_LIBS "If enabled, build shared libraries" ON)
 option(VIAMCPPSDK_ENFORCE_COMPILER_MINIMA "Control whether we validate the selected compiler against known minima" ON)
 option(VIAMCPPSDK_OFFLINE_PROTO_GENERATION "Use local tools for proto generation" OFF)
 option(VIAMCPPSDK_USE_DYNAMIC_PROTOS "Regenerate protos as part of building (network access may be required)" OFF)
+option(VIAMCPPSDK_USE_WALL_WERROR "Build with -Wall and -Werror flags" ON)
+# N.b.: `VIAMCPPSDK_SANITIZED_BUILD` may cause runtime failures if gRPC is not also build with sanitizers.
```
clintpurser, line 85: [nitpick] spelling

```suggestion
# N.b.: `VIAMCPPSDK_SANITIZED_BUILD` may cause runtime failures if gRPC is not also built with sanitizers.
```

src/tests/test_utils.cpp
stuqdog: the `std::move` was leading to `moving a temporary object prevents copy elision` warning from `Wall`

clintpurser: I see, thank you for helping me understand! 👍

CMakeLists.txt
```diff
@@ -141,6 +144,18 @@ set(CMAKE_MODULE_PATH
 )
 
 
+# Set linker and compiler flags based on arguments.
+if (VIAMCPPSDK_USE_WALL_WERROR)
+  SET(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wall -Werror")
```
acmorrow, line 149: Applying these to `CMAKE_CXX_FLAGS` may be too big a hammer. I'd suggest looking into whether you can achieve this with `add_compile_options` and `add_link_options` at the directory level:
- https://cmake.org/cmake/help/latest/command/add_compile_options.html
- https://cmake.org/cmake/help/latest/command/add_link_options.html
You probably also want to apply these as `PRIVATE` so they don't propagate to SDK consumers that use the library. I'm actually not sure if you can do that with `add_compile_options` / `add_link_options`. You could do it with `target_compile_options` for each target but that's a pain to need to add everywhere. Hopefully there is a way to both make it private and apply it once.

CMakeLists.txt
```diff
@@ -141,6 +144,18 @@ set(CMAKE_MODULE_PATH
...
+  SET(CMAKE_LINKER_FLAGS "${CMAKE_LINKER_FLAGS} -Wall -Werror")
+endif()
+
+if (VIAMCPPSDK_SANITIZED_BUILD)
```
acmorrow, line 153: @stuqdog - You were having crashes when doing this last time we talked about it. Did you figure out what was going wrong? Do builds with these flags work?

Also, all the same thoughts above about attaching these at the target / directory level. Need to think through whether these should be `PRIVATE` or not.

src/components/camera/client.cpp
```diff
@@ -72,6 +72,8 @@ Camera::properties CameraClient::get_properties() {
 };
 
 CameraClient::CameraClient(std::string name, std::shared_ptr<grpc::Channel> channel_)
-    : channel_(channel_), stub_(viam::component::camera::v1::CameraService::NewStub(channel_)){};
+    : name_(std::move(name)),
+      stub_(viam::component::camera::v1::CameraService::NewStub(channel_)),
+      channel_(channel_){};
```
acmorrow, line 77: Move on `channel` too.

src/components/camera/client.hpp
```diff
@@ -18,8 +19,8 @@ class CameraClient : public Camera {
 
    protected:
     CameraClient(std::string name);
-    std::unique_ptr<viam::component::camera::v1::CameraService::StubInterface> stub_;
     std::string name_;
+    std::unique_ptr<viam::component::camera::v1::CameraService::StubInterface> stub_;
```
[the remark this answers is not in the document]
acmorrow: I actually think you don't even need the `nullptr`: I think `std::unique_ptr<...> stub_{}` will do.

CMakeLists.txt
acmorrow: Some interesting thoughts on doing this here: https://www.foonathan.net/2018/10/cmake-warnings/

stuqdog: Ack, appreciate the feedback! Will look into this and push updates.

CMakeLists.txt
stuqdog: The crashes still exist for purposes of integration testing - as I recall this was the due to gRPC not being built with the sanitizers, so we were running into size issues. I added a comment to this effect. 

Where we stand now is still valuable in the meantime however, as it allows us to still run all unit tests with sanitized builds on. Updating the gRPC build is on the docket along with integration testing support.

src/viam/sdk/components/generic/server.cpp
```diff
@@ -28,13 +28,7 @@ ::grpc::Status GenericServer::DoCommand(::grpc::ServerContext* context,
...
 
 void GenericServer::register_server(std::shared_ptr<Server> server) {
-    viam::component::generic::v1::GenericService::Service* generic =
-        static_cast<viam::component::generic::v1::GenericService::Service*>(this);
-    try {
-        server->register_service(generic);
-    } catch (std::exception& exc) {
-        throw exc;
-    }
+    server->register_service(this);
```
stuqdog, line 31: (flyby) another unnecessary `static_cast`

src/viam/sdk/registry/registry.cpp
```diff
@@ -77,31 +77,7 @@ Registry::registered_resources() {
...
-            BOOST_LOG_TRIVIAL(debug)
-                << "Unable to get resource name from resource base " << resource->name();
-        };
-    } else {
-        throw "unable to create status; provided resource was of an unknown "
-          "type: " +
-        resource->type().to_string();
-    }
-    *status.mutable_name() = *name;
+    *status.mutable_name() = resource->get_resource_name(resource->name());
```
stuqdog, line 80: (flyby) `get_resource_name` is now baked into `ResourceBase`, so no need to do all that casting into other types and checking. if we can get a `ResourceName`.

src/viam/sdk/resource/resource.cpp
```diff
@@ -80,7 +80,7 @@ std::string Name::to_string() const {
 }
 
 std::string Name::short_name() const {
-    if (remote_name == "") {
+    if (remote_name != "") {
```
stuqdog, line 83: (flyby) small bug, easy to fix here

src/viam/sdk/robot/client.cpp
```diff
@@ -191,7 +191,7 @@ void RobotClient::refresh() {
     }
     bool is_equal = current_resources.size() == resource_names_.size();
     if (is_equal) {
-        for (int i = 0; i < current_resources.size(); ++i) {
+        for (long unsigned int i = 0; i < resource_names_.size(); ++i) {
```
stuqdog, line 194: @acmorrow I got compiler errors here when I used both `int` and `auto`. having `long unsigned int` here feels super wordy and a bit smelly, but I wasn't sure how to fix it. Any thoughts?

src/viam/sdk/robot/client.hpp
```diff
@@ -64,10 +64,10 @@ class RobotClient {
     /// @param options Options for connecting and refreshing.
     /// Connects directly to a pre-existing channel. A robot created this way must be
     /// `close()`d manually.
-    static std::shared_ptr<RobotClient> with_channel(ViamChannel channel, Options options);
-    RobotClient(ViamChannel channel);
+    static std::shared_ptr<RobotClient> with_channel(std::shared_ptr<ViamChannel> channel,
+                                                     Options options);
+    RobotClient(std::shared_ptr<ViamChannel> channel);
     std::vector<ResourceName>* resource_names();
-    std::unique_ptr<RobotService::Stub> stub_;
```
stuqdog, line 70: (flyby) had to move this anyways to make sure params were being set in the right order, might as well move it to `private`.

src/viam/sdk/robot/client.cpp
```diff
@@ -253,21 +254,21 @@ std::shared_ptr<RobotClient> RobotClient::with_channel(ViamChannel channel, Opti
...
-    address = "unix://" + address;
-    const char* uri = address.c_str();
+    std::string addr = "unix://" + address;
+    const char* uri = addr.c_str();
     std::shared_ptr<grpc::Channel> channel =
         grpc::CreateChannel(uri, grpc::InsecureChannelCredentials());
     std::unique_ptr<viam::robot::v1::RobotService::Stub> st =
         viam::robot::v1::RobotService::NewStub(channel);
-    ViamChannel viam_channel = ViamChannel(channel, uri, nullptr);
+    auto viam_channel = std::make_shared<ViamChannel>(channel, address.c_str(), nullptr);
```
stuqdog, line 271: this was a fun one! we were passing an updated `addr` with the `unix://` prefix to the rust_utils `free_string` function. But! `free_string` calls rust's `c_str::from_raw`, which must take as an argument a string that was previously created by a call to `c_str::into_raw`. So, the `unix://` prefix was breaking things.

clintpurser:
removing myself as reviewer, I think andrew has covered things well. 👍

CMakeLists.txt
stuqdog: (update) everything works fine with ubsan and asan on mac. asan is causing issues on linux so I removed it from this PR, will look into it in a subsequent PR.

src/viam/sdk/robot/client.cpp
```diff
@@ -69,7 +69,8 @@ void RobotClient::close() {
         t->~thread();
     }
     stop_all();
-    viam_channel.close();
+    viam_channel->close();
+    viam_channel = nullptr;
```
zaporter-work, line 73: [q] Why do we do this here?

zaporter-work, line 73: [q/nit] I don't know if we should set this to `nullptr`. `ViamChannel` has a bool called `closed` that I would expect to be able to check after calling `close()` on this client.

src/viam/sdk/CMakeLists.txt
```diff
@@ -158,6 +158,18 @@ target_link_libraries(viamsdk
   PRIVATE Threads::Threads
 )
 
+# Build api as a system library so we Wall/Werror don't cause unsolvable compilation failures
+target_include_directories(viamapi SYSTEM INTERFACE)
```
zaporter-work, line 162: [q] Should this be done in the api `CMakeLists`?

zaporter-work (approved):
LGTM

src/viam/sdk/robot/client.cpp
stuqdog: Good catch, thanks! This was an artifact of some testing/debugging around an asan issue, but you're right, it's not necessary. removed.

src/viam/sdk/CMakeLists.txt
stuqdog: That's a good question. I put it here under the thinking that we're determining how the `sdk` consumes the `api`, so it's a decision _for_ the sdk, if that makes sense. But there are likely subtleties at play here I'm not considering. I think I'd defer to @acmorrow's thinking on this.

CMakeLists.txt
```diff
@@ -81,6 +81,9 @@ option(BUILD_SHARED_LIBS "If enabled, build shared libraries" ON)
 option(VIAMCPPSDK_ENFORCE_COMPILER_MINIMA "Control whether we validate the selected compiler against known minima" ON)
 option(VIAMCPPSDK_OFFLINE_PROTO_GENERATION "Use local tools for proto generation" OFF)
 option(VIAMCPPSDK_USE_DYNAMIC_PROTOS "Regenerate protos as part of building (network access may be required)" OFF)
+option(VIAMCPPSDK_USE_WALL_WERROR "Build with -Wall and -Werror flags" ON)
+# N.b.: `VIAMCPPSDK_SANITIZED_BUILD` may cause runtime failures if gRPC is not also built with sanitizers.
+option(VIAMCPPSDK_SANITIZED_BUILD "Build with address and UB sanitizers (less performant)" OFF)
```
acmorrow, line 86: There are detailed comments above for each option. Consider adding some text for these as well.

acmorrow: If you want, there are probably enough options to warrant interleaving the comments and the options too.

src/viam/sdk/components/generic/client.hpp
```diff
@@ -33,8 +34,8 @@ class GenericClient : public Generic {
         : Generic(std::move(name)), stub_(std::move(stub)){};
 
    private:
-    std::unique_ptr<viam::component::generic::v1::GenericService::StubInterface> stub_;
-    std::shared_ptr<grpc::Channel> channel_;
+    std::unique_ptr<viam::component::generic::v1::GenericService::StubInterface> stub_{};
+    std::shared_ptr<grpc::Channel> channel_{};
```
acmorrow, line 38: I wouldn't expect this to be needed, the default ctor's for unique and shared pointer should initialize them correctly.

src/viam/sdk/robot/client.cpp
acmorrow: Did you try `size_t`?

acmorrow (approved):
LGTM with a few suggestions, no need to post an update for me.

src/viam/sdk/CMakeLists.txt
acmorrow: I don't think this is quite right. It looks like this is applying something to the include directories for `viamapi` itself, rather than altering how `viamsdk` references the includes it uses to refer to `viamapi`?

Up on line 17 (new) we are already doing some magic to put the `viamapi` includes absolutely first on the include path for `viamsdk` so that we can ensure we point into our generated API stuff if that is in use. I think what should be happening is that that line should be updated to make that a `SYSTEM` include.

acmorrow (changes requested):
Hey, withdrawing that LGTM as I think there is actually a change that needs to be made here.

src/viam/sdk/CMakeLists.txt
stuqdog: So a funny thing came out of testing. It looks like we don't need anything at all for this. If I remove this line entirely and leave line 17 above untouched, we successfully build on linux and mac, despite the proto files continuing to generate warnings. Warning-generating code in `sdk` still causes a compilation failure, as desired.

I'm pushing up the change without any sort of extra special treatment for `api`. Happy to chat more async or in person if you have concerns about this.

acmorrow: Fascinating. I wonder how that comes to be?

src/viam/sdk/CMakeLists.txt
```diff
@@ -159,6 +159,16 @@ target_link_libraries(viamsdk
 )
 
 
+# Set compile and link options based on arguments
```
acmorrow, line 162: Content here looks fine, but can you move this block up above `target_sources`? I feel like setting compile flags is something that should happen early in the sequence of defining the target. So either before or after where we stick the special API include path in.

src/viam/sdk/components/camera/camera.hpp
```diff
@@ -25,6 +25,7 @@ namespace sdk {
 /// @ingroup Camera
 class CameraSubtype : public ResourceSubtype {
    public:
+    virtual ~CameraSubtype() = default;
```
acmorrow, line 28: [nit]: This is equivalent to making the virtual dtor inline. I'd prefer not to do that. I'd move the `=default` into the .cpp file:

In `camera.hpp`:
```
class CameraSubtype ... {
    public:
     virtual ~CameraSubtype();
...
```

In `camera.cpp`:
```
CameraSubtype::~CameraSubtype() = default;
```

No need to change it in this review, but maybe you can enqueue it to your TODOs of small improvements.

src/viam/sdk/tests/test_camera.cpp
```diff
@@ -167,7 +167,7 @@ BOOST_AUTO_TEST_SUITE_END()
 
 BOOST_AUTO_TEST_SUITE(test_camera_client)
 
-MockCameraClient client = MockCameraClient("camera");
+MockCameraClient client("camera");
```
acmorrow, line 170: I like this improvement, and I feel like there are several places in this review where the style is moving to be less prolix with types.

src/viam/sdk/tests/test_utils.cpp
```diff
@@ -11,8 +11,7 @@ namespace sdktests {
 using namespace viam::sdk;
 
 AttributeMap fake_map() {
-    std::shared_ptr<ProtoType> proto_ptr =
-        std::make_shared<ProtoType>(std::move(std::string("hello")));
+    std::shared_ptr<ProtoType> proto_ptr = std::make_shared<ProtoType>(std::string("hello"));
```
acmorrow, line 14: [nit] could `auto` here too like above.

acmorrow (approved):
Nits and small things only, no need to post for re-review for my part.

Regarding the issue of `SYSTEM`-ing `viamapi`, I did some experiments and have concluded that it is not the right thing for us to do.

As it currently stands, the setup here will work because the proto headers don't *currently* generate any warnings.

If they start to in the future, we will need to magic up a different solution to suppress warnings in the API headers. One possibility would be to write a prelude/postlude header pair that used `#pragma diagnostic` or the local equivalent, and then require that within the SDK all inclusions of generated API headers are wrapped in those headers.

src/viam/sdk/components/camera/camera.hpp
stuqdog: This is easy enough to do here, gonna take care of it now so we don't have to keep track.

PR #66, opened 2023-04-04
RSDK-2566 - add from_robot to wrappers

stuqdog:
#### Major Changes ####
- Adds `typed_resource_from_robot()` template function
- Update `dial` example to show that we can successfully get a `GenericComponent` from a robot via the `typed_resource_from_robot()` function

#### Minor Changes ####
- Remove `name_` field from `generic` and `camera`, use `ResourceBase::name_` instead
- Make `name_` and `type_` protected in `ResourceBase`


src/resource/resource_manager.hpp
```diff
@@ -8,7 +8,7 @@
 
 class ResourceManager {
    public:
-    static std::unordered_map<std::string, std::shared_ptr<ResourceBase>> resources;
```
stuqdog, line 11: driveby: this should never have been static, as it made replacing the `resource_manager` during a `refresh` call not actually do anything.

examples/dial/example_dial.cpp
```diff
@@ -33,28 +34,37 @@ int main() {
...
     std::vector<Status> status_plural = robot->get_status();
     std::cout << "Status plural len " << status_plural.size() << std::endl;
     for (Status s : status_plural) {
         std::cout << " Status! " << s.name().subtype() << std::endl;
     }
 
-    std::vector<ResourceName> just_one = std::vector<ResourceName>();
-    just_one.push_back(the_one_we_care_about);
+    // ensure we can send requests for specific resources
+    std::vector<ResourceName> just_one = {resource_names->at(0)};
```
clintpurser, line 57: Just wondering, is there a particular reason you are using `at()` vs the `[]` operator?

I haven't seen `at()` very much, so i'm intrigued.

clintpurser: Ok :) my minor suggestion is to use `[]` - but of course there are tradeoffs! And they both do what you're wanting.

I am not too familiar with `at()`, I read through this [SO question](https://stackoverflow.com/questions/9376049/vectorat-vs-vectoroperator) and it gave me some bearings. Seems like `at()` is a little slower, but will do bounds checking for you.

I've typically seen `[]` used more in C++, in my opinion it's more preferable and idiomatic.

```suggestion
    std::vector<ResourceName> just_one = {resource_names[0]};
```

zaporter-work: I have generally seen `at(..)` used when the underlying type is a pointer (resouce_names is a pointer to a vector) because `(*resource_names)[0]` is uglier.

examples/dial/example_dial.cpp
```diff
@@ -16,9 +16,10 @@
 #include <string>
 #include <vector>
 
-#include "robot/client.hpp"
-#include "robot/service.hpp"
-#include "rpc/dial.hpp"
+#include <components/generic/client.hpp>
+#include <robot/client.hpp>
+#include <robot/service.hpp>
+#include <rpc/dial.hpp>
```
zaporter-work, line 22: [q] Why are we moving to system-first imports?

examples/dial/example_dial.cpp
clintpurser: thats a really good point. I didn't realize that thank you for clarifying, I agree with Zack in that case.

src/services/service_base.cpp
```diff
@@ -7,14 +7,14 @@
...
-ResourceName ServiceBase::get_resource_name(std::string name_) {
+ResourceName ServiceBase::get_resource_name(std::string name) {
     // TODO (RSDK-1631): test, confirm whether we need to split on
     // "viam.components" here
     ResourceName r;
     *r.mutable_namespace_() = RDK;
     *r.mutable_type() = SERVICE;
-    *r.mutable_subtype() = name;
-    *r.mutable_name() = name_;
+    *r.mutable_subtype() = this->name();
```
zaporter-work, line 16: [q] Where is the name set for this ResourceBase?

src/components/component_base.cpp
```diff
@@ -11,14 +11,14 @@
 
 using viam::common::v1::ResourceName;
 
-ResourceName ComponentBase::get_resource_name(std::string name_) {
+ResourceName ComponentBase::get_resource_name(std::string name) {
```
clintpurser, line 14: in this code base what are you having the trailing `_` denote?

zaporter-work (approved):
LGTM % questions

examples/dial/example_dial.cpp
stuqdog: Just needed to pick an idiom and stick to it, really. Everywhere under `src/` is following "first main include, then system headers, then third party, then our proto, then our code" so we should probably do the same here.

src/components/camera/client.cpp
```diff
@@ -48,6 +50,18 @@ Camera::raw_image CameraClient::get_image(std::string mime_type) {
     return from_proto(resp);
 };
 
+std::shared_ptr<CameraClient> CameraClient::from_robot(std::shared_ptr<RobotClient> robot,
+                                                       std::string name) {
+    ResourceName r;
+    *r.mutable_namespace_() = RDK;
+    *r.mutable_type() = COMPONENT;
+    *r.mutable_subtype() = "camera";
+    *r.mutable_name() = std::move(name);
```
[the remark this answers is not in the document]
stuqdog: Nope, I should be doing it in `service_base.cpp` as well, thanks for the catch!

examples/dial/example_dial.cpp
zaporter-work: Talked in person: We are moving to `<..>` imports for everything for stylistic reasons.

src/resource/resource_base.hpp
```diff
@@ -14,11 +14,15 @@ class ResourceBase;
...
-
-    std::string name;
-    ResourceType type;
+    ResourceBase(ResourceType type) : type_(std::move(type)){};
+    static ResourceType base_type;
     virtual grpc::StatusCode stop(std::unordered_map<std::string, ProtoType*> extra);
     virtual grpc::StatusCode stop();
     virtual void reconfigure(Dependencies deps, Resource cfg);
+    virtual std::string name();
+    ResourceType type();
```
benjirewis, line 23: Is there a way to enforce that all subclasses of a `ResourceBase` implement their own `from_robot`?

benjirewis (approved):
LGTM, and the example CPP module works with the RDK again thanks to your fixes with `name`. Just a question.

src/components/component_base.cpp
stuqdog: Good question! Would like to have a trailing `_` for all private data members of classes. Still a WIP getting to that point :)

src/components/component_base.cpp
```diff
@@ -8,17 +8,18 @@
...
     // TODO (RSDK-1631): test, confirm whether we need to split on
     // "viam.components" here
     ResourceName r;
     *r.mutable_namespace_() = RDK;
     *r.mutable_type() = COMPONENT;
-    *r.mutable_subtype() = name;
-    *r.mutable_name() = name_;
+    *r.mutable_subtype() = this->type().to_string();
```
stuqdog, line 21: driveby: this should be the type (e.g. "camera") and not the name (e.g. "camera1").

src/components/component_type.cpp
stuqdog, line 1: driveby: this (and related `service_type`) wasn't doing anything, so fine to delete!

src/resource/resource_base.hpp
stuqdog: update: from speaking with Drew I think we actually have a good solution for this. It breaks a bit from other SDKs (specifically, it keeps the logic within `robot` rather than the specific resources) but should be cleaner, more idiomatic, and enforce method existence. Stand by :)

benjirewis (approved):
Thanks for looking into compile-time enforcing that resources can be fetched from robots. This seems a bit more complex for us now but I think will help in the long run as we write more wrappers.

src/robot/client.hpp
```diff
@@ -86,3 +105,26 @@ class RobotClient {
...
+std::shared_ptr<T> typed_resource_from_robot(const std::shared_ptr<RobotClient> client,
+                                             const std::string& name) {
+    std::string resource_type;
+    if (std::is_base_of<ComponentBase, T>()) {
+        resource_type = COMPONENT;
+    } else if (std::is_base_of<ServiceBase, T>()) {
+        resource_type = SERVICE;
+    } else {
+        throw std::runtime_error("attempted to get unknown resource type from robot");
+    }
```
acmorrow, line 119: I don't see that `resource_type` is actually used anywhere here.

src/robot/client.hpp
```diff
@@ -86,3 +105,26 @@ class RobotClient {
     ResourceManager resource_manager;
     void refresh_every();
 };
+
+template <typename T>
+std::shared_ptr<T> typed_resource_from_robot(const std::shared_ptr<RobotClient> client,
```
acmorrow, line 110: This feels like it should be a member of `RobotClient`. I think it could even be overloaded as `resource_by_name` because it has different arguments:

```
class RobotClient {
pubilc:
    std::shared_ptr<ResourceBase> resource_by_name(ResourceName name, ResourceType type = ...);

    template<typename T>
    std::shared_ptr<T> resource_by_name(const std::string& name);
```

src/components/camera/client.hpp
```diff
@@ -19,8 +20,10 @@ class CameraClient : public Camera {
    protected:
     CameraClient(std::string name);
     std::unique_ptr<viam::component::camera::v1::CameraService::StubInterface> stub_;
-    std::string name_;
 
    private:
     std::shared_ptr<grpc::Channel> channel_;
 };
+
+template std::shared_ptr<CameraClient> typed_resource_from_robot(const std::shared_ptr<RobotClient>,
```
acmorrow, line 28: I don't think you need this explicit instantiation. Though, there might be a reason to do it. I'd expect to see things organized a little differently though if that was what you were after (in particular, I wouldn't expect to see a definition of typed_resource_from_robot in a header).

src/components/camera/client.cpp
```diff
@@ -72,6 +74,10 @@ Camera::properties CameraClient::get_properties() {
 };
 
 CameraClient::CameraClient(std::string name, std::shared_ptr<grpc::Channel> channel_)
-    : channel_(channel_), stub_(viam::component::camera::v1::CameraService::NewStub(channel_)){};
+    : channel_(channel_), stub_(viam::component::camera::v1::CameraService::NewStub(channel_)) {
+    name_ = std::move(name);
```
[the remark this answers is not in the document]
stuqdog: So I was unable to initialize `name_` in the initializer list because it's not a part of `CameraClient`, it's part of `ResourceBase`. Seems like there's a workaround in this [SO](https://stackoverflow.com/questions/2290733/initialize-parents-protected-members-with-initialization-list-c) post, but it seemed not worth it to me because it necessitates several chains of `protected` constructors (`ResourceBase` -> `ComponentBase` -> `Camera` -> CameraClient`)

PR #67, opened 2023-04-04
RSDK-1920 Fix non-graceful shutdown of modules

benjirewis:
[RSDK-1920](https://viam.atlassian.net/browse/RSDK-1920)

Spawns thread in example module to shutdown server when `SIGTERM` or `SIGINT` is received. Refactors `Server` to be wrapped by various services instead of being global.

This change allows the RDK to remove and reconfigure C++ modules.

[RSDK-1920]: https://viam.atlassian.net/browse/RSDK-1920?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/examples/modules/example_module.cpp
+++ b/examples/modules/example_module.cpp
@@ -2,6 +2,7 @@
 #include <component/generic/v1/generic.grpc.pb.h>
 #include <grpcpp/grpcpp.h>
 #include <grpcpp/server_context.h>
+#include <pthread.h>
 #include <robot/v1/robot.pb.h>
 #include <signal.h>
 
@@ -64,35 +65,24 @@ class MyModule : public GenericService::Service, public ComponentBase {
 };
 
 int MyModule::which = 0;
-std::shared_ptr<ModuleService_> my_mod;
-volatile sig_atomic_t shutdown;
-
-void check_for_shutdown() {
-    while (!shutdown) {
-        // Check for shutdown every 100ms.
-        std::this_thread::sleep_for(std::chrono::milliseconds(100));
-    }
-    // Assign my_mod to nullptr and let destructor be called.
-    my_mod = nullptr;
-}
 
 int main(int argc, char** argv) {
     if (argc != 2) {
         throw "need socket path as command line argument";
     }
 
-    // Make sure to register signal handlers for SIGINT and SIGTERM that set a
-    // global shutdown flag. Start a thread to destroy my_mod if global shutdown
-    // flag is set.
-    shutdown = false;
-    signal(SIGINT, [](int) { shutdown = true; });
-    signal(SIGTERM, [](int) { shutdown = true; });
-    std::shared_ptr<std::thread> check_thread = std::make_shared<std::thread>(check_for_shutdown);
-    check_thread->detach();
+    // C++ modules must handle SIGINT and SIGTERM. Make sure to create a sigset
+    // for SIGINT and SIGTERM that can be later awaited in a thread that cleanly
+    // shuts down your module.
+    sigset_t sigset;
+    sigemptyset(&sigset);
+    sigaddset(&sigset, SIGINT);
+    sigaddset(&sigset, SIGTERM);
+    pthread_sigmask(SIG_BLOCK, &sigset, NULL);
 
     Subtype generic = Generic::subtype();
-    my_mod = std::make_shared<ModuleService_>(argv[1]);
     Model m("acme", "demo", "printer");
+
     std::shared_ptr<ModelRegistration> rr = std::make_shared<ModelRegistration>(
         ResourceType("MyModule"),
         generic,
@@ -112,10 +102,25 @@ int main(int argc, char** argv) {
         });
 
     Registry::register_resource(rr);
-    my_mod->add_model_from_registry(generic, m);
 
-    my_mod->start();
-    Server::start();
-    Server::wait();
+    // The `ModuleService_` must outlive the Server, so the declaration order
+    // here matters.
+    auto my_mod = std::make_shared<ModuleService_>(argv[1]);
+    Server server;
+
+    my_mod->add_model_from_registry(&server, generic, m);
+    my_mod->start(&server);
+
+    std::thread signal_wait_thread([&server, &sigset]() {
+        int sig = 0;
+        auto result = sigwait(&sigset, &sig);
+        server.shutdown();
+    });
+
+    server.start();
+    server.wait();
+
+    signal_wait_thread.join();
+
     return 0;
 };
--- a/examples/modules/example_module.cpp
+++ b/examples/modules/example_module.cpp
@@ -112,16 +112,15 @@ int main(int argc, char** argv) {
     my_mod->add_model_from_registry(server, generic, m);
     my_mod->start(server);
 
-    std::thread signal_wait_thread([&server, &sigset]() {
+    std::thread server_thread([&server, &sigset]() {
+        server->start();
         int sig = 0;
         auto result = sigwait(&sigset, &sig);
         server->shutdown();
     });
 
-    server->start();
     server->wait();
-
-    signal_wait_thread.join();
+    server_thread.join();
 
     return 0;
 };
--- a/examples/modules/example_module.cpp
+++ b/examples/modules/example_module.cpp
@@ -107,19 +107,19 @@ int main(int argc, char** argv) {
     // The `ModuleService_` must outlive the Server, so the declaration order
     // here matters.
     auto my_mod = std::make_shared<ModuleService_>(argv[1]);
-    Server server;
+    auto server = std::make_shared<Server>();
 
-    my_mod->add_model_from_registry(&server, generic, m);
-    my_mod->start(&server);
+    my_mod->add_model_from_registry(server, generic, m);
+    my_mod->start(server);
 
     std::thread signal_wait_thread([&server, &sigset]() {
         int sig = 0;
         auto result = sigwait(&sigset, &sig);
-        server.shutdown();
+        server->shutdown();
     });
 
-    server.start();
-    server.wait();
+    server->start();
+    server->wait();
 
     signal_wait_thread.join();
 
--- a/examples/modules/example_module.cpp
+++ b/examples/modules/example_module.cpp
@@ -73,7 +73,8 @@ int main(int argc, char** argv) {
 
     // C++ modules must handle SIGINT and SIGTERM. Make sure to create a sigset
     // for SIGINT and SIGTERM that can be later awaited in a thread that cleanly
-    // shuts down your module.
+    // shuts down your module. pthread_sigmask should be called near the start
+    // of main so that later threads inherit the mask.
     sigset_t sigset;
     sigemptyset(&sigset);
     sigaddset(&sigset, SIGINT);
--- a/examples/modules/example_module.cpp
+++ b/examples/modules/example_module.cpp
@@ -3,7 +3,6 @@
 #include <grpcpp/grpcpp.h>
 #include <grpcpp/server_context.h>
 #include <robot/v1/robot.pb.h>
-#include <signal.h>
 
 #include <components/component_base.hpp>
 #include <components/generic/generic.hpp>
@@ -30,7 +29,6 @@ class MyModule : public GenericService::Service, public ComponentBase {
         std::cout << "config in reconfigure: " << cfg.name << std::endl;
     }
 
-    void signal_handler(int signum);
     std::string name;
     static int which;
     int inner_which;
@@ -67,21 +65,10 @@ class MyModule : public GenericService::Service, public ComponentBase {
 int MyModule::which = 0;
 std::shared_ptr<ModuleService_> my_mod;
 
-void signal_handler(int signum) {
-    my_mod->close();
-}
-
 int main(int argc, char** argv) {
     if (argc != 2) {
         throw "need socket path as command line argument";
     }
-
-    // TODO(RSDK-1920) This is still causing non-graceful shutdown. Figure out why, and fix.
-    struct sigaction sig_handler;
-    sig_handler.sa_handler = signal_handler;
-    sigaction(SIGTERM, &sig_handler, nullptr);
-    sigemptyset(&sig_handler.sa_mask);
-    sig_handler.sa_flags = 0;
 
     Subtype generic = Generic::subtype();
     my_mod = std::make_shared<ModuleService_>(argv[1]);
--- a/examples/modules/example_module.cpp
+++ b/examples/modules/example_module.cpp
@@ -3,6 +3,7 @@
 #include <grpcpp/grpcpp.h>
 #include <grpcpp/server_context.h>
 #include <robot/v1/robot.pb.h>
+#include <signal.h>
 
 #include <components/component_base.hpp>
 #include <components/generic/generic.hpp>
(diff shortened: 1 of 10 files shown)


src/module/service.cpp
```diff
@@ -198,6 +199,10 @@ ModuleService_::ModuleService_(std::string addr) {
     module = std::make_shared<Module>(addr);
 }
 
+void ModuleService_::signal_handler(int signum) {
```
clintpurser, line 202: just wondering, why pass in `signum` if it is then not used?

benjirewis: yeah they can take in an unnamed `int`; thanks, @clintpurser.

stuqdog (approved):
Just to confirm, in testing this successfully closed sockets and shutdown gracefully? Assuming so, lgtm!

cheukt (commented):
looks good, how much work would it be to add a test to validate that we're shutting down gracefully?

benjirewis:
@stuqdog can confirm that the gRPC connection to the parent is closed and the module's gRPC server is shutdown. The module's `.sock` file is not removed, but [it will be removed by the parent](https://github.com/viamrobotics/rdk/blob/main/module/modmanager/manager.go#L457-L462).

@cheukt I'll look into it.

src/module/service.cpp
```diff
@@ -210,6 +215,11 @@ void ModuleService_::start() {
     std::string address = "unix://" + module->addr;
     Server::add_listening_port(address);
 
+    // Call exit(0) on SIGTERM or SIGINT so module can cleanly shutdown with its
+    // destructor.
+    signal(SIGINT, signal_handler);
```
acmorrow, line 220: Similar thoughts as above: library code shouldn't be making decisions about the disposition or handling of signals. That's the responsibility of the program.

src/module/service.cpp
```diff
@@ -198,6 +199,10 @@ ModuleService_::ModuleService_(std::string addr) {
     module = std::make_shared<Module>(addr);
 }
 
+void ModuleService_::signal_handler(int) {
+    exit(0);
```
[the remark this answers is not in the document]
benjirewis: That's a good point; I'll call the destructor directly instead of exiting. For others' [reference](https://en.cppreference.com/w/cpp/utility/program/exit) vis-a-vis `exit` and destructors.

src/module/service.cpp
benjirewis: While our handling of `SIGINT` and `SIGTERM` in CPP modules is more explicit here, Go and Python do the same sort of "swallowing" of the signals and execution of a graceful shutdown within their libraries (the graceful handling of those signals is [a requirement](https://github.com/viamrobotics/rdk/blob/main/module/doc.go#L65) of module design). The example modules in those languages do not do any sort of signal handling and instead rely on the underlying library. I do agree that it's unorthodox to handle these signals within the library, but it also seems like a potentially unnecessary burden to hand over to users. It seems relatively easy for a user to incorrectly handle these signals and shoot themselves in the foot. What are your thoughts @stuqdog ?

stuqdog: I'm inclined to keep the code within the library. While I agree that it's a strange pattern, I think I favor consistency with the other SDKs - and with the `module` spec - in this case. Since the `module` spec specifically says that what is built, I think there's potential for additional confusion beyond what @benjirewis highlighted above if we fail to handle shutdown. Happy to keep the conversation going if there's a strong contrary preference, however.

src/module/service.cpp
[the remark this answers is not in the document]
benjirewis: Calling the destructor directly does not actually seem feasible, as signal handling functions must be `static` and therefore cannot be associated with a particular instance of a module (no access to `this` and no access to the destructor). If we want to handle signals from within the library we'll probably need a more complex solution here with a thread that checks whether `SIGTERM` or `SIGINT` has been called periodically.

src/module/service.cpp
acmorrow: - Regarding the requirement that modules handle signals gracefully, I think that the requirement is good from the perspective of a module as an executable that gets run. But that does*not mean that the SDK used to produce modules must itself automatically handle the signals. As long as the author of each module ensures that signals are properly handled, the requirement is still met, though with some unfortunate complexity pushed to the module author.

- The C++ SDK is not in a position to fully know what should be done on receipt of a signal because it cannot know what requirements *other* pieces of software in use by the module may have. For instance, if you were using the C++ SDK to produce a modular resource that supported the operation of an arm, that arm may come with its own SDK with its own shutdown requirements. Only the author of the module is in a position to coordinate the graceful termination of all of the components in use, of which the Viam C++ SDK is only one.

- It isn't actually legal per the standard to go invoking destructors in a signal handler. In fact, it isn't legal to do more or less anything in a signal handler except call `_Exit` (no dtors run) or `quick_exit` (ditto), or `std::abort` (same). About the only thing you are allowed to do is toggle an atomic flag, and the rules get even crazier in C++17. Even if C++ weren't in play you are still more or less restricted to calling the functions in the "async signal safe" set, which is very limited (https://man7.org/linux/man-pages/man7/signal-safety.7.html). A better way to go about all of this is to use https://man7.org/linux/man-pages/man2/sigwaitinfo.2.html instead of `signal`. But again, that's on the module author, not the SDK.

What you could do, and might make sense, would be to offer a free function in the SDK somewhere that sets up a minimal and correct signal handling environment under the assumption that there are no other shutdown requirements, and then have the example module call that, since in the examples there are no other requirements than tearing down the SDK correctly. That would make it easy for authors of simple modules to obtain the required signal handling behavior with low effort, but not put the SDK in the position of taking on a burden it can't actually handle in the more complex cases.

benjirewis: Thanks for all the info @acmorrow ; it seems like it's neither feasible nor optimal to handle signals from within the SDK. I've moved the signal handling to the example module. Let me know what you think.

benjirewis:
@cheukt We should eventually have an integration test for this functionality. I would not say it's trivial to implement, and we may be better off waiting until we have C++ integration tests in [this repo](https://github.com/viamrobotics/sdk-integration-tests). The README for that repo already says we should consider tests for "Module and module reconfiguration"; the test you're suggesting seems to fall under "module reconfiguration".

stuqdog (approved):
I don't think I quite have the context to speak to best practices here but lgtm

acmorrow:
Given that this example is likely to become the starting point / template for future module development I think it is worth spending a little extra time to get it working the way it ought to. I'll share some additional thoughts soon, hopefully tomorrow AM.

acmorrow:
I started writing an essay describing the changes that I had in mind, but I decided it would be easier to just show:

https://github.com/acmorrow/viam-cpp-sdk/commit/21c694d8540146d3d77ed43a1f5145364bf2b8f0

Basically, I think one of the difficulties in the current approach is the global nature of `Server` and the global nature of `my_mod` which makes the teardown order confusing. Similarly, the asymmetry between the fact that the (global, static) `Server` gets started in `main` but has the `shutdown` method called from `ModuleService_::close`. But `ModuleService` doesn't own the `Server`, so it shouldn't really be managing the lifecycle, in my opinion. Please give it a look and let me know what you think. I didn't do a particularly careful job here because I just wanted to get the idea across.

benjirewis:
@acmorrow thanks for the example code! I've modified it slightly but it works for module reconfiguration (sending SIGTERM to the module, restarting it and re-adding resources works). The concept of adding a constructor and destructor to `Server` and passing the created server into `ModuleService_::start` makes sense to me, but would love your opinion @stuqdog.

acmorrow:
> @acmorrow thanks for the example code! I've modified it slightly but it works for module reconfiguration (sending SIGTERM to the module, restarting it and re-adding resources works). The concept of adding a constructor and destructor to `Server` and passing the created server into `ModuleService_::start` makes sense to me, but would love your opinion @stuqdog.

That was the shortest path to what I wanted to demonstrate, but I'm also not staking out a particularly strong opinion that it is the only way. I think it depends on whether `Server` really represents a re-usable component or not. In theory the `ModuleService_` could also directly manage the lifecycle of the GRPC service, as could `CameraService`. But if there will be many services, I think having a `Server` abstraction probably does make sense. Another thing that would argue for it would be if you might find that you wanted to create a binary that supported multiple Modules simultaneously.

examples/modules/example_module.cpp
```diff
@@ -65,27 +65,24 @@ class MyModule : public GenericService::Service, public ComponentBase {
...
         throw "need socket path as command line argument";
     }
 
-    // TODO(RSDK-1920) This is still causing non-graceful shutdown. Figure out why, and fix.
-    struct sigaction sig_handler;
-    sig_handler.sa_handler = signal_handler;
-    sigaction(SIGTERM, &sig_handler, nullptr);
-    sigemptyset(&sig_handler.sa_mask);
-    sig_handler.sa_flags = 0;
+    // C++ modules must handle SIGINT and SIGTERM. Make sure to create a sigset
```
acmorrow, line 74: The comment is good, but it actually doesn't speak to one of the most important things here, which is the call to `pthread_sigmask` happening as early as possible. It is important to do this more or less at the beginning of main before other threads are created so that the new threads also inherit the mask. I'd suggest updating the comment to include that information to make it clear that the placement is not arbitrary.

examples/modules/example_module.cpp
```diff
@@ -105,10 +102,25 @@ int main(int argc, char** argv) {
...
+    my_mod->add_model_from_registry(&server, generic, m);
+    my_mod->start(&server);
+
+    std::thread signal_wait_thread([&server, &sigset]() {
+        int sig = 0;
+        auto result = sigwait(&sigset, &sig);
+        server.shutdown();
+    });
+
+    server.start();
```
acmorrow, line 120: I think there are some windows for races here, like if the signal arrives after calling `sigwait` but before `server.start` is that OK? I think it is worth considering whether the thread is started at the right place and thinking through the potential ordering issues to make sure everything is well behaved.

examples/modules/example_module.cpp
```diff
@@ -65,27 +65,24 @@ class MyModule : public GenericService::Service, public ComponentBase {
...
-    // TODO(RSDK-1920) This is still causing non-graceful shutdown. Figure out why, and fix.
-    struct sigaction sig_handler;
-    sig_handler.sa_handler = signal_handler;
-    sigaction(SIGTERM, &sig_handler, nullptr);
-    sigemptyset(&sig_handler.sa_mask);
-    sig_handler.sa_flags = 0;
+    // C++ modules must handle SIGINT and SIGTERM. Make sure to create a sigset
+    // for SIGINT and SIGTERM that can be later awaited in a thread that cleanly
+    // shuts down your module.
+    sigset_t sigset;
```
acmorrow, line 77: If you wanted to reduce the amount of boilerplate in the example, you could create a static method on `Server` or `ModuleService` to mask and return a sigset appropriate for the signals the module must handle.

examples/modules/example_module.cpp
```diff
@@ -105,10 +102,25 @@ int main(int argc, char** argv) {
...
-    Server::wait();
+    // The `ModuleService_` must outlive the Server, so the declaration order
+    // here matters.
+    auto my_mod = std::make_shared<ModuleService_>(argv[1]);
+    Server server;
+
+    my_mod->add_model_from_registry(&server, generic, m);
+    my_mod->start(&server);
+
+    std::thread signal_wait_thread([&server, &sigset]() {
```
acmorrow, line 114: You could also make a helper function to create and return this thread somehow.

examples/modules/example_module.cpp
benjirewis: My worry about helper functions for sigset creation and signal-checking-thread creation is that it will make this logic harder to read without significantly reducing the amount of code here. If a user is building any sort of complex module, I have a feeling they'll want to do their own custom signal handling and capture (potentially shutting down custom child processes or destroying custom objects). It might be best to show explicitly how to handle signals entirely within the module and let the user either copy that code exactly or modify it to what they need. What are your thoughts?

examples/modules/example_module.cpp
benjirewis: See my comment above.

examples/modules/example_module.cpp
benjirewis: Great point; elaborated on necessary positioning of this code.

examples/modules/example_module.cpp
benjirewis: > if the signal arrives after calling sigwait but before server.start is that OK?

Good point; I don't think so. `shutdown` is a no-op when the server hasn't been started yet, but the thread is only able to catch a single `SIGTERM` or `SIGWAIT`. Looking into a better ordering here (putting `server.start()` before thread creation causes an `abort trap` error).

examples/modules/example_module.cpp
stuqdog: I think having more boilerplate is fine - or even desirable - in an example. Would rather have slightly wordier code here that spells out explicitly how to do signal handling.

src/components/camera/server.hpp
```diff
@@ -27,7 +27,7 @@ class CameraServer : public ResourceServerBase,
         const ::viam::component::camera::v1::GetPropertiesRequest* request,
         ::viam::component::camera::v1::GetPropertiesResponse* response) override;
 
-    void register_server() override;
+    void register_server(Server* server) override;
```
stuqdog, line 30: For consistency's sake, can we use a smart pointer to `server` here and in generic?

stuqdog (changes requested):

stuqdog:
> > @acmorrow thanks for the example code! I've modified it slightly but it works for module reconfiguration (sending SIGTERM to the module, restarting it and re-adding resources works). The concept of adding a constructor and destructor to `Server` and passing the created server into `ModuleService_::start` makes sense to me, but would love your opinion @stuqdog.
> 
> That was the shortest path to what I wanted to demonstrate, but I'm also not staking out a particularly strong opinion that it is the only way. I think it depends on whether `Server` really represents a re-usable component or not. In theory the `ModuleService_` could also directly manage the lifecycle of the GRPC service, as could `CameraService`. But if there will be many services, I think having a `Server` abstraction probably does make sense. Another thing that would argue for it would be if you might find that you wanted to create a binary that supported multiple Modules simultaneously.

I like the idea of passing around the `server` in general because I like avoiding it as global mutable state. But, my mental model is that there should only be one of each service. I'm not sure what the value is of having multiple `Module`s given that a single `ModuleServer` can manage an arbitrary number of modular resources.

src/components/camera/server.hpp
benjirewis: Switched to `std::shared_ptr<Server>` everywhere.

stuqdog:
To @acmorrow's point about the module owning/managing the `Server`, I wonder if it makes sense for the `Server` to just belong to the `ModuleServer_` explicitly? I'm not sure if there's a case where it would make sense for a `ModuleServer_` to operate on multiple servers, or for multiple `ModuleServer_`s to operate on the same server. So enforcing the 1:1 relationship through explicit ownership might be good.

examples/modules/example_module.cpp
acmorrow: Sounds good to me.

acmorrow (approved):
I'm happy with this when Ethan is.

examples/modules/example_module.cpp
benjirewis: Decided to call `server.start()` at the top of the created thread instead of after thread creation. There are still windows where `SIGTERM` could be sent but not caught (i.e. before `sigwait`), but I think we get rid of the case described above.

benjirewis:
@stuqdog filed [RSDK-2655](https://viam.atlassian.net/browse/RSDK-2655) which I hope is at least a semi-accurate description of your request above. I tried doing it locally in this PR but it wasn't _super_ trivial, and I worry that we're already expanding out of the scope of this ticket.

stuqdog (approved):

PR #71, opened 2023-04-07
Fix uninitialized pointer in dial.cpp

zaporter-work:
Without this fix, the pointer is initialized to a random value and is not nullptr. This causes problems in the rust code that checks if it is null.

To replicate: 
- edit example_dial to not pass a payload. The rust code will return a very non-helpful error.


stuqdog (approved):
lgtm

zaporter-work:
Testing in docker image only -- decided to talk in person.  Possibly a architecture / os / compiler issue.

PR #72, opened 2023-04-10
RSDK-2352 - add doxygen support

stuqdog:
#### Major changes
- Adds `Doxyfile` and a workflow for updating docs
- Creates `apidocmenu.md` as a landing page for docs

#### Minor changes
- Add brief doxygen-formatted documentation to `generic` component.
- (driveby) fix bug in `resource_names_by_resource` whereby we would never return more than a single resource name.


.github/workflows/docs.yml
```diff
@@ -0,0 +1,37 @@
...
+        uses: actions/checkout@v3
+
+      - name: Generate docs
+        run: doxygen Doxyfile
+
+      - name: Upload artifacts
+        uses: actions/upload-artifact@v3
+        with:
+          name: html-docs
+          path: etc/docs/api/current
```
maximpertsov, line 28: discussed IRL - I'm not 100% sure if we need this - @njooma do you know?

maximpertsov (approved):

.github/workflows/docs.yml
njooma: I'm not sure if it's needed either. Did you test without this? Does it work?

stuqdog: I'm not sure! I don't really know how to test this without it, since it's not yet merged so the branch that the job actually runs from doesn't know about this job.

I think the stakes are pretty low here though, so if y'all are comfortable I might just merge and then we can iterate from there if anything isn't working right.

maximpertsov: yeah I'm okay merging as is - I kept it for TS doc generation as well - was more curious than anything else.

PR #75, opened 2023-04-10
Add examples/camera 

zaporter-work:
While working on the other wrappers, I decided to add an example for the camera component. 

I think it would be ideal if all of our examples worked with rdk running the "rdk::etc/config/fake.json" config.

This is not ready for review. I still want to fix a bug with the resulting image being corrupted and a strange memory issue that might be related to this camera impl.


zaporter-work:
Moved to https://github.com/viamrobotics/viam-cpp-sdk/pull/86

PR #78, opened 2023-04-13
RSDK-2703 - Document existing wrappers

stuqdog:
#### Major Changes
- Provides more comprehensive, doxygen-formatted documentation for `robot`, `camera`, `generic`, `rpc/server`, and `registry`

#### Minor changes
- Make use of `should_close_channel` in robot client
- make `threads` `private` in robot client
-

### DIFF
--- a/src/robot/client.hpp
+++ b/src/robot/client.hpp
@@ -30,12 +30,14 @@ using Viam::SDK::Options;
 using Viam::SDK::ViamChannel;
 
 // TODO(RSDK-1742) replace all `ResourceName` references in API with `Name`
+/// @defgroup Robot Classes related to a `Robot` representation.
 
 /// @class RobotClient client.h "robot/client.h"
 /// @brief gRPC client for a robot, to be used for all interactions with a robot.
 /// There are two ways to instantiate a robot:
 ///   - `RobotClient::at_address(...)`
 ///   - `RobotClient::with_channel(...)`
+/// @ingroup Robot
 ///
 /// You must `close()` a robot when finished with it in order to release its resources.
 /// Robots creates via `at_address` will automatically close, but robots created via
--- a/src/robot/client.hpp
+++ b/src/robot/client.hpp
@@ -1,3 +1,6 @@
+/// @file robot/client.cpp
+///
+/// @brief gRPC client implementation for a `robot`.
 #pragma once
 
 #include <string>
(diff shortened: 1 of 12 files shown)


src/robot/client.hpp
```diff
@@ -81,29 +95,53 @@ class RobotClient {
...
                                      std::unordered_map<std::string, std::shared_ptr<ProtoType>>,
                                      ResourceNameHasher,
                                      ResourceNameEqual> extra);
+
+    /// @brief Cancel a specified operation on the robot.
+    /// @param id The ID of the operation to cancel.
     void cancel_operation(std::string id);
 
    private:
+    std::vector<std::shared_ptr<std::thread>> threads_;
```
stuqdog, line 144: (driveby) no one outside of robot client was actually using `threads` so it was trivial to make it private.

src/rpc/server.cpp
```diff
@@ -19,7 +19,7 @@ void Server::register_service(grpc::Service* service) {
 
 void Server::start() {
     if (server_) {
-        throw "Attempted to start server that was already running";
+        throw std::runtime_error("Attempted to start server that was already running");
```
stuqdog, line 22: (driveby) adding documentation, wanted to throw an actual error rather than a string. easy enough to fix!

cheukt (approved):
LGTM - not sure if this pr does it, but is there a way to group all related classes together in the class list page (e.g. Camera client, server, and Camera in one dropdown)

stuqdog:
> LGTM - not sure if this pr does it, but is there a way to group all related classes together in the class list page (e.g. Camera client, server, and Camera in one dropdown)

Good question! This pr does not do that, but I'll do some research into it.

acmorrow:
> > LGTM - not sure if this pr does it, but is there a way to group all related classes together in the class list page (e.g. Camera client, server, and Camera in one dropdown)
> 
> Good question! This pr does not do that, but I'll do some research into it.

One thing that comes to mind is using namespaces or nesting more aggressively. Things overall still need to get wrapped in a top-level namespace, and usually I recommend not using complex namespacing schemes, but in principle you could have a `camera` namespace under whatever is the top-level. What I don't love about that is you end up with `camera::Camera`, which feels chatty. But I do maybe like `camera::Server` and `camera::Client` over `CameraServer` and `CameraClient`?

Another idea might be to nest the `Client` and `Server` under `Camera`:

```
class Camera {
public:
    class Server;
    class Client;
};

class Camera::Server : public ... {
};

class Camera::Client : public Camera {
};
```

The definitions of `Camera::Server` and `Camera::Client` don't even need to be in the same header file once you have forward declared them as nested. On the other hand, doing it this way makes it impossible to forward declare the camera server / client since classes are closed and namespaces are open. So I think namespaces are probably a better choice.

Finally, doxygen itself has some grouping features: https://www.doxygen.nl/manual/grouping.html

stuqdog:
> > > LGTM - not sure if this pr does it, but is there a way to group all related classes together in the class list page (e.g. Camera client, server, and Camera in one dropdown)
> > 
> > 
> > Good question! This pr does not do that, but I'll do some research into it.
> 
> One thing that comes to mind is using namespaces or nesting more aggressively. Things overall still need to get wrapped in a top-level namespace, and usually I recommend not using complex namespacing schemes, but in principle you could have a `camera` namespace under whatever is the top-level. What I don't love about that is you end up with `camera::Camera`, which feels chatty. But I do maybe like `camera::Server` and `camera::Client` over `CameraServer` and `CameraClient`?
> 
> Another idea might be to nest the `Client` and `Server` under `Camera`:
> 
> ```
> class Camera {
> public:
>     class Server;
>     class Client;
> };
> 
> class Camera::Server : public ... {
> };
> 
> class Camera::Client : public Camera {
> };
> ```
> 
> The definitions of `Camera::Server` and `Camera::Client` don't even need to be in the same header file once you have forward declared them as nested. On the other hand, doing it this way makes it impossible to forward declare the camera server / client since classes are closed and namespaces are open. So I think namespaces are probably a better choice.
> 
> Finally, doxygen itself has some grouping features: https://www.doxygen.nl/manual/grouping.html

Thanks for the guidance! Adding `namespace`ing is another quick PR I hope to get out by early next week. Not that I necessarily want to go deep into the "nested namespace" rabbithole, but does Doxygen respect/group `namespace`s by default?

stuqdog:
@cheukt just added `group`s for `camera`, `robot`, and `generic`. The class list page is unchanged, but it'll create a new tab (`modules`) which shows that grouping.

zaporter-work (approved):
LGTM

PR #79, opened 2023-04-13
RSDK-2719 - Namespace all the things

stuqdog:
#### Major changes
- Add consistent `namespace`ing to the whole repo.

### DIFF
--- a/examples/dial/example_dial.cpp
+++ b/examples/dial/example_dial.cpp
@@ -22,16 +22,15 @@
 #include <rpc/dial.hpp>
 
 using viam::robot::v1::Status;
-using Viam::SDK::Credentials;
-using Viam::SDK::Options;
+using namespace viam::cppsdk;
 
 int main() {
     const char* uri = "<your robot URI here>";
-    Viam::SDK::DialOptions dial_options = Viam::SDK::DialOptions();
+    DialOptions dial_options;
     std::string payload = "<your payload here>";
     Credentials credentials(payload);
     dial_options.credentials = credentials;
-    boost::optional<Viam::SDK::DialOptions> opts(dial_options);
+    boost::optional<DialOptions> opts(dial_options);
     std::string address(uri);
     Options options = Options(1, opts);
 
--- a/examples/modules/CMakeLists.txt
+++ b/examples/modules/CMakeLists.txt
@@ -17,6 +17,7 @@ add_executable(example_module
 )
 
 target_link_libraries(example_module
+  PRIVATE Threads::Threads
   viamcpp
 )
 
--- a/src/components/service_base.cpp
+++ b/src/components/service_base.cpp
@@ -1,1 +1,3 @@
 #include <components/service_base.hpp>
+
+using namespace viam::cppsdk;
--- a/src/components/service_base.cpp
+++ b/src/components/service_base.cpp
@@ -1,4 +1,5 @@
 #include <components/service_base.hpp>
 
 namespace viam {
-namespace cppsdk {
+namespace cppsdk {}
+}  // namespace viam
--- a/src/components/service_base.cpp
+++ b/src/components/service_base.cpp
@@ -1,3 +1,4 @@
 #include <components/service_base.hpp>
 
-using namespace viam::cppsdk;
+namespace viam {
+namespace cppsdk {
--- a/src/resource/resource.hpp
+++ b/src/resource/resource.hpp
@@ -6,6 +6,9 @@
 
 #include <common/v1/common.pb.h>
 
+namespace viam {
+namespace cppsdk {
+
 class Type {
    public:
     Type(std::string namespace_, std::string resource_type);
@@ -124,3 +127,6 @@ struct std::hash<Subtype> {
         return std::hash<std::string>()(key.to_string());
     };
 };
+
+}  // namespace cppsdk
+}  // namespace viam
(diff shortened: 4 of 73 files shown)


src/tests/mocks/camera_mocks.hpp
```diff
@@ -8,6 +8,9 @@
 #include <components/camera/client.hpp>
 #include <components/camera/server.hpp>
 
+namespace viam {
+namespace cppsdk {
```
zaporter-work, line 12: Can we namespace this one layer deeper? Like viam::cppsdk::cameramocks? Otherwise functions like `fake_...` can collide when compiling the tests. (I ran into this when adding additional motor tests.)

zaporter-work (approved):
This is beautiful. I had been thinking about doing this myself.

examples/dial/example_dial.cpp
```diff
@@ -22,16 +22,15 @@
 #include <rpc/dial.hpp>
 
 using viam::robot::v1::Status;
-using Viam::SDK::Credentials;
-using Viam::SDK::Options;
+using namespace viam::cppsdk;
```
acmorrow, line 25: This is fine in an example, but see below for how it should be done in SDK source files.

acmorrow, line 25: I want to think a bit about whether `viam` is right at the top-level (I think so), and whether `cppsdk` is the right inner name (I suspect it is).

src/tests/mocks/camera_mocks.hpp
acmorrow: A case can be made that the mocks or other test code doesn't actually belong in the `viam:cppsdk` namespace at all.

I'm also not usually too keen on a complex set of "inner namespaces" due to the fact that namespaces have interactions with name lookup: https://en.cppreference.com/w/cpp/language/adl

acmorrow (changes requested):

acmorrow:
> This is beautiful. I had been thinking about doing this myself.

Agreed I'm very happy to see this happening.

examples/modules/CMakeLists.txt
```diff
@@ -17,6 +17,7 @@ add_executable(example_module
 )
 
 target_link_libraries(example_module
+  PRIVATE Threads::Threads
```
stuqdog, line 20: (driveby) make sure we link threads properly. this was necessary to fix builds on linux.

src/tests/mocks/camera_mocks.cpp
```diff
@@ -8,6 +8,9 @@
 #include <components/camera/server.hpp>
 #include <tests/test_utils.hpp>
 
+namespace viam {
+namespace cppsdktests {
```
zaporter-work, line 12: This still has the problem of declaring functions like `fake_properties` in the namespace for all of the tests but other tests (like motor) will also have `fake_properties` (if it follows the same pattern). Thoughts?

zaporter-work (approved):

src/tests/mocks/camera_mocks.hpp
zaporter-work: This still has the problem of declaring functions like `fake_properties` in the namespace for all of the tests but other tests (like motor) will also have `fake_properties` (if it follows the same pattern). Thoughts?

stuqdog: Ah, good catch. I added an additional layer of namespacing for the `mock`s only, and using the appropriate namespace in associated tests. This should keep generic functions like `fake_properties` separate, and hopefully minimize namespace bloat. @acmorrow given that this is confined to tests my intuition is it's probably okay as far as the "inner namespace" issue goes but I'm curious to your thoughts.

src/components/service_base.cpp
```diff
@@ -1 +1,4 @@
 #include <components/service_base.hpp>
+
+namespace viam {
+namespace cppsdk {
```
acmorrow, line 4: Missing closing namespace `}`'s here. Is this file actually registered in the build system?

src/resource/resource.hpp
```diff
@@ -102,6 +105,10 @@ class Model {
     friend bool operator==(const Model& lhs, const Model& rhs);
 };
 
+}  // namespace cppsdk
+}  // namespace viam
+
+using namespace viam::cppsdk;
```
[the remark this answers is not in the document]
acmorrow: No, you can't do `using namespace viam::cppsdk` in a header, because it defeats the namespacing for consumers (direct or indirect) of that header. Instead, I'd expect you can just fully qualify our types:

```
template<>
struct std::hash<::viam::cppsdk::Name> {
...
```

Please give that a try and LMK if it doesn't work out as you expect for some reason.

acmorrow (changes requested):
Looking good. I think this is ready to go mod the hash customization.

src/components/service_base.cpp
stuqdog: Thanks for the catch, fixed. It looks like we have the header in the `CmakeLists.txt` but not the implementation file (which is probably fine since it's empty). Would rather not touch in this PR, but I actually already made a ticket to look into this file (which I suspect we can get rid of entirely with some light refactoring) [here](https://viam.atlassian.net/browse/RSDK-2718).

src/resource/resource.hpp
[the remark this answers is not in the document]
stuqdog: That works! Change made.

acmorrow (approved):
LGTM. Ship it! If I have some sort of naming epiphany it is a sed away given what is now in place.

PR #81, opened 2023-04-17
RSDK-2740 - Fix robot service

stuqdog:
#### Major Changes
- Add `robot/service.cpp` to `src/CMakeLists.txt`
- Fix compilation issues in `robot/service.cpp`
- (as part of above) make `get_resource_name` a virtual method of `ResourceBase`.

#### Minor Changes
- Refactor to eliminate awkwardly trying to cast a resource into a `{Component|Service}Base` before getting its name.
- change the `extra` param in `stop()` to be an `AttributeMap` to maintain consistency.

@acmorrow it seemed worth trying to sneak this in quickly because it's a pretty small change and is a bug fix. Happy to wait to merge until you're done with the structural work though if you'd prefer.

### DIFF
--- a/src/resource/resource_base.cpp
+++ b/src/resource/resource_base.cpp
@@ -11,7 +11,7 @@
 namespace viam {
 namespace cppsdk {
 
-grpc::StatusCode ResourceBase::stop(std::unordered_map<std::string, ProtoType*> extra) {
+grpc::StatusCode ResourceBase::stop(AttributeMap extra) {
     return stop();
 }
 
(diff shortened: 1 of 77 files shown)


acmorrow:
@stuqdog - No, I think this won't cause much difficulty, please go ahead when the other reviewers are happy with it.

zaporter-work (approved):

src/resource/resource_base.cpp
```diff
@@ -33,5 +33,15 @@ ResourceType ResourceBase::type() {
     return {"resource"};
 }
 
+ResourceName ResourceBase::get_resource_name(std::string name) {
+    ResourceName r;
+    *r.mutable_namespace_() = RDK;
+    *r.mutable_type() = "resource";
```
benjirewis, line 39: [nit] add `const std::string RESOURCE = "resource";` to `src/common/utils.hpp`.

stuqdog: Yeah good catch, updated :)

PR #83, opened 2023-04-20
RSDK-2797 - Allow more args to modules

cheukt:
not strictly necessary but if people are going to use this as an example, this might save some headache in the future


cheukt:
we'll be adding a debug mode at some point in the future, this is sort of to pre-empt backward compatibility issues if people build modules off of the example and don't change the line

maximpertsov (approved):

purplenicole730 (approved):
LGTM

PR #86, opened 2023-04-24
RSDK-2751 Add Example Camera

zaporter-work:
This is part of: https://viam.atlassian.net/browse/RSDK-2751

I want to create a model example that the other components can use to show off proper use.

I think it makes sense for us to have a number of examples at the component/service level for two reasons:
1) It allows customers to find the code example they are looking for faster
2) It makes it easier for us to run manual QA on our code and debug customer issues

Though, we might not want/need examples every component (ex: simple ones like encoder).

Important design decisions:
1) `using namespace std` because I think it is cleaner for examples. Happy to remove if we feel otherwise.
2) Included an `example_config.json` that spins up a `fake` camera so people can easily test locally.
3) Placed info for example in root `README.md`. This is following the current paradigm for examples, but I also think we could create a `src/viam/examples/README.md` that keeps all of the information close to the actual examples and then link to that README. Opinions?
4) The README info does not say how to install RDK. Do we want to include this info? We could add the classic `git clone github.com/.../rdk && make setup && go run web/server/.../main.go` however this is very ugly. We could also direct them to download the static binaries. Looking for feedback here.
5) Added try/catch statements around main blocks that can fail but not every every block that could throw (ex: `get_image` could fail if grpc conn dies). 
6) Didn't use `auto`. I find `auto` in examples frustrating because it forces me to go look elsewhere to figure out what type is actually being returned. Happy to discuss this.

### DIFF
--- a/README.md
+++ b/README.md
@@ -41,22 +41,6 @@ packages.
 Instructions for running examples follows. More robust instructions
 for setting up a project to come.
 
-### Echo Streaming Example
-The echo example communicates with the goutils sample server. It
-demonstrates individual, streamed, and bidirectional communication. To
-test, navigate to your goutils clone and run
-
-``` shell
-go run rpc/examples/echo/server/cmd/main.go
-```
-
-Then, run the `example_echo` program (it will be in the `bin`
-directory of your SDK installation)
-
-``` shell
-ninja install && ./build/install/bin/example_echo
-```
-
 ### Dial Example
 
 If you are connecting to a robot with authentication you will need to
@@ -106,6 +90,21 @@ override and only use direct gRPC calls, you will need to use a
 `.local` address for your robot's uri,
 e.g. `name.xxxx.local.viam.cloud:8080` instead of
 `name.xxxx.viam.cloud`.
+
+### Camera Example
+The camera example demonstrates how to get an image from the a remote camera and save it to a local file.
+
+First, download and start the [rdk](https://github.com/viamrobotics/rdk) with the provided config
+``` shell
+./viam-server-binary --config ./src/viam/examples/camera/example_config.json
+```
+
+Then build the c++ sdk using the instructions below and run
+``` shell
+./build/src/viam/examples/camera/example_camera
+```
+This will save the image from the fake camera to `img.png`.
+
 
 ## Building Documentation Locally for Testing
 The C++ sdk uses [Doxygen](https://www.doxygen.nl/) to generate documentation.
--- a/README.md
+++ b/README.md
@@ -6,10 +6,14 @@
 > is not guaranteed. Breaking changes are likely to occur, and occur
 > often.
 
+## Resources
+* [Documentation](https://cpp.viam.dev)
+* [Examples](https://github.com/viamrobotics/viam-cpp-sdk/tree/main/src/viam/examples)
+
 ## Repository Layout
-- `src` The implementation of the C++ sdk library
-- `src/gen` All the google and viam api autogen files
-- `examples` A list of examples
+- `src/viam/sdk` The implementation of the C++ sdk library
+- `src/viam/api` All the google and viam api autogen files
+- `src/viam/examples` A list of examples
 
 ## Getting Started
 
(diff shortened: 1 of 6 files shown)


README.md
```diff
@@ -41,22 +41,6 @@ packages.
 Instructions for running examples follows. More robust instructions
 for setting up a project to come.
 
-### Echo Streaming Example
-The echo example communicates with the goutils sample server. It
```
zaporter-work, line 45: Driveby. This no longer exists

src/viam/examples/camera/example_camera.cpp
```diff
@@ -0,0 +1,106 @@
...
+    // the part secret) You can find this on app.viam.com
+    Credentials credentials("");
+
+    DialOptions dial_options;
+
+    // If you have credentials, use this to pass them to the robot
+    // dial_options.credentials = credentials;
+
+    // This is for an example. You should **not** have this option enabled in real systems.
+    dial_options.allow_insecure_downgrade = credentials.payload.length() == 0;
```
zaporter-work, line 48: Any thoughts on `= true` vs `= credentials.payload.length() == 0`? I did the latter in case someone accidentally copy&pastes this into prod without reading carefully.

src/viam/examples/camera/example_camera.cpp
```diff
@@ -0,0 +1,106 @@
...
+    // the part secret) You can find this on app.viam.com
+    Credentials credentials("");
+
+    DialOptions dial_options;
+
+    // If you have credentials, use this to pass them to the robot
+    // dial_options.credentials = credentials;
+
+    // This is for an example. You should **not** have this option enabled in real systems.
+    dial_options.allow_insecure_downgrade = credentials.payload.empty();
```
zaporter-work, line 48: Any thoughts on `= true` vs `= credentials.payload.empty()`? I did the latter in case someone accidentally copy&pastes this into prod without reading carefully.

src/viam/examples/camera/example_camera.cpp
zaporter-work: Changed to `.empty()` and moved discussion.

src/viam/examples/camera/example_camera.cpp
```diff
@@ -0,0 +1,106 @@
...
+    // If you want to connect to a remote robot, you need the robot location secret (not
+    // the part secret) You can find this on app.viam.com
+    Credentials credentials("");
+
+    DialOptions dial_options;
+
+    // If you have credentials, use this to pass them to the robot
+    // dial_options.credentials = credentials;
+
+    // This is for an example. You should **not** have this option enabled in real systems.
```
stuqdog, line 47: This seems like a pretty strong claim. Users might want to enable insecure downgrade. And if we want to discourage that, a comment in an example is probably not the best place to express that. I'd either remove this or weaken the claim a bit (e.g., `This is for an example. Care should be taken before exercising this option in production.`, or something to that effect).

src/viam/examples/camera/example_camera.cpp
```diff
@@ -0,0 +1,106 @@
...
+
+#include <viam/sdk/components/camera/camera.hpp>
+#include <viam/sdk/components/camera/client.hpp>
+#include <viam/sdk/components/camera/server.hpp>
+#include <viam/sdk/robot/client.hpp>
+#include <viam/sdk/robot/service.hpp>
+#include <viam/sdk/rpc/dial.hpp>
+
+using namespace viam::sdk;
+using namespace std;
```
stuqdog, line 32: I am not a fan of `using namespace std`, I think it buys us very little here and creates collision risks. Preference is to remove it.

stuqdog (changes requested):

src/viam/examples/camera/example_camera.cpp
```diff
@@ -0,0 +1,106 @@
...
+    string camera_name("camera1");
+
+    cout << "Getting camera: " << camera_name << endl;
+    shared_ptr<CameraClient> camera;
+    try {
+        camera = robot->resource_by_name<CameraClient>(camera_name);
+    } catch (const std::exception& e) {
+        cout << "Failed to find " << camera_name << ". Exiting." << endl;
+        cout << "Exception: " << e.what() << endl;
+        robot->close();
```
acmorrow, line 80: Is it important to call `robot->close()` before exiting? The `RobotClient` is held by shared pointer so the destructor will be called. Does the destructor not call `close`? If not, should it?

acmorrow (changes requested):

src/viam/examples/camera/example_camera.cpp
zaporter-work: Good catch. This was done in `example_dial`. Will remove from both.

src/viam/examples/camera/example_camera.cpp
```diff
@@ -0,0 +1,106 @@
...
+    // This is for an example. You should **not** have this option enabled in real systems.
+    dial_options.allow_insecure_downgrade = credentials.payload.empty();
+
+    // Set the refresh interval of the robot (in seconds) (0 = auto refresh) and the dial options
+    Options options = Options(1, dial_options);
+
+    shared_ptr<RobotClient> robot;
+    try {
+        robot = RobotClient::at_address(robot_address, options);
+        robot->refresh();
```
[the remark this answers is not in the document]
zaporter-work: Good catch. This was done in `example_dial`. Will remove from both.

zaporter-work:
@acmorrow Do you have experience with std::nested_exception or boost::exception? Both seem to provide the ability to wrap exceptions (however they both seem pretty heavy-weight). Exception-wrapping is pretty common in some languages.

Do you have opinions / experience with wrapping exceptions in cpp?

src/viam/examples/camera/example_camera.cpp
```diff
@@ -0,0 +1,106 @@
...
+#include <boost/program_options.hpp>
+#include <grpcpp/channel.h>
+#include <grpcpp/client_context.h>
+#include <grpcpp/grpcpp.h>
+#include <grpcpp/support/status.h>
+#include <unistd.h>
+
+#include <viam/api/common/v1/common.pb.h>
+#include <viam/api/robot/v1/robot.grpc.pb.h>
+#include <viam/api/robot/v1/robot.pb.h>
```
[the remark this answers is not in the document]
zaporter-work: I just removed many dependencies. Will refine to ensure that I am not accidentally indirectly including.

src/viam/examples/camera/example_camera.cpp
```diff
@@ -0,0 +1,106 @@
...
+    string output_file = "img.png";
+    string image_mime_type = "image/png";
+
+    cout << "Getting image from camera " << endl;
+    Camera::raw_image img = camera->get_image(image_mime_type);
+    cout << "Got image of mime type: " << img.mime_type << endl;
+
+    // Depending on how you use the image, you may need to convert the
+    // data to a string
+    string img_data(img.bytes.begin(), img.bytes.end());
```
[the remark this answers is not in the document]
zaporter-work: I had done this originally because there was a memory bug I had been hitting with writing from the vector. In retrospect, it was silly. Thanks for the push to fix this.

src/viam/examples/camera/example_camera.cpp
zaporter-work: Good idea. Will do

acmorrow:
> @acmorrow Do you have experience with std::nested_exception or boost::exception? Both seem to provide the ability to wrap exceptions (however they both seem pretty heavy-weight). Exception-wrapping is pretty common in some languages but I don't know if it is a pattern or an anti-pattern in cpp...

I haven't used either, personally. My instinct is that we should't bother. The SDK exists largely as a presentation layer above the gRPC stuff, so I don't anticipate having particularly deep or complex exception throw/rethrow stacks. Maybe it would be more interesting if gRPC itself was throwing (nested) exceptions, but it isn't.

src/viam/examples/camera/example_camera.cpp
```diff
@@ -91,16 +71,11 @@ int main() {
...
 
-    // Depending on how you use the image, you may need to convert the
-    // data to a string
-    string img_data(img.bytes.begin(), img.bytes.end());
-
     cout << "Getting and saving image to " << output_file << endl;
     ofstream fout;
     fout.open(output_file, std::ios::binary | std::ios::out);
-    fout.write(img_data.c_str(), img_data.length());
+    fout.write((char*)&img.bytes[0], img.bytes.size());
```
acmorrow, line 77: Please use a C++ cast here (static/reinterpret/const) rather than a C style one.

acmorrow (approved):
LGTM for my part.

clintpurser:
removing myself as reviewer, I think andrew has covered things well. 👍

stuqdog (approved):
lgtm!

PR #87, opened 2023-04-25
Remove src/CMakeLists.txt 

zaporter-work:
Removing `src/CMakeLists.txt` means that Github will auto-collapse `src` into `src/viam` like the Python SDK https://github.com/viamrobotics/viam-python-sdk/

I don't think there is any clear benefit of having a `CMakeLists.txt` file at this level and it makes our repo slightly more annoying to navigate.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -343,7 +343,7 @@ else()
 endif()
 
 # Pull in our subdirectories
-add_subdirectory(src)
+add_subdirectory(src/viam)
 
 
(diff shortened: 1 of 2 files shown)


stuqdog (approved):
LGTM

acmorrow:
I suspect this file will return sometime in the future, but I don't mind this for now.

zaporter-work:
Should've added you as a reviewer -- glad you're okay with it.

PR #88, opened 2023-04-26
RSDK-2751 Clean up example motor

zaporter-work:
Ticket: https://viam.atlassian.net/browse/RSDK-2751

This applies the style of `examples/camera` discussed in (https://github.com/viamrobotics/viam-cpp-sdk/pull/86) to `examples/motor`

### DIFF
--- a/src/viam/examples/motor/example_motor.cpp
+++ b/src/viam/examples/motor/example_motor.cpp
@@ -72,7 +72,7 @@ int main() {
             throw;
         }
 
-        // Motors are always turned off if there is no attached client.
+        // Motors are always turned off if there is no attached client so this is expected to be 0
         cout << "Motor power: " << motor->get_power_status().power_pct << endl;
         cout << "Setting motor power to 50%" << endl;
         motor->set_power(0.5);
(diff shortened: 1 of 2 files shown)


src/viam/examples/motor/example_motor.cpp
```diff
@@ -1,67 +1,127 @@
...
+            waittime.tv_nsec = 500000000;  // 500ms
+            nanosleep(&waittime, &remaining);
+            print_motor_position(motor);
+        }
+
+        // Explicitly stop the motor
+        cout << "Stopping motor" << endl;
+        vs::AttributeMap stop_map =
+            std::make_shared<std::unordered_map<std::string, std::shared_ptr<vs::ProtoType>>>();
+        motor->stop(stop_map);
```
zaporter-work, line 116: This is unfortunate.

stuqdog: We do expose an overload of `stop` in `resource_base.cpp` that takes no argument, so we could move this into `motor` and have the argless version just generate a map, hiding the ugliness from the user.

It's also worth noting that the implementation of `stop(...)` in `motor/client.cpp` doesn't actually use the `AttributeMap` currently. So we could just change that to be an override of the argless `stop()` overload.

stuqdog (approved):
LGTM. Left a note on the `AttributeMap` issue, happy to see that change here or in a subsequent PR.

PR #89, opened 2023-04-27
RSDK-2859 - Private data members

stuqdog:
#### Major Changes
- Make all class data members private
- consistently add trailing `_` to private data members
- Provide getters/setters as necessary for accessing/modifying data members

#### Minor changes/flybys
- remove `sub_svc` from existing wrappers, instead rely on `sub_svc_` in `ResourceServerBase`
- use `std::lock_guard` instead of interacting with a `mutex` directly
- move `module` lock into `moduleservice`
- Make more consistent use of `const` and `&` for accessors. 
- Remove unused fields from `Options`
- geometry classes -> structs

### DIFF
--- a/src/viam/sdk/common/proto_type.hpp
+++ b/src/viam/sdk/common/proto_type.hpp
@@ -10,6 +10,35 @@ namespace sdk {
 
 class ProtoType {
    public:
+    ProtoType() {
+        proto_type_ = boost::blank();
+    }
+    ProtoType(bool b) {
+        proto_type_ = b;
+    }
+    ProtoType(std::string s) {
+        proto_type_ = s;
+    }
+    ProtoType(int i) {
+        proto_type_ = i;
+    }
+    ProtoType(double d) {
+        proto_type_ = d;
+    }
+    ProtoType(std::shared_ptr<std::unordered_map<std::string, std::shared_ptr<ProtoType>>> m) {
+        proto_type_ = m;
+    }
+    ProtoType(std::vector<ProtoType*> v) {
+        proto_type_ = v;
+    }
+
+    // TODO: RSDK-2421 consider switching this to a constructor
+    static ProtoType of_value(google::protobuf::Value value);
+
+    google::protobuf::Value proto_value();
+    friend bool operator==(const ProtoType& lhs, const ProtoType& rhs);
+
+   private:
     boost::variant<boost::blank,
                    bool,
                    std::string,
@@ -17,33 +46,7 @@ class ProtoType {
                    double,
                    std::shared_ptr<std::unordered_map<std::string, std::shared_ptr<ProtoType>>>,
                    std::vector<ProtoType*>>
-        proto_type;
-    ProtoType() {
-        proto_type = boost::blank();
-    }
-    ProtoType(bool b) {
-        proto_type = b;
-    }
-    ProtoType(std::string s) {
-        proto_type = s;
-    }
-    ProtoType(int i) {
-        proto_type = i;
-    }
-    ProtoType(double d) {
-        proto_type = d;
-    }
-    ProtoType(std::shared_ptr<std::unordered_map<std::string, std::shared_ptr<ProtoType>>> m) {
-        proto_type = m;
-    }
-    ProtoType(std::vector<ProtoType*> v) {
-        proto_type = v;
-    }
-
-    // TODO: RSDK-2421 consider switching this to a constructor
-    static ProtoType of_value(google::protobuf::Value value);
-
-    google::protobuf::Value proto_value();
+        proto_type_;
 };
 
 std::shared_ptr<std::unordered_map<std::string, std::shared_ptr<ProtoType>>> struct_to_map(
@@ -51,8 +54,6 @@ std::shared_ptr<std::unordered_map<std::string, std::shared_ptr<ProtoType>>> str
 
 google::protobuf::Struct map_to_struct(
     std::shared_ptr<std::unordered_map<std::string, std::shared_ptr<ProtoType>>> dict);
-
-bool operator==(const ProtoType& lhs, const ProtoType& rhs);
 
 }  // namespace sdk
 }  // namespace viam
--- a/src/viam/sdk/components/service_base.cpp
+++ b/src/viam/sdk/components/service_base.cpp
@@ -1,5 +1,9 @@
 #include <viam/sdk/components/service_base.hpp>
 
 namespace viam {
-namespace sdk {}
+namespace sdk {
+const ResourceManager ComponentServiceBase::manager() const {
+    return manager_;
+}
+}  // namespace sdk
 }  // namespace viam
--- a/src/viam/sdk/components/service_base.cpp
+++ b/src/viam/sdk/components/service_base.cpp
@@ -2,7 +2,7 @@
 
 namespace viam {
 namespace sdk {
-const ResourceManager ComponentServiceBase::manager() const {
+const ResourceManager& ComponentServiceBase::manager() const {
     return manager_;
 }
 }  // namespace sdk
--- a/src/viam/sdk/module/handler_map.hpp
+++ b/src/viam/sdk/module/handler_map.hpp
@@ -13,7 +13,7 @@ class HandlerMap_ {
     void add_model(Model model, RPCSubtype subtype);
 
     const viam::module::v1::HandlerMap to_proto() const;
-    static HandlerMap_ from_proto(viam::module::v1::HandlerMap proto);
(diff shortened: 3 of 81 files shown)


src/viam/sdk/resource/resource_server_base.cpp
```diff
@@ -1,5 +1,10 @@
 #include <viam/sdk/resource/resource_server_base.hpp>
 
 namespace viam {
-namespace sdk {}  // namespace sdk
+namespace sdk {
+const std::shared_ptr<SubtypeService>& ResourceServerBase::get_sub_svc() const {
```
zaporter-work, line 5: [nit] It seems like we have decided on `attribute()`, not `get_attribute()` semantics. I think this should be `sub_svc()` to be consistent.

src/viam/sdk/rpc/dial.hpp
```diff
@@ -12,72 +12,69 @@ namespace sdk {
...
-    // should update dial logic to actually care about this
-    bool disable_webrtc;
+    DialOptions()
+        : auth_entity_(boost::none), credentials_(boost::none), allow_insecure_downgrade_(false) {}
+
+    const boost::optional<Credentials>& credentials() const;
+    void set_allow_insecure_downgrade(bool allow);
+    const bool allows_insecure_downgrade() const;
+
+    void set_credentials(boost::optional<Credentials> creds);
```
zaporter-work, line 47: [nit] These functions getters/setters aren't grouped

zaporter-work (approved):
LGTM.

src/viam/sdk/common/proto_type.hpp
```diff
@@ -52,7 +55,5 @@ std::shared_ptr<std::unordered_map<std::string, std::shared_ptr<ProtoType>>> str
 google::protobuf::Struct map_to_struct(
     std::shared_ptr<std::unordered_map<std::string, std::shared_ptr<ProtoType>>> dict);
 
-bool operator==(const ProtoType& lhs, const ProtoType& rhs);
```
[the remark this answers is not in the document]
stuqdog: I moved it into the `ProtoType` class as a friend operator.

src/viam/sdk/module/handler_map.hpp
```diff
@@ -11,10 +11,12 @@ class HandlerMap_ {
    public:
     HandlerMap_();
     void add_model(Model model, RPCSubtype subtype);
-    std::unordered_map<RPCSubtype, std::vector<Model>> handles;
 
-    viam::module::v1::HandlerMap to_proto();
+    const viam::module::v1::HandlerMap to_proto() const;
```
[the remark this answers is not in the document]
stuqdog: good catch, will update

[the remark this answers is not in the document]
acmorrow: FYI returning a `const x` from a function doesn't constrain the caller in any way since they implicitly get a copy. There are clang-tidy rules to warn about this. I recommend dropping the first `const` here and in any similar decls.

src/viam/sdk/module/service.cpp
```diff
@@ -54,36 +54,36 @@ Dependencies get_dependencies(ModuleService_* m,
...
 
 ::grpc::Status ModuleService_::AddResource(::grpc::ServerContext* context,
                                            const ::viam::module::v1::AddResourceRequest* request,
                                            ::viam::module::v1::AddResourceResponse* response) {
     viam::app::v1::ComponentConfig proto = request->config();
     Resource cfg = Resource::from_proto(proto);
-    std::shared_ptr<Module> module = this->module;
-    module->lock.lock();
+    std::shared_ptr<Module> module = this->module_;
+    const std::lock_guard<std::mutex> lock(lock_);
```
acmorrow, line 70: 👍

src/viam/sdk/registry/registry.cpp
```diff
@@ -82,8 +92,12 @@ Status ModelRegistration::create_status(std::shared_ptr<ResourceBase> resource)
     return status;
 }
 
-std::unordered_map<std::string, std::shared_ptr<ModelRegistration>> Registry::resources;
-std::unordered_map<Subtype, std::shared_ptr<ResourceSubtype>> Registry::subtypes;
+const google::protobuf::ServiceDescriptor* ResourceSubtype::service_descriptor() {
```
acmorrow, line 95: On the other hand, this const return *is* meaningful because the const-ness is about the object pointed to.

src/viam/sdk/resource/resource.cpp
```diff
@@ -60,38 +60,46 @@ Subtype::Subtype(std::string namespace_, std::string resource_type, std::string
...
+        return subtype_name + "/" + name_;
     }
-    return subtype_name + "/" + remote_name + ":" + name;
+    return subtype_name + "/" + remote_name_ + ":" + name_;
 }
 
-std::string Name::short_name() const {
-    if (remote_name != "") {
-        return remote_name + ":" + name;
+const std::string Name::short_name() const {
```
acmorrow, line 102: This is another `const` that doesn't have the intended effect.

src/viam/sdk/rpc/dial.hpp
```diff
@@ -12,72 +12,69 @@ namespace sdk {
...
-    /// Bypass webRTC and connect directly to the robot
-    // TODO (RSDK-917): This field is currently just for show, we
-    // should update dial logic to actually care about this
-    bool disable_webrtc;
+    DialOptions()
+        : auth_entity_(boost::none), credentials_(boost::none), allow_insecure_downgrade_(false) {}
+
+    const boost::optional<Credentials>& credentials() const;
+    void set_allow_insecure_downgrade(bool allow);
+    const bool allows_insecure_downgrade() const;
```
acmorrow, line 44: Not a meaningful `const bool` here.

src/viam/sdk/rpc/dial.hpp
```diff
@@ -12,72 +12,69 @@ namespace sdk {
...
    public:
-    /// Bypass webRTC and connect directly to the robot
-    // TODO (RSDK-917): This field is currently just for show, we
-    // should update dial logic to actually care about this
-    bool disable_webrtc;
+    DialOptions()
+        : auth_entity_(boost::none), credentials_(boost::none), allow_insecure_downgrade_(false) {}
+
+    const boost::optional<Credentials>& credentials() const;
+    void set_allow_insecure_downgrade(bool allow);
```
acmorrow: A fun trick with options types like this is to make the setters return "ref to self":

```
class FooOptions {
public:
    FooOptions& set_x(int x);
    FooOptions& set_y(int y);
};
```

This allows a caller to chain:

```
FooOptions fo;
fo.set_x(42).set_y(4.2);
```

Some people like it, some don't. Not suggesting a change here, but should these options become more complex it can be a nice thing to do for users.

src/viam/sdk/rpc/dial.hpp
```diff
@@ -12,72 +12,69 @@ namespace sdk {
...
-    }
+    bool allow_insecure_downgrade_;
 };
 
 class Options {
    public:
+    Options(unsigned int refresh_interval, boost::optional<DialOptions> dial_options)
+        : refresh_interval_(std::move(refresh_interval)), dial_options_(std::move(dial_options)) {}
+
+    const unsigned int refresh_interval() const;
```
acmorrow, line 70: Another const value return here.

acmorrow (changes requested):
This looks great. Thanks for doing this. A few small comments here and there. The biggest thing I saw is that you ahve some places where you are returning a const value, but that doesn't do what it looks like it does: the caller gets a copy so is free to mutate it.

src/viam/sdk/components/service_base.cpp
```diff
@@ -1,5 +1,9 @@
 #include <viam/sdk/components/service_base.hpp>
 
 namespace viam {
-namespace sdk {}
+namespace sdk {
+const ResourceManager& ComponentServiceBase::manager() const {
```
stuqdog, line 5: @acmorrow if I understand correctly, this `const` is still meaningful since it prevents the underlying data being referred from being changed. Is this correct?

acmorrow (approved):
LGTM

src/viam/sdk/resource/resource_server_base.cpp
zaporter-work: @stuqdog Is there a reason you didn't do this swap?

stuqdog: nope! slipped the mind when I was updating, apologies :) we can add it as a flyby in a future feature though!

PR #91, opened 2023-04-28
RSDK-2718 - factor out component/service_base

stuqdog:
#### Major Changes
- Removes resource_manager.cpp
- Removes components/service_base.cpp
- Makes consistent use of `subtype_service` in lieu of the above.

This change helps set up a better testing story for robot tests, coming next.


cheukt:
I can see why we'd want to do this - but subtype_service would be a pretty bad name for what we're using it for. we'd need to change it in this pr as well

stuqdog:
> I can see why we'd want to do this - but SubtypeService would be a pretty bad name for what we're using it for. we'd need to change it in this pr as well

Hah yeah, totally valid. I renamed to `ResourceManager`.

@zaporter-work per convo in previous PR, note that `get_sub_svc()` is now `resource_manager()`

src/viam/sdk/resource/resource_manager.cpp
```diff
@@ -1,58 +1,156 @@
...
-ResourceManager::ResourceManager(std::vector<std::shared_ptr<ResourceBase>> resources) {
-    for (auto& resource : resources) {
-        register_resource(resource);
+    if (short_names_.find(name) != short_names_.end()) {
+        std::string short_name = short_names_.at(name);
+        if (resources_.find(short_name) != resources_.end()) {
+            return resources_.at(short_name);
+        }
     }
+    return nullptr;
```
zaporter-work, line 39: [nit] This seems like we should throw an exception here. Otherwise we have a lot of code that directly dereferences something that could be nullptr.

src/viam/sdk/resource/resource_manager.hpp
```diff
@@ -1,33 +1,38 @@
...
+    void remove(Name name);
+    void replace_one(Name name, std::shared_ptr<ResourceBase> resource);
     const std::unordered_map<std::string, std::shared_ptr<ResourceBase>>& resources() const;
+    ResourceManager(){};
 
    private:
-    ResourceManager(std::vector<std::shared_ptr<ResourceBase>> resources);
+    std::mutex lock_;
     std::unordered_map<std::string, std::shared_ptr<ResourceBase>> resources_;
+    std::unordered_map<std::string, std::string> short_names_;
```
zaporter-work, line 32: [nit] Can you add a brief comment here explaining what a short name is

zaporter-work (approved):
LGTM

src/viam/sdk/resource/resource_manager.hpp
stuqdog: Yep! Also added more documentation to `resource_manager.hpp` in general.

cheukt (approved):

PR #92, opened 2023-04-28
RSDK-2484 Add encoder wrapper

zaporter-work:
Ticket: https://viam.atlassian.net/browse/RSDK-2484

Important design decisions:
1) Unsure if Encoder::position_type should be a subclass of Encoder::postion
2) I used peeking-based mocks because it would be difficult to do a behavior-based mock for this component.

Pre-merge:
Bring in refactoring changes. Right now there is so much changing that it would be a waste of time to try and keep up before it is stable.

### DIFF
--- a/src/viam/sdk/CMakeLists.txt
+++ b/src/viam/sdk/CMakeLists.txt
@@ -44,8 +44,8 @@ target_sources(viamsdk
     components/camera/camera.cpp
     components/camera/client.cpp
     components/camera/server.cpp
-    components/encoder/encoder.cpp
     components/encoder/client.cpp
+    components/encoder/encoder.cpp
     components/encoder/server.cpp
     components/component_base.cpp
     components/generic/client.cpp
@@ -86,8 +86,8 @@ target_sources(viamsdk
       ../../viam/sdk/components/camera/camera.hpp
       ../../viam/sdk/components/camera/client.hpp
       ../../viam/sdk/components/camera/server.hpp
-      ../../viam/sdk/components/encoder/encoder.hpp
       ../../viam/sdk/components/encoder/client.hpp
+      ../../viam/sdk/components/encoder/encoder.hpp
       ../../viam/sdk/components/encoder/server.hpp
       ../../viam/sdk/components/component_base.hpp
       ../../viam/sdk/components/generic/client.hpp
--- a/src/viam/sdk/CMakeLists.txt
+++ b/src/viam/sdk/CMakeLists.txt
@@ -44,6 +44,9 @@ target_sources(viamsdk
     components/camera/camera.cpp
     components/camera/client.cpp
     components/camera/server.cpp
+    components/encoder/encoder.cpp
+    components/encoder/client.cpp
+    components/encoder/server.cpp
     components/component_base.cpp
     components/generic/client.cpp
     components/generic/generic.cpp
@@ -72,6 +75,7 @@ target_sources(viamsdk
     tests/test_utils.cpp
     tests/mocks/camera_mocks.cpp
     tests/mocks/generic_mocks.cpp
+    tests/mocks/mock_encoder.cpp
     tests/mocks/mock_motor.cpp
   PUBLIC FILE_SET viamsdk_public_includes TYPE HEADERS
     BASE_DIRS
@@ -82,6 +86,9 @@ target_sources(viamsdk
       ../../viam/sdk/components/camera/camera.hpp
       ../../viam/sdk/components/camera/client.hpp
       ../../viam/sdk/components/camera/server.hpp
+      ../../viam/sdk/components/encoder/encoder.hpp
+      ../../viam/sdk/components/encoder/client.hpp
+      ../../viam/sdk/components/encoder/server.hpp
       ../../viam/sdk/components/component_base.hpp
       ../../viam/sdk/components/generic/client.hpp
       ../../viam/sdk/components/generic/generic.hpp
@@ -111,6 +118,7 @@ target_sources(viamsdk
       ../../viam/sdk/tests/test_utils.hpp
       ../../viam/sdk/tests/mocks/camera_mocks.hpp
       ../../viam/sdk/tests/mocks/generic_mocks.hpp
+      ../../viam/sdk/tests/mocks/mock_encoder.hpp
       ../../viam/sdk/tests/mocks/mock_motor.hpp
 )
 
--- a/src/viam/sdk/components/encoder/client.hpp
+++ b/src/viam/sdk/components/encoder/client.hpp
@@ -1,0 +1,38 @@
+/// @file components/encoder/client.hpp
+///
+/// @brief Implements a gRPC client for the `Encoder` component.
+#pragma once
+
+#include <grpcpp/channel.h>
+
+#include <viam/api/component/encoder/v1/encoder.grpc.pb.h>
+
+#include <viam/sdk/components/encoder/encoder.hpp>
+#include <viam/sdk/components/encoder/server.hpp>
+#include <viam/sdk/config/resource.hpp>
+#include <viam/sdk/robot/client.hpp>
+
+namespace viam {
+namespace sdk {
+
+/// @class EncoderClient
+/// @brief gRPC client implementation of a `Encoder` component.
+/// @ingroup Encoder
+class EncoderClient : public Encoder {
+   public:
+   position get_position(position_type position_type) override;
+   void reset_position() override;
+   properties get_properties() override;
+   AttributeMap do_command(ERROR TODO) override;
+    EncoderClient(std::string name, std::shared_ptr<grpc::Channel> channel)
+        : Encoder(std::move(name)),
+          stub_(viam::component::encoder::v1::EncoderService::NewStub(channel)),
+          channel_(std::move(channel)){};
+
+   private:
+    std::unique_ptr<viam::component::encoder::v1::EncoderService::StubInterface> stub_;
+    std::shared_ptr<grpc::Channel> channel_;
+};
+
+}  // namespace sdk 
+}  // namespace viam
--- a/src/viam/sdk/components/encoder/client.hpp
+++ b/src/viam/sdk/components/encoder/client.hpp
@@ -20,14 +20,12 @@ namespace sdk {
 /// @ingroup Encoder
 class EncoderClient : public Encoder {
    public:
+    EncoderClient(std::string name, std::shared_ptr<grpc::Channel> channel);
+
     position get_position(position_type position_type) override;
     void reset_position() override;
     properties get_properties() override;
     AttributeMap do_command(AttributeMap command) override;
-    EncoderClient(std::string name, std::shared_ptr<grpc::Channel> channel)
-        : Encoder(std::move(name)),
-          stub_(viam::component::encoder::v1::EncoderService::NewStub(channel)),
-          channel_(std::move(channel)){};
 
    private:
     std::unique_ptr<viam::component::encoder::v1::EncoderService::StubInterface> stub_;
--- a/src/viam/sdk/components/encoder/client.hpp
+++ b/src/viam/sdk/components/encoder/client.hpp
@@ -20,10 +20,10 @@ namespace sdk {
 /// @ingroup Encoder
 class EncoderClient : public Encoder {
    public:
-   position get_position(position_type position_type) override;
-   void reset_position() override;
-   properties get_properties() override;
-   AttributeMap do_command(ERROR TODO) override;
+    position get_position(position_type position_type) override;
+    void reset_position() override;
+    properties get_properties() override;
+    AttributeMap do_command(AttributeMap command) override;
     EncoderClient(std::string name, std::shared_ptr<grpc::Channel> channel)
         : Encoder(std::move(name)),
           stub_(viam::component::encoder::v1::EncoderService::NewStub(channel)),
@@ -34,5 +34,5 @@ class EncoderClient : public Encoder {
     std::shared_ptr<grpc::Channel> channel_;
 };
 
-}  // namespace sdk 
+}  // namespace sdk
 }  // namespace viam
--- a/src/viam/sdk/components/encoder/encoder.cpp
+++ b/src/viam/sdk/components/encoder/encoder.cpp
@@ -38,58 +38,46 @@ Subtype Encoder::subtype() {
     return Subtype(RDK, COMPONENT, "encoder");
 }
 
-
-Encoder::position Encoder::from_proto(viam::component::encoder::v1::GetPositionResponse proto){
+Encoder::position Encoder::from_proto(viam::component::encoder::v1::GetPositionResponse proto) {
     Encoder::position position;
     position.value = proto.value();
-    
-    position.position_type = proto.position_type();
+
+    position.type = proto.position_type();
     return position;
 }
 
-Encoder::properties Encoder::from_proto(viam::component::encoder::v1::GetPropertiesResponse proto){
+Encoder::properties Encoder::from_proto(viam::component::encoder::v1::GetPropertiesResponse proto) {
     Encoder::properties properties;
     properties.ticks_count_supported = proto.ticks_count_supported();
-    
+
     properties.angle_degrees_supported = proto.angle_degrees_supported();
     return properties;
 }
 
-
-
-
-viam::component::encoder::v1::GetPositionResponse Encoder::to_proto( position position) {
+viam::component::encoder::v1::GetPositionResponse Encoder::to_proto(position position) {
     viam::component::encoder::v1::GetPositionResponse proto;
     proto.set_value(position.value);
-    
+
     proto.set_position_type(position.position_type);
     return proto;
 }
 
-
-viam::component::encoder::v1::GetPropertiesResponse Encoder::to_proto( properties properties) {
+viam::component::encoder::v1::GetPropertiesResponse Encoder::to_proto(properties properties) {
     viam::component::encoder::v1::GetPropertiesResponse proto;
     proto.set_ticks_count_supported(properties.ticks_count_supported);
-    
+
     proto.set_angle_degrees_supported(properties.angle_degrees_supported);
     return proto;
 }
 
-
-
 bool operator==(const Encoder::position& lhs, const Encoder::position& rhs) {
-    return (
-        lhs.value == rhs.value && 
-        lhs.position_type == rhs.position_type);
+    return (lhs.value == rhs.value && lhs.position_type == rhs.position_type);
 }
 
 bool operator==(const Encoder::properties& lhs, const Encoder::properties& rhs) {
-    return (
-        lhs.ticks_count_supported == rhs.ticks_count_supported && 
-        lhs.angle_degrees_supported == rhs.angle_degrees_supported);
+    return (lhs.ticks_count_supported == rhs.ticks_count_supported &&
+            lhs.angle_degrees_supported == rhs.angle_degrees_supported);
 }
-
-
 
 namespace {
 bool init() {
--- a/src/viam/sdk/components/encoder/encoder.cpp
+++ b/src/viam/sdk/components/encoder/encoder.cpp
@@ -1,0 +1,104 @@
+#include <viam/sdk/components/encoder/encoder.hpp>
+
+#include <google/protobuf/descriptor.h>
+
+#include <viam/api/component/encoder/v1/encoder.grpc.pb.h>
+#include <viam/api/component/encoder/v1/encoder.pb.h>
+
+#include <viam/sdk/common/utils.hpp>
+#include <viam/sdk/components/encoder/client.hpp>
+#include <viam/sdk/components/encoder/server.hpp>
+#include <viam/sdk/registry/registry.hpp>
+#include <viam/sdk/resource/resource.hpp>
+
+namespace viam {
+namespace sdk {
+
+std::shared_ptr<ResourceServerBase> EncoderSubtype::create_resource_server(
+    std::shared_ptr<SubtypeService> svc) {
+    return std::make_shared<EncoderServer>(svc);
+};
+
+std::shared_ptr<ResourceBase> EncoderSubtype::create_rpc_client(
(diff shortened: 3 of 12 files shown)


src/viam/sdk/components/encoder/encoder.hpp
```diff
@@ -0,0 +1,114 @@
...
+        // Provided by absolute encoders
+        ANGLE_DEGREES
+    };
+
+    /// @struct position
+    /// @brief reported position.
+    struct position {
+        float value;
+        position_type type;
+    };
```
zaporter-work, line 57: React :+1:  if you think this is good as it is. React :-1: if you think `position_type` should be a subtype of this class (unsure what the member name would become but can solve that later)

src/viam/sdk/components/encoder/encoder.cpp
```diff
@@ -0,0 +1,125 @@
...
+std::shared_ptr<ResourceBase> EncoderSubtype::create_rpc_client(
+    std::string name, std::shared_ptr<grpc::Channel> chan) {
+    return std::make_shared<EncoderClient>(std::move(name), std::move(chan));
+};
+
+std::shared_ptr<ResourceSubtype> Encoder::resource_subtype() {
+    const google::protobuf::DescriptorPool* p = google::protobuf::DescriptorPool::generated_pool();
+    const google::protobuf::ServiceDescriptor* sd =
+        p->FindServiceByName(viam::component::encoder::v1::EncoderService::service_full_name());
+    if (sd == nullptr) {
```
stuqdog, line 31: (nit) can we change here and elsewhere in this pr as necessary (not sure if we need to update cryvex as well) to `(if !sd)`?

src/viam/sdk/components/encoder/server.cpp
```diff
@@ -0,0 +1,110 @@
...
+    std::shared_ptr<Encoder> encoder = std::dynamic_pointer_cast<Encoder>(rb);
+    AttributeMap result = encoder->do_command(struct_to_map(request->command()));
+
+    *response->mutable_result() = map_to_struct(result);
+
+    return ::grpc::Status();
+}
+
+void EncoderServer::register_server(std::shared_ptr<Server> server) {
+    viam::component::encoder::v1::EncoderService::Service* encoder =
```
stuqdog, line 100: (nit) This cast isn't necessary, I suspect `cryvex` needs to be updated?

src/viam/sdk/components/encoder/server.hpp
```diff
@@ -0,0 +1,52 @@
...
+        const ::viam::component::encoder::v1::GetPropertiesRequest* request,
+        ::viam::component::encoder::v1::GetPropertiesResponse* response) override;
+
+    ::grpc::Status DoCommand(grpc::ServerContext* context,
+                             const viam::common::v1::DoCommandRequest* request,
+                             viam::common::v1::DoCommandResponse* response) override;
+
+    void register_server(std::shared_ptr<Server> server) override;
+
+    std::shared_ptr<SubtypeService> resource_manager();
```
stuqdog, line 42: Just to confirm: has `cryvex` been updated to not include `ResourceManager`/`resource_manager()` in server code in the future?

stuqdog (approved):
LGTM with a couple nits and a q.

src/viam/sdk/components/encoder/encoder.cpp
zaporter-work: Good catch. Forgot about this. Will update in other pr as well.

src/viam/sdk/components/encoder/server.hpp
zaporter-work: I did this because of this comment https://github.com/viamrobotics/viam-cpp-sdk/pull/91#issuecomment-1528082450 . This was a temporary solution to make it so that updating this class post #91 is much easier.

src/viam/sdk/components/encoder/encoder.hpp
```diff
@@ -0,0 +1,114 @@
...
+/// @ingroup Encoder
+///
+/// This acts as an abstract base class to be inherited from by any drivers representing
+/// specific encoder implementations. This class cannot be used on its own.
+class Encoder : public ComponentBase {
+   public:
+    /// @enum position_type
+    enum position_type {
+        // Unspecified position type
+        UNSPECIFIED,
```
acmorrow, line 45: Is it meaningful to reflect the UNSPECIFIED value of the underlying proto here? If the real values that are supported by the encoder are TICKS_COUNT and ANGLE_DEGREES, should it be possible in the C++ API for anything to meaningfully be UNSPECIFIED? Or should an error happen somewhere if UNSPECIFIED is obtained over the wire, and therefore the programmatic API has no need of UNSPECIFIED?

src/viam/sdk/components/encoder/encoder.hpp
```diff
@@ -0,0 +1,114 @@
...
+/// @class Encoder encoder.hpp "components/encoder/encoder.hpp"
+/// @brief An encoder is a device that is hooked up to motors to report a position
+/// @ingroup Encoder
+///
+/// This acts as an abstract base class to be inherited from by any drivers representing
+/// specific encoder implementations. This class cannot be used on its own.
+class Encoder : public ComponentBase {
+   public:
+    /// @enum position_type
+    enum position_type {
```
acmorrow, line 43: This is an un-scoped enum, so its names leak into the enclosing scope, in this case the `Encoder` class. So that means that `Encoder::ANGLE_DEGREES` is a meaningful symbol. There is nothing wrong with that if that is what you want, but it might be good to put `POSITION_TYPE_`  in before each value if you want to do that:

```
class Encoder ... {
public:
    enum position_type {
        POSITION_TYPE_TICKS_COUNT,
        POSITION_TYPE_ANGLE_DEGREES,
    };
```

That way `Encoder::POSITION_TYPE_ANGLE_DEGREES` is clearly a position type.

Alternatively, you can make it an `enum class`, and then the prefix is the actual typename of the `enum` and the values don't leak into the surrounding scope:

```
class Encoder ... {
public:
    enum class position_type {
        TICKS_COUNT,
        ANGLE_DEGREES
    }
```

Means you get `Encoder::position_type::TICKS_COUNT`.

So I recommend either going to an `enum class` or prefixing the enumeration values.

acmorrow, line 43: It is a good practice when writing enums for something which may someday have an ABI to explicitly declare the width of the enum (by giving it a fixed width integral base type) and explicitly provide values for the enumeration.

src/viam/sdk/components/encoder/encoder.hpp
```diff
@@ -0,0 +1,114 @@
...
+        // Unspecified position type
+        UNSPECIFIED,
+        // Provided by incremental encoders
+        TICKS_COUNT,
+        // Provided by absolute encoders
+        ANGLE_DEGREES
```
acmorrow, line 49: Just whining here, but I'd love to switch the convention for constants (including enumerator values) in this project away from all-caps `CONSTANTS`. The capitalization of constants is from the old C days when const-ness wasn't a thing and the preprocessor was a major player. https://accu.org/journals/overload/22/121/wakely_1923/. Personally I like the Google style guide convention of prefixing constants with `k`, so `k_ticks_count` or `k_angles_degrees` because I do find it useful to be able to tell at a glance that a given symbol is a constant. But the all-caps is definitely distracting.

src/viam/sdk/components/encoder/encoder.hpp
```diff
@@ -0,0 +1,114 @@
...
+    /// @struct position
+    /// @brief reported position.
+    struct position {
+        float value;
+        position_type type;
+    };
+
+    /// @struct properties
+    /// @brief Encodes the supported modes of this encoder
+    struct properties {
```
acmorrow, line 61: Are all states of this struct meaningful / possible? Or only two?

src/viam/sdk/components/encoder/encoder.hpp
```diff
@@ -0,0 +1,114 @@
...
+
+    /// @brief Converts a `position` struct to its proto representation.
+    static viam::component::encoder::v1::GetPositionResponse to_proto(position position);
+
+    /// @brief Converts a `properties` struct to its proto representation.
+    static viam::component::encoder::v1::GetPropertiesResponse to_proto(properties properties);
+
+    /// @brief Returns position of the encoder which can either be ticks since last zeroing for an
+    /// incremental encoder or degrees for an absolute encoder.
+    /// @param position_type If supplied, the response will return the specified position type. If
```
acmorrow, line 90: The parameter is always supplied though.

src/viam/sdk/components/encoder/encoder.hpp
```diff
@@ -0,0 +1,114 @@
...
+    /// the driver does not implement the requested type, this call will return an error. If
+    /// position type is UNSPECIFIED, the response will return a default according to the driver.
```
acmorrow, line 92: Rather than passing `unspecified`, could you take an `optional<position_type>`?

src/viam/sdk/components/encoder/encoder.hpp
```diff
@@ -0,0 +1,114 @@
...
+    /// @brief Returns a list of all the position_types that are supported by the encoder.
+    virtual properties get_properties() = 0;
+
+    /// @brief Send/receive arbitrary commands to the resource.
+    /// @param Command the command to execute.
+    /// @return The result of the executed command.
+    virtual AttributeMap do_command(AttributeMap command) = 0;
+
+   protected:
+    explicit Encoder(std::string name) : ComponentBase(std::move(name)){};
```
acmorrow, line 107: Let's define this out of line in the .cpp file.

src/viam/sdk/components/encoder/encoder.hpp
```diff
@@ -0,0 +1,114 @@
...
+    static properties from_proto(viam::component::encoder::v1::GetPropertiesResponse proto);
+
+    /// @brief Converts a `position` struct to its proto representation.
+    static viam::component::encoder::v1::PositionType to_proto(position_type position_type);
+
+    /// @brief Converts a `position` struct to its proto representation.
+    static viam::component::encoder::v1::GetPositionResponse to_proto(position position);
+
+    /// @brief Converts a `properties` struct to its proto representation.
+    static viam::component::encoder::v1::GetPropertiesResponse to_proto(properties properties);
```
acmorrow, line 86: Are these a necessary part of the `Encoder` class API? Could they be declared/defined somewhere that isn't public and made available only to the `EncoderClient` and `EncoderServer` implementations?

src/viam/sdk/components/encoder/encoder.hpp
acmorrow: By "subtype" do you mean nested type?

src/viam/sdk/components/encoder/encoder.cpp
```diff
@@ -0,0 +1,125 @@
...
+            lhs.angle_degrees_supported == rhs.angle_degrees_supported);
+}
+
+namespace {
+bool init() {
+    Registry::register_subtype(Encoder::subtype(), Encoder::resource_subtype());
+    return true;
+};
+
+bool inited = init();
```
acmorrow, line 121: `const bool inited = ...`

src/viam/sdk/components/encoder/client.hpp
```diff
@@ -0,0 +1,38 @@
...
+/// @class EncoderClient
+/// @brief gRPC client implementation of a `Encoder` component.
+/// @ingroup Encoder
+class EncoderClient : public Encoder {
+   public:
+    position get_position(position_type position_type) override;
+    void reset_position() override;
+    properties get_properties() override;
+    AttributeMap do_command(AttributeMap command) override;
+    EncoderClient(std::string name, std::shared_ptr<grpc::Channel> channel)
```
acmorrow, line 27: - It is generally a good practice to put constructor declarations near the top of the class.
- Please move the definition to the .cpp file.

src/viam/sdk/components/encoder/encoder.hpp
```diff
@@ -0,0 +1,114 @@
...
+    struct properties {
+        bool ticks_count_supported;
+        bool angle_degrees_supported;
+    };
+
+    // functions shared across all components
+    static std::shared_ptr<ResourceSubtype> resource_subtype();
+    static Subtype subtype();
+
+    /// @brief Creates a `properties` struct from its proto representation.
```
acmorrow, line 70: Should that say `position_type`?

src/viam/sdk/components/encoder/encoder.hpp
```diff
@@ -0,0 +1,114 @@
...
+    static position_type from_proto(viam::component::encoder::v1::PositionType proto);
+
+    /// @brief Creates a `position` struct from its proto representation.
+    static position from_proto(viam::component::encoder::v1::GetPositionResponse proto);
+
+    /// @brief Creates a `properties` struct from its proto representation.
+    static properties from_proto(viam::component::encoder::v1::GetPropertiesResponse proto);
+
+    /// @brief Converts a `position` struct to its proto representation.
```
acmorrow, line 79: Ditto `position_type`?

src/viam/sdk/components/encoder/encoder.hpp
```diff
@@ -0,0 +1,114 @@
...
+
+    /// @brief Converts a `properties` struct to its proto representation.
+    static viam::component::encoder::v1::GetPropertiesResponse to_proto(properties properties);
+
+    /// @brief Returns position of the encoder which can either be ticks since last zeroing for an
+    /// incremental encoder or degrees for an absolute encoder.
+    /// @param position_type If supplied, the response will return the specified position type. If
+    /// the driver does not implement the requested type, this call will return an error. If
+    /// position type is UNSPECIFIED, the response will return a default according to the driver.
+    virtual position get_position(position_type position_type) = 0;
```
acmorrow, line 93: Given that the encoder will return a default if the argument is unspecified (or not-provided), should that be the defautl? After all, the return value carries with it the type information about the result, should one be obtained.

src/viam/sdk/components/encoder/server.hpp
```diff
@@ -0,0 +1,47 @@
...
+        const ::viam::component::encoder::v1::GetPropertiesRequest* request,
+        ::viam::component::encoder::v1::GetPropertiesResponse* response) override;
+
+    ::grpc::Status DoCommand(grpc::ServerContext* context,
+                             const viam::common::v1::DoCommandRequest* request,
+                             viam::common::v1::DoCommandResponse* response) override;
+
+    void register_server(std::shared_ptr<Server> server) override;
+
+    EncoderServer() : ResourceServerBase(std::make_shared<ResourceManager>()){};
```
acmorrow, line 42: Please move the ctor declarations to the top of the class, and move the definitions out of line into the .cpp file.

src/viam/sdk/components/encoder/server.hpp
```diff
@@ -0,0 +1,47 @@
...
+    EncoderServer(std::shared_ptr<ResourceManager> manager) : ResourceServerBase(manager){};
```
acmorrow, line 43: This single arg ctor should be `explicit`

acmorrow (changes requested):

src/viam/sdk/components/encoder/encoder.hpp
zaporter-work: yes

src/viam/sdk/components/encoder/encoder.hpp
zaporter-work: I don't fully understand this question. My (elementary) understanding of encoders is that all four permutations of these booleans are possible values.

src/viam/sdk/components/encoder/encoder.hpp
zaporter-work: That is a good question. I had assumed that this was done this way so that if we are late to implementing part of the API, customers are able to quickly patch/build it themselves. However that may not be what we want.

src/viam/sdk/components/encoder/encoder.hpp
zaporter-work: Translation typo. Will fix.

src/viam/sdk/components/encoder/encoder.hpp
zaporter-work: I chose the UNSPECIFIED route because it seemed cleaner / more readable at the time. `get_position(position_type::UNSPECIFIED)` vs `get_position(boost::none)`. The former is easier for me to read & understand but happy to change this.

src/viam/sdk/components/encoder/encoder.hpp
zaporter-work: Will also do for the other components.

src/viam/sdk/components/encoder/encoder.hpp
zaporter-work: Ahh, I hadn't used `enum class` before. Will switch. Thank you.

src/viam/sdk/components/encoder/client.hpp
zaporter-work: Will also switch for the other components.

src/viam/sdk/components/encoder/encoder.hpp
acmorrow: Yeah, I just didn't know, and I try to avoid types that have invalid states. If only one of these bools could meaningfully be true at once I would have suggested another approach.

It feels a little odd to me still that you have the enum, but then you also have a bool associated with each value of the enum. Just throwing out some ideas but maybe you could return `std::vector<position_type>` where only the supported position types were returned? Or `std::map<position_type, bool>`? As it stands you will need to change the size of this structure if ever a new `position_type` were added. And you need to know the name of two symbols to see if a certain `position_type` is supported, rather than just knowing the enumeration names.

src/viam/sdk/components/encoder/encoder.hpp
acmorrow: I think you can just make the default be the default!:

```
virtual position get_position(boost::optional<position_type> position_type = {});
```

Then `encoder.get_position()` means you will get back a position of the default type, but `encoder.get_position(encoder::position_type::TICKS_COUNT)` will get you the result in the type you want, if supported.

Meanwhile, how will errors about requests for unsupported position types be reported? An exception?

src/viam/sdk/components/encoder/encoder.hpp
acmorrow: Let's keep it for now. So far we haven't made a decision either way about whether the `proto` types should form part of the public API for the SDK. Currently they do, but we also write wrapper types (as you are doing with e.g. `struct position` above), so we sort of have a foot in each camp. I'm not sure which way we will resolve it, but it won't happen in this review. I think we can leave them for now but something to think about going forward.

src/viam/sdk/components/encoder/encoder.hpp
zaporter-work: Thank you for exposing me to this bit of history.

[Cpp reference](https://en.cppreference.com/w/cpp/language/enum) says `ticks_count` and `angle_degrees`.

[Google style guide](https://google.github.io/styleguide/cppguide.html#Enumerator_Names) says `kTicksCount` and `kAngleDegrees`

I am leaning towards `position_type::angle_degrees` because `position_type::k_angle_degrees` seems like the `k` means something to the user and the risk of collision is not obvious to me. 

Am I wrong about this? Happy to switch to whatever.

src/viam/sdk/components/encoder/encoder.hpp
zaporter-work: I agree that this is confusing and a bit unfortunate. 

I am hesitant to modify this too much because I suspect it was done this way to allow Viam to easily modify the `properties` struct going forward. Dropping to `vec<position_type>` would block us in and would make us different from the other sdks.

src/viam/sdk/components/encoder/encoder.hpp
zaporter-work: Typescript SDK uses:
```
async getPosition(
    positionType: EncoderPositionType = EncoderPositionType.POSITION_TYPE_UNSPECIFIED,
    extra = {}
  ) {
```

In order to support additional types of encoders, I am leaning towards keeping UNSPECIFIED as otherwise we would have to throw an error for get_position() if it returned an UNSPECIFIED type. (which is a judgement call for whether or not that is good or bad behavior)

src/viam/sdk/components/encoder/encoder.hpp
zaporter-work: >  think you can just make the default be the default

That is much better. Will do.

> Meanwhile, how will errors about requests for unsupported position types be reported? An exception?

If we pass an unsupported value, then the encoder impl (in go) returns an error that is turned into a grpc error that is turned into a thrown `std::runtime_error`

src/viam/sdk/components/encoder/encoder.hpp
zaporter-work: +1. Changed to `enum class position_type : uint8_t`

src/viam/sdk/components/encoder/encoder.hpp
zaporter-work: Updated to use optional type. 
Closing.

src/viam/sdk/components/encoder/encoder.hpp
zaporter-work: +1. Set to `position_type position_type = position_type::unspecified` for now. 

Given this, do we need the boost::optional<> wrapping?

acmorrow (approved):
LGTM

src/viam/sdk/CMakeLists.txt
```diff
@@ -44,6 +44,9 @@ target_sources(viamsdk
     components/camera/camera.cpp
     components/camera/client.cpp
     components/camera/server.cpp
+    components/encoder/encoder.cpp
+    components/encoder/client.cpp
```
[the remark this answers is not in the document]
zaporter-work: Good catch. I wonder if there is a tool that auto-sorts lists like these in cmake files.

PR #94, opened 2023-05-03
RSDK-2888 - Fix robot tests

stuqdog:


src/viam/sdk/registry/registry.hpp
```diff
@@ -32,10 +32,7 @@ class ResourceSubtype {
    public:
     virtual ~ResourceSubtype();
     /// @brief Add `Reconfigure` functionality to a resource.
-    std::function<Resource(Resource, Name)> create_reconfigurable;
-
-    // TODO: it doesn't look like we actually use this. Confirm, then remove.
-    std::function<ProtoType(Resource)> create_status;
```
stuqdog, line 38: (flyby) had to touch this signature anyways, seemed an easy time to follow up on the `TODO`.

src/viam/sdk/resource/resource_base.cpp
```diff
@@ -31,15 +31,15 @@ Subtype Resource::subtype() {
...
+ResourceType Resource::type() const {
     return {RESOURCE};
 }
 
 ResourceName Resource::get_resource_name(std::string name) {
     ResourceName r;
     *r.mutable_namespace_() = RDK;
     *r.mutable_type() = RESOURCE;
-    *r.mutable_subtype() = this->type().to_string();
+    *r.mutable_subtype() = this->instance_subtype().resource_subtype();
```
stuqdog, line 42: (flyby) this was being assigned incorrectly before. Caught in robot tests.

src/viam/sdk/resource/resource_base.hpp
```diff
@@ -20,12 +20,30 @@ class Resource {
...
+
+    /// @brief Reconfigures a resource. By default a no-op.
+    /// @param deps Dependencies of the resource.
+    /// @param cfg The resource's config.
     virtual void reconfigure(Dependencies deps, ResourceConfig cfg);
-    virtual std::string name();
-    virtual ResourceType type();
+
+    /// @brief Return the resource's name.
+    virtual std::string name() const;
```
stuqdog, line 43: (flyby) add const

src/viam/sdk/resource/resource_base.hpp
```diff
@@ -20,12 +20,30 @@ class Resource {
     virtual ~Resource();
     explicit Resource(std::string name) : name_(std::move(name)){};
     static Subtype subtype();
+
+    /// @brief Returns the `Subtype` associated with a particular resource.
+    virtual Subtype instance_subtype() const = 0;
```
stuqdog, line 25: This was necessary for subtype comparison. Previously calling to `subtype()` would return the base level `Resource` subtype unless the class in question happened to remember to implement a new static `Subtype()` method. This forces us to provide more granular access to a resource's subtype.

src/viam/sdk/robot/service.cpp
```diff
@@ -44,11 +44,12 @@ std::vector<Status> RobotService_::generate_status(RepeatedPtrField<ResourceName
         std::shared_ptr<Resource> resource = cmp.second;
         for (auto& registry : Registry::registered_resources()) {
             std::shared_ptr<ModelRegistration> registration = registry.second;
-            if (registration->resource_type() == resource->type()) {
+            if (registration->subtype().resource_subtype() ==
+                resource->instance_subtype().resource_subtype()) {
```
stuqdog, line 48: Fix comparison, exposed by tests.

src/viam/sdk/robot/service.cpp
```diff
@@ -152,19 +153,19 @@ ::grpc::Status RobotService_::StopAll(::grpc::ServerContext* context,
         AttributeMap value_map = struct_to_map(struct_);
         std::string name = ex.name().SerializeAsString();
         extra.emplace(name, value_map);
+    }
```
stuqdog, line 156: (flyby) This closing brace was misplaced, led to some bugs.

src/viam/sdk/resource/resource_base.hpp
zaporter-work: Could we make `subtype()` virtual? It is a bit confusing to have two `subtype()` functions, especially when one calls the other

stuqdog: We cannot, because `subtype()` is `static`. Unfortunately we have both the following cases:

- I know what resource type I want a `subtype` for, but I may not have access to an actual resource at the moment
- I have some resource but I don't know which kind it is, I would like to look at its subtype and compare

The static `subtype()` solves the first case, the virtual `instance_subtype()` solves the second. I would have loved to use the same name for these but I don't think you can overload a static method with a virtual one, at least when they both have the same signature.

zaporter-work: Ahh I see. I had missed the `static` on `subtype`...

acmorrow: You are correct, you cannot overload on static vs virtual.

Could you provide a little more background on what the 'subtype' system is about? I've been meaning to ask anyway, and I'd like to make sure I understand it before I think about ways you could simplify this.

acmorrow: Also a clarification on`Subtype` vs `ResourceSubtype` vs `ResourceType`.

src/viam/sdk/tests/test_robot.cpp
```diff
@@ -1,323 +1,212 @@
...
+            std::shared_ptr<Resource> rb = service.resource_manager()->resource("mock_motor");
+            auto motor = std::dynamic_pointer_cast<motor::MockMotor>(rb);
+            BOOST_CHECK(motor);
 
-    if (from_call == names_for_testing()) {
-        return 0;
-    }
-    return 1;
+            motor->set_power(10.0);
+            BOOST_CHECK(motor->get_power_status().power_pct == 10.0);
```
zaporter-work, line 200: [nit] can use BOOST_CHECK_EQUAL here for better logging

src/viam/sdk/tests/test_robot.cpp
```diff
@@ -1,323 +1,212 @@
...
+
+            // get all resource statuses
+            auto statuses = client->get_status();
+
+            auto status_strs = vec_to_string_util(statuses);
+            auto mock_strs = vec_to_string_util(mock_statuses);
+
+            // ensure we get statuses for all resources, and that they are as expected.
+            BOOST_CHECK(statuses.size() == 3);
+            BOOST_CHECK(vec_to_string_util(statuses) == vec_to_string_util(mock_statuses));
```
zaporter-work, line 131: [nit] I was doing some reading on `boost::test` a few days ago, and I think the right way to do this is with `BOOST_TEST(vec_to_string_util(statuses) == vec_to_string_util(mock_statuses), boost::test_tools::per_element() );`

This should do element by element comparison and give a much nicer error message. https://www.boost.org/doc/libs/1_75_0/libs/test/doc/html/boost_test/testing_tools/extended_comparison/collections.html#boost_test_coll_perelement


If the `statuses` are comparable we could just do `BOOST_TEST(statuses == mock_statuses, boost::test_tools::per_element() );`

src/viam/sdk/tests/test_robot.cpp
```diff
@@ -1,323 +1,212 @@
...
+
+            // get only a subset of status responses
+            auto names = mock_resource_names_response();
+            std::vector<ResourceName> some_names{names[0], names[1]};
+            auto some_statuses = client->get_status(some_names);
+            auto some_status_strs = vec_to_string_util(some_statuses);
+
+            // ensure that we only get two of the three existing statuses
+            BOOST_CHECK(some_status_strs.size() == 2);
```
zaporter-work, line 140: [nit] BOOST_CHECK_EQUAL will also print the value of `some_status_strs.size()` if the comparison fails

src/viam/sdk/resource/resource_base.hpp
stuqdog: Yeah, so `Subtype`/`Type` as names are unfortunately a bit overloaded here. The short version is `Subtype` is the equivalent of RDK's `API` in `resource/api.go`. Basically it identifies a resource by its namespace (e.g. RDK), type (component/service), and subtype (camera/SLAM/motor). The name was changed recently from `Subtype` to `API`: I had been echoing the naming conventions in RDK and Python, and should probably continue coing so by renaming these types in a future PR.

`ResourceType` is just a string wrapper around a particular resource type, e.g. Camera or Motor.

`ResourceSubtype` is a collection of methods specific to a particular resource that get registered so we know how to, e.g., create a `CameraClient`. It is very poorly named, I wish I had a good answer for how I ended up with that name but I'm not sure 😅. Should definitely be renamed to `ResourceRegistration`.

Would prefer to not muddy the waters here by doing all the renaming, but a ticket is [here](https://viam.atlassian.net/browse/RSDK-2984).

acmorrow: Yeah, definitely don't do the renaming in this review, but thanks for the clarification. Naming is hard!

Maybe call these `static_subtype` and `dynamic_subtype`. And then add `static_subtype` to `Resource`, `Component`, and `Service` so that it is consistently available throughout the hierarchy.

Another way forward would be to leave the dynamic version as `subtype` in `Resource`, and then use a template and specialization to make a `static Subtype Subtype::for<T>()` function where each T provides the result:


// In `subtype.hpp`:
```
class Subtype {
public:
    ....

    // Specialize subtype_map for each type T for which
    // Subtype::for<T>() should be defined. The specialization
    // should declare `static Subtype type()`.
    template<typename T>
    struct subtype_map;

    template<typename T>
    static Subtype for() {
        return subtype_map<T>::type();
    };
};
```

In `camera.hpp`:
```
class Camera {
public:
    ....
    /// @override Resource::subtype
    Subtype subtype() const override { return SubType::for<Camera>() };
    ...
};

template<>
struct Subtype::subtype_map<Camera> {
    static Subtype type() {
        return {RDK, COMPONENT, "camera"};
    }
};
```

Obviously, the actual implementations do not need to be inline / declared in the class body. Basically, this gets rid of the collision between the static / virtual by making the way to get the static subtype of something part of the `Subtype` API rather than the per-type API.

Usage now looks like:

```
// Get the runtime subtype of an object via its `Resource` pointer (or anywhere else in class hierarchy).
auto subtype = r->subtype(); 

// Get the static subtype of a thing:
auto subtype = Subtype::for<Camera>();
```

Just some (untested) ideas!

src/viam/sdk/resource/resource_base.hpp
```diff
@@ -20,12 +20,30 @@ class Resource {
...
     virtual viam::common::v1::ResourceName get_resource_name(std::string name);
+
+    /// @brief Stops a resource from running.
+    /// @param extra Extra arguments to pass to the resource's `stop` method.
     virtual grpc::StatusCode stop(AttributeMap extra);
+
+    /// @brief Stops a resource from running.
     virtual grpc::StatusCode stop();
+
+    /// @brief Reconfigures a resource. By default a no-op.
```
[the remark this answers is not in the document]
stuqdog: Fair point, removing that line.

src/viam/sdk/robot/client.hpp
```diff
@@ -107,10 +109,14 @@ class RobotClient {
     /// @return The list of operations currently running on the calling robot.
     std::vector<viam::robot::v1::Operation> get_operations();
 
-    /// @brief Get the status of the robot's components.
-    /// @param components An optional list of the specific components for which status is desired.
+    /// @brief Get the status of the requested robot components.
+    /// @param components A list of the specific components for which status is desired.
+    /// @return A list of statuses.
+    std::vector<Status> get_status(std::vector<ResourceName>& components);
```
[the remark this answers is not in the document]
stuqdog: Yep, completely agree! I think it's outside the scope of this PR though, since that's a bit of a refactor of the robot client in general.

src/viam/sdk/resource/resource_base.hpp
```diff
@@ -20,12 +20,30 @@ class Resource {
     virtual ~Resource();
     explicit Resource(std::string name) : name_(std::move(name)){};
     static Subtype subtype();
+
```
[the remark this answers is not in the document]
stuqdog: Ahh yeah, I actually made a comment in the description of that PR about that. I weakly prefer keeping the name as is. I realize this creates an inconsistency between class and file names, but I view `resource.cpp` as not particularly descriptive on its own, whereas `Resource` as a class is quite clear given the context of comments and its use as an inherited class. Additionally I think the risk of collision with `base` being overloaded is much lower given the general file structure.

That said, my opinion here is a weak one and I do recognize the value of maintaining parity between class and file names. So, happy to revisit if you feel strongly!

src/viam/sdk/resource/resource_base.hpp
stuqdog: Oh, I do like the look of `Subtype::for<T>();`! That looks fun and clean. For the purpose of this PR though I think I'm going to go with the low hanging fruit and just rename to `{static|dynamic}_subtype()`. Ticket for looking into better, more involved solutions [here](https://viam.atlassian.net/browse/RSDK-2990).

cheukt (approved):
can't comment specifically, but the additional tests look good

src/viam/sdk/resource/resource_base.hpp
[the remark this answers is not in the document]
acmorrow: I feel pretty strongly that the file should be renamed, but I also don't think it is urgent to do so. I also wouldn't want to wait too long. My suggestion is to leave it for now and file a ticket to revisit the mapping of classes to files. There are other cases where they aren't 1:1, and they probably should be. It makes navigation a lot easier when you can know that `class FooBar` is always going to be in a file `foo_bar.hpp`.

PR #96, opened 2023-05-03
RSDK-793 - Add board wrapper

zaporter-work:
Ticket: https://viam.atlassian.net/browse/RSDK-793

Important decisions are highlighted in my review.

I think this component is complex enough to deserve an example. I will do that in a separate PR.


src/viam/sdk/common/utils.hpp
```diff
@@ -44,5 +44,8 @@ class ResourceNameEqual {
 std::vector<unsigned char> string_to_bytes(std::string const& s);
 std::string bytes_to_string(std::vector<unsigned char> const& b);
 
+std::chrono::duration<double> duration_from_proto(const google::protobuf::Duration& proto);
+google::protobuf::Duration duration_to_proto(const std::chrono::duration<double>& duration);
```
zaporter-work, line 48: These feel very awkward here. I cannot call them `from_proto` in global namespace as otherwise there are discovery problems

If people think this is a problem, I would propose either:
1) a shared `common/duration.hpp` in a new `common` namespace (move Vector3 into this new namespace as well)
2) a `viam::cpp::duration` class that wraps `std::chrono::duration<double>` (I think this would be confusing to the user)

zaporter-work, line 48: How do we feel about `std::chrono::duration<double>` as the default viam::sdk duration unit? Some online research seems to say that this is the approved and portable way of doing things (and it lets people use whatever units they want).

I was thinking of `typedef`ing this like elsewhere but I am worried that it will make the type even harder to grok

src/viam/sdk/components/board/board.hpp
```diff
@@ -0,0 +1,162 @@
...
+///
+/// This acts as an abstract base class to be inherited from by any drivers representing
+/// specific board implementations. This class cannot be used on its own.
+class Board : public Component {
+   public:
+    /// @brief Represents the value received by the registered analog to digital converter (ADC).
+    /// The range and conversion mechanism to voltage will vary depending on the specific ADC
+    /// registered to the pin. Consult your ADC's documentation and Viam's `Board` documentation for
+    /// more details.
+    typedef int32_t analog_value;
```
zaporter-work, line 46: Name justification: Very little. This is what it is called in the other sdks but that doesn't mean that it isn't confusing

Typedef justification: Treating this as a special type may give us a bit more flexibility going forward (on major version bumps), and may prevent surprising rounding errors for customers.

src/viam/sdk/components/board/board.hpp
```diff
@@ -0,0 +1,162 @@
...
+
+    /// @brief Depending on the type of digital interrupt, this can have different meanings. If a
+    /// `basic` (default) interrupt is configured, then this is the total interrupt count. If a
+    /// `servo` interrupt is configured this tracks the pulse width value. Consult Viam's `Board`
+    /// documentation for more details.
+    typedef int64_t digital_value;
```
zaporter-work, line 52: ditto

src/viam/sdk/components/board/board.hpp
```diff
@@ -0,0 +1,162 @@
...
+
+    /// @struct board_reader_status
+    /// @brief This contains all of the values for all of the registered analog readers and digital
+    /// interrupts that have been registered on the board.
+    struct status {
+        std::map<std::string, analog_value> analog_reader_values;
+        std::map<std::string, digital_value> digital_interrupt_values;
+    };
```
zaporter-work, line 60: The other sdk's have also chosen the name `status` to indicate this very specific and not-general structure. I am hesitant about this name but have left it in for now. 

I am leaning towards switching to `board_reader_status` but am not sure if that is better. 

The difficult part is that the digital values are `interrupts` but the analog values are `readers` (and GPIO info is not present)

src/viam/sdk/components/board/board.hpp
```diff
@@ -0,0 +1,162 @@
...
+    static viam::common::v1::AnalogStatus to_proto(analog_value analog_value);
+
+    /// @brief Converts a `digital_value` struct to its proto representation.
+    static viam::common::v1::DigitalInterruptStatus to_proto(digital_value digital_value);
+
+    /// @brief Converts a `power_mode` enum to its proto representation.
+    static viam::component::board::v1::PowerMode to_proto(power_mode power_mode);
+
+    /// @brief Get the status of all of the registered analog readers and digital interrupt readers
+    virtual status get_status() = 0;
```
zaporter-work, line 96: If we switch status to `board_reader_status` this will become `get_reader_status`

src/viam/sdk/components/board/board.hpp
```diff
@@ -0,0 +1,162 @@
...
+    /// @brief Sets the given pin on a board to the given PWM frequency. 0 will use the board's
+    /// default PWM frequency.
+    /// @param pin board pin name
+    /// @param frequency_hz frequency in hz (0 = use board default frequency)
+    virtual void set_pwm_frequency(const std::string& pin, uint64_t frequency_hz) = 0;
+
+    /// @brief Reads off the current value of an analog reader on a board. Consult your ADC's docs
+    /// or Viam's `Board` docs for more information.
+    /// @param analog_reader_name analog reader to read from
+    virtual analog_value read_analog(const std::string& analog_reader_name) = 0;
```
zaporter-work, line 130: This is awkward. `get_analog` was another contender but then people might accidentally pass in a `pin` rather than the name of the reader.

src/viam/sdk/components/board/board.hpp
```diff
@@ -0,0 +1,162 @@
...
+
+    /// @brief Returns the current value of the interrupt which is based on the type of interrupt.
+    /// Consult Viam's `Board` docs for more information.
+    /// @param digital_interrupt_name digital interrupt to check
+    virtual digital_value read_digital_interrupt(const std::string& digital_interrupt_name) = 0;
```
zaporter-work, line 135: This is even more awkward but was done for similar reasons as above.

src/viam/sdk/components/board/board.hpp
```diff
@@ -0,0 +1,162 @@
...
+
+    /// @brief Get all defined analog readers defined for this board
+    virtual std::vector<std::string> get_analog_reader_names() = 0;
+
+    /// @brief Get all defined digital interrupts for this board
+    virtual std::vector<std::string> get_digital_interrupt_names() = 0;
```
zaporter-work, line 141: Added to keep parity with python sdk and our docs which assert that these methods exist

src/viam/sdk/components/board/board.hpp
```diff
@@ -0,0 +1,162 @@
...
+    virtual bool get_gpio(const std::string& pin) = 0;
+
+    /// @brief Set the gpio high/low state of the given pin on a board
+    /// @param high true if the pin should be set to high (on) or false if it should be low (off)
+    virtual void set_gpio(const std::string& pin, bool high) = 0;
+
+    /// @brief Gets the duty cycle of the given pin on a board.
+    /// @param pin board pin name
+    /// @return duty cycle percentage (0 to 1)
+    virtual double get_pwm_duty_cycle(const std::string& pin) = 0;
```
zaporter-work, line 110: The other sdks call this function pair `get_pwm` and `set_pwm`. I found this confusing because we also have `get_pwm_frequency` and `set_pwm_frequency` and it is not obvious (to me) that `set_pwm` would refer to the duty cycle rather than to the frequency.

src/viam/sdk/components/board/board.hpp
```diff
@@ -0,0 +1,162 @@
...
+};
+
+/// @class Board board.hpp "components/board/board.hpp"
+/// @brief Represents a physical board with gpio pins, digital interrupts, and analog voltage
+/// reading
+/// @ingroup Board
+///
+/// This acts as an abstract base class to be inherited from by any drivers representing
+/// specific board implementations. This class cannot be used on its own.
+class Board : public Component {
```
zaporter-work, line 40: Python SDK chose OOP and created the following subclasses:
1) AnalogReader
2) DigitalInterrupt
3) GPIOPin

The Typescript SDK did not follow this route. 

I also chose the less-OOP method with a single board class it did not seem to improve readability to have many subclasses. I might be mistaken in this.

src/viam/sdk/common/utils.cpp
```diff
@@ -55,5 +55,22 @@ std::string bytes_to_string(std::vector<unsigned char> const& b) {
...
+
+google::protobuf::Duration duration_to_proto(const std::chrono::duration<double>& duration) {
+    google::protobuf::Duration proto;
+    using namespace std::chrono_literals;
+    double total_seconds = duration / 1.0s;
+    long integer_seconds = std::truncl(total_seconds);
+    long integer_nanos = std::truncl(1e9 * (total_seconds - integer_seconds));
+
+    proto.set_seconds(integer_seconds);
+    proto.set_nanos(integer_nanos);
```
zaporter-work, line 71: `std::chrono::duration_from` had rounding errors with my nanosecond tests. This did not have those errors (though this is clearly very hacky so I would be happy to hear suggestions.)

src/viam/sdk/components/board/board.hpp
acmorrow: It's OK I think. I agree that `status` is a bit of a land grab-y name, but the usage is scoped to `Board` so it isn't that bad.

src/viam/sdk/common/utils.hpp
acmorrow: I'm not quite sure what you mean by "discovery problems", can you elaborate?

acmorrow: I wouldn't typedef it. I think `std::chrono::duration<double>` makes it very clear what it is. I need to think about how I feel about using `double` with `std::chrono::duratuion` a bit. It was surprising to me at first but I think I'm seeing why it might be a good choice.

src/viam/sdk/components/board/board.hpp
acmorrow: I think you made the right call. I don't see that making types for those achieves much in the context of the C++ SDK.

src/viam/sdk/components/board/board.hpp
```diff
@@ -0,0 +1,162 @@
...
+
+    /// @brief Get all defined digital interrupts for this board
+    virtual std::vector<std::string> get_digital_interrupt_names() = 0;
+
+    /// @brief Sets the power consumption mode of the board to the requested setting for the given
+    /// duration.
+    /// @param power_mode Requested power mode
+    /// @param duration Requested duration to stay in `power_mode`
+    virtual void set_power_mode(power_mode power_mode,
+                                const std::chrono::duration<double>& duration) = 0;
```
acmorrow, line 148: I would have expected this to take an optional duration, as I believe that the proto for set_power_mode declares it so.

stuqdog (commented):
First pass look. Generally looks good, may need to think some more on some of the naming questions.

acmorrow (commented):
This looks pretty good overall. I've replied to some of your comments with my thoughts.

src/viam/sdk/common/utils.hpp
zaporter-work: If I remove the `duration_` prefix in utils.hpp and utils.cpp I get the following error:
```
v/nix/store/dkw46jgi8i0bq64cag95v4ywz6g9bnga-gcc-wrapper-11.3.0/bin/g++ -DBOOST_ATOMIC_DYN_LINK -DBOOST_ATOMIC_NO_LIB -DBOOST_CHRONO_DYN_LINK -DBOOST_CHRONO_NO_LIB -DBOOST_FILESYSTEM_DYN_LINK -DBOOST_FILESYSTEM_NO_LIB -DBOOST_LOG_DYN_LINK -DBOOST_LOG_NO_LIB -DBOOST_REGEX_DYN_LINK -DBOOST_REGEX_NO_LIB -DBOOST_THREAD_DYN_LINK -DBOOST_THREAD_NO_LIB -Dviamsdk_EXPORTS -I/home/zack/work/viam-cpp-sdk/build/src/viam/api/gen/viam/api/../.. -I/home/zack/work/viam-cpp-sdk/build/src/viam/api/gen/viam/api -I/home/zack/work/viam-cpp-sdk/src/viam/sdk/../.. -g -fPIC -Wall -Werror -std=c++14 -MD -MT src/viam/sdk/CMakeFiles/viamsdk.dir/components/board/client.cpp.o -MF src/viam/sdk/CMakeFiles/viamsdk.dir/components/board/client.cpp.o.d -o src/viam/sdk/CMakeFiles/viamsdk.dir/components/board/client.cpp.o -c /home/zack/work/viam-cpp-sdk/src/viam/sdk/components/board/client.cpp
/home/zack/work/viam-cpp-sdk/src/viam/sdk/components/board/client.cpp: In member function ‘virtual void viam::sdk::BoardClient::set_power_mode(viam::sdk::Board::power_mode, const std::chrono::duration<double>&)’:
/home/zack/work/viam-cpp-sdk/src/viam/sdk/components/board/client.cpp:194:43: error: no matching function for call to ‘viam::sdk::BoardClient::to_proto(const std::chrono::duration<double>&)’
  194 |     *request.mutable_duration() = to_proto(duration);
      |                                   ~~~~~~~~^~~~~~~~~~
```

zaporter-work: The board client tries to find a locally (i.e. Board) overloaded `to_proto` rather than the global utils.hpp `to_proto`

src/viam/sdk/components/board/client.cpp
```diff
@@ -0,0 +1,226 @@
...
+      stub_(viam::component::board::v1::BoardService::NewStub(channel)),
+      channel_(std::move(channel)){};
+
+Board::status BoardClient::get_status() {
+    viam::component::board::v1::StatusRequest request;
+    viam::component::board::v1::StatusResponse response;
+
+    grpc::ClientContext ctx;
+
+    *request.mutable_name() = this->name();
```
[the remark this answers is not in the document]
zaporter-work: Because, for reasons I dont fully understand, `protoc` does not generate a `set_name` function. There are a couple other members like this (like `duration`) where there is no setter.

src/viam/sdk/components/board/board.cpp
```diff
@@ -0,0 +1,133 @@
...
+        }
+        default: {
+            throw std::runtime_error("Invalid board power_mode to encode");
+        }
+    }
+}
+
+Board::Board(std::string name) : Component(std::move(name)){};
+
+bool operator==(const Board::status& lhs, const Board::status& rhs) {
```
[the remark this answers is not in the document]
zaporter-work: The compiler doesn't seem to generate an equality operator https://github.com/viamrobotics/viam-cpp-sdk/pull/96#discussion_r1185307150. 

I was also very surprised by this.

src/viam/sdk/common/utils.hpp
acmorrow: I'll bet you can make them visible with a `using` statement:

```
void viam::sdk::BoardClient::set_power_mode(...) {
    using ::viam::to_proto;
    
}
```

acmorrow: @zaporter-work - I think it would be helpful if you outlined your argument for using `double` as the type for `Rep` in `duration`.

acmorrow: Essentially, you are running into the fact that `from_proto` and `to_proto` are overloaded at different scopes so the resolution gets a little weird.

I think go with explicitly qualifying it for now, and keep the name consistent. I overall have some interest in how the proto conversion stuff evolves but I don't think now is the time to pursue it.

zaporter-work: I did this for two reasons: 
1)
We are given a `int64_t` seconds and a `int32_t` from proto duration. 
To the best of my understanding, there is no `int128_t` in C++ (even `long long` is 64b).
So if we want to store everything under a single number, we must use a 64 bit type.

I had thought about using `duration<long long, nanoseconds>` but then I am sacrificing precision in the important dimension (seconds) for a bit of extra precision in the less important dimension (nanoseconds).

This led me towards creating a `duration<double, seconds>` from the proto generated code.

This has the risk of losing precision in the nanoseconds dimension (however we have a test that tests no loss of precision at least for small numbers).

2) 
It makes it slightly easier to convert between different time units in an agnostic way (ex: if a user casts our `seconds`-based duration to `minutes` they will have to intentionally chose to lose precision. If it is a double, then no information will be rounded or truncated)

zaporter-work: I do not have deep experience with this type though, so if either of you have strong opinions, I am happy to go with `std::chrono::duration<long long>`

src/viam/sdk/components/board/board.hpp
zaporter-work: Switched to `using` and resolving.

src/viam/sdk/CMakeLists.txt
```diff
@@ -128,6 +135,7 @@ target_sources(viamsdk
       ../../viam/sdk/tests/mocks/camera_mocks.hpp
       ../../viam/sdk/tests/mocks/generic_mocks.hpp
       ../../viam/sdk/tests/mocks/mock_base.hpp
+      ../../viam/sdk/tests/mocks/mock_board.hpp
```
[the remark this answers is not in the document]
zaporter-work: Agreed. That would be great!

src/viam/sdk/components/board/board.hpp
zaporter-work: I have made this modification in https://github.com/viamrobotics/viam-cpp-sdk/pull/96/commits/835acd99f4e8ce8897850287ae20961805e457a8 .

src/viam/sdk/components/board/board.hpp
zaporter-work: I had an idea:

We could drop the `status` object entirely and delete the `get_status` method. 

Then we could add a `get_analog_readers() -> std::unordered_map<string, analog_value>`, `get_digital_readers() -> ...` which internally calls the `Status()` grpc call and then breaks out the struct.

This would clean up one of the more confusing parts of the API.

src/viam/sdk/common/utils.hpp
acmorrow: So, interestingly, the 64-bit seconds part of a `proto.Duration` has a limited range, smaller than the full signed 64-bit range: https://protobuf.dev/reference/protobuf/google.protobuf/#duration.

That range happens to be an interesting looking number, so I went looking for a justification, and found one: https://github.com/protocolbuffers/protobuf/blob/main/src/google/protobuf/duration.proto#L103.

We cannot represent +/- 10ky in int64_t nanos: (2^63) / (10,000 * 365.25 * 24 * 60 * 60 * 1,000,000,000) = 0.0292271023...

But we **can** represent +/- 10ky in int64_t micros: (2^63) / (10,000 * 365.25 * 24 * 60 * 60 * 1,000,000) = 29.227...

We do lose a tiny bit of precision, but at most one microsecond. So I'd be pretty comfortable declaring that our vocabulary duration type in the Viam C++ SDK is denominated in microseconds.

If we are uncomfortable with loosing a whole microsecond, it looks to me like we have another decimal place, so if we denominated in units of 100 nanoseconds that would be our maximum precision loss when going to/from proto. It is a non-standard chrono ratio though.

src/viam/sdk/components/board/board.hpp
acmorrow: That would mean though that if you really did want both you would need to make two grpc calls, right? I'm not sure that is a good idea. What if you really are interested in snapshotting both an analog and digital reading at the "same time"?

zaporter-work: Good point I hadn't considered that.

acmorrow: Per discussion, if you want an API that returns just the analog / just the digitals, make some helper functions that call the big function and drop the uninteresting result.

zaporter-work: @stuqdog Are you also onboard with this?

zaporter-work: If I do this, I think it makes sense to delete `get_analog_reader_names` and `get_digital_interrrupt_names` and replace then with `get_analog_readers()` and `get_digital_interrupts()` (the former two already call `Status` and then create a vector from the resulting keys of the maps)

acmorrow: I think personally I'd keep the `names` ones for alignment with other APIs. The implementation is trivial, just vectorize the map keys.

src/viam/sdk/components/board/board.hpp
```diff
@@ -0,0 +1,162 @@
...
+    /// @brief Send/receive arbitrary commands to the resource.
+    /// @param Command the command to execute.
+    /// @return The result of the executed command.
+    virtual AttributeMap do_command(AttributeMap command) = 0;
+
+   protected:
+    explicit Board(std::string name);
+};
+
+bool operator==(const Board::status& lhs, const Board::status& rhs);
```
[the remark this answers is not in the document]
acmorrow: That's odd, but not worth worrying that much about. I do wonder if defining comparison operators for `board::status` is actually useful. If you just need it for testing, you can always just define one in the test file.

src/viam/sdk/components/board/board.cpp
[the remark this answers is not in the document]
zaporter-work: Resolving in favor of that conversation.

src/viam/sdk/common/utils.hpp
stuqdog: I'm a fan of representing time with int over float and I don't expect that nano precision is going to be especially important here, so I'm in favor of using micro denomination with int64_t. The potential loss of a microsecond is a bit sad but I don't think it'll be terrible in practice, and is something that we can easily account for in tests as necessary.

zaporter-work: I have just made this change in c89803d4cf8db3ffa1389573a79f7e96e6870752. I think it is slightly less ergonomic but it is nice to have integer-based durations.

zaporter-work: @stuqdog @acmorrow If you could skim that commit and confirm that is what is preferred, that would be appreciated.

src/viam/sdk/components/board/board.hpp
zaporter-work: Done in https://github.com/viamrobotics/viam-cpp-sdk/pull/96/commits/c4a1e93b1259b406bc3547d85cce10b79e492fb1. Feel free to unresolve if you have more to discuss.

stuqdog (approved):
lgtm

src/viam/sdk/common/utils.cpp
```diff
@@ -55,5 +55,21 @@ std::string bytes_to_string(std::vector<unsigned char> const& b) {
     return img_string;
 };
 
+std::chrono::duration<int64_t, std::micro> from_proto(const google::protobuf::Duration& proto) {
+    return std::chrono::microseconds(
+        (int64_t)std::ceil(proto.seconds() * 1e6 + proto.nanos() / 1e3));
```
acmorrow, line 60: This seems like there is a lot going on. There is `std::ceil`, floating point multiplications and divisions, etc.

Can you simplify it by decomposing the `protos.seconds` and `protos.nanos` separately into the `chrono` types and then combining?

```
... from_proto(Duration& proto) {
    namespace sc  = std::chrono;

    const sc::seconds seconds_part{proto.seconds()};
    const sc::nanoseconds nanos_part{proto.nanos()};
    
    const sc::microseconds from_seconds = std::duration_cast<sc::microseconds>(seconds_part);

    sc::microseconds from_nanos = 
std::duration_cast<sc::microseconds>(nanos_part);

    if ((nanos_part < 0) && (from_nanos > nanos_part))
        from_nanos -= 1;
    else if ((nanos_part > 0) && (from_nanos < nanos_part))
        from_nanos += 1;

    return from_seconds + from_nanos;
}
```

Rounding above is untested but inspired by
- https://en.cppreference.com/w/cpp/chrono/duration/ceil
- https://en.cppreference.com/w/cpp/chrono/duration/floor

Unfortunately, we can't use those because they are C++17 only.

Also, I'd definitely like to see some explicit unit tests for these conversions.

   
    

}

src/viam/sdk/common/utils.cpp
```diff
@@ -55,5 +55,21 @@ std::string bytes_to_string(std::vector<unsigned char> const& b) {
...
+}
+
+google::protobuf::Duration to_proto(const std::chrono::duration<int64_t, std::micro>& duration) {
+    google::protobuf::Duration proto;
+    int64_t total_micros = duration.count();
```
acmorrow, line 65: Again, I think you can achieve these conversions more seamlessly by using duration_cast, and tests.

src/viam/sdk/components/board/board.hpp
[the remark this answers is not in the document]
acmorrow: I find it unlikely that analog results are stable enough for that to work, but I buy the argument in theory. Resolving.

src/viam/sdk/components/board/board.hpp
```diff
@@ -0,0 +1,174 @@
...
+    /// @param digital_interrupt_name digital interrupt to check
+    virtual digital_value read_digital_interrupt(const std::string& digital_interrupt_name) = 0;
+
+    /// @brief Sets the power consumption mode of the board to the requested setting for the given
+    /// duration.
+    /// @param power_mode Requested power mode
+    /// @param duration Requested duration to stay in `power_mode`
+    virtual void set_power_mode(power_mode power_mode,
+                                boost::optional<const std::chrono::duration<int64_t, std::micro>&>
+                                    duration = boost::none) = 0;
```
acmorrow, line 160: I think you can just say ` = {}` rather than needing to say `boost::none`.

src/viam/sdk/components/board/board.hpp
```diff
@@ -0,0 +1,174 @@
...
+    /// Consult Viam's `Board` docs for more information.
+    /// @param digital_interrupt_name digital interrupt to check
+    virtual digital_value read_digital_interrupt(const std::string& digital_interrupt_name) = 0;
+
+    /// @brief Sets the power consumption mode of the board to the requested setting for the given
+    /// duration.
+    /// @param power_mode Requested power mode
+    /// @param duration Requested duration to stay in `power_mode`
+    virtual void set_power_mode(power_mode power_mode,
+                                boost::optional<const std::chrono::duration<int64_t, std::micro>&>
```
acmorrow, line 159: - Maybe just `std::chrono::microseconds`?
- Something looks a little off here. Why is the `&` inside the `optional`?

acmorrow (changes requested):

src/viam/sdk/common/utils.cpp
```diff
@@ -55,5 +55,21 @@ std::string bytes_to_string(std::vector<unsigned char> const& b) {
     return img_string;
 };
 
+std::chrono::duration<int64_t, std::micro> from_proto(const google::protobuf::Duration& proto) {
```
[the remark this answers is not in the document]
acmorrow: I think that's some C++ standards pedantry there. There is surely a good reason for it but I don't think it is relevant to our purposes. That's just how many bits is required to meet the needs of `microseconds`, but given that there is no machine I'm aware of with a signed integral type meeting that requirement other than int64_t I think we are safe. If you want to be paranoid, you could always:
```
static_assert(sizeof(std::chrono::microseconds::rep) >= sizeof(std::int64_t));
```
And at least we would fail to compile if our assumptions were wrong.

src/viam/sdk/common/utils.cpp
zaporter-work: Good idea. This is significantly cleaner.

src/viam/sdk/common/utils.cpp
zaporter-work: +1. I appreciate the push. I didn't trust chrono duration enough until I added tests. Now more confident in its arithmetic and rounding behavior.

zaporter-work: https://github.com/viamrobotics/viam-cpp-sdk/pull/96/commits/5e45493c4a18e7fc10e0f05dc87ae44700344fb1

src/viam/sdk/tests/test_common.cpp
```diff
@@ -0,0 +1,96 @@
...
+BOOST_AUTO_TEST_CASE(test_zero) {
+    Duration input;
+    input.set_nanos(0);
+    input.set_seconds(0);
+    auto converted = from_proto(input);
+    BOOST_CHECK_EQUAL(converted.count(), 0);
+    auto reconverted = to_proto(converted);
+    BOOST_CHECK_EQUAL(reconverted.nanos(), 0);
+    BOOST_CHECK_EQUAL(reconverted.seconds(), 0);
+}
```
zaporter-work, line 30: This test logic could be refactored into a shared helper function but then the error message points to the wrong line and it is a bit of a pain to understand.

acmorrow (approved):
LGTM!

PR #100, opened 2023-05-11
RSDK-2758 Consistent error handling & exception scheme

zaporter-work:
Ticket: https://viam.atlassian.net/browse/RSDK-2758

This provides a base `ViamException` class and a few derived classes. The purpose of these exceptions are threefold: 
- To provide a base to build on if we wish to pursue more advanced error creation in the future.
- To help users easily understand if an error came from our code or not.
- To facilitate error-catching and recovery behavior (ex: a `DuplicateResourceException` might be handled differently than a `PermissionDeniedException`)

This PR also creates `ViamErrorCode`s which are 1 to 1 compatible with the exceptions and can be used interchangeably (though conversion will lose the message inside the exception). 

While working on this, I experimented with three main design options:
1)
```
   template <typename... Args>
    ViamException(Args const&... what)
        : std::runtime_error("ViamException: " + concat_varargs(what...)){};
```
I really liked this approach because it let me write very clean constructor calls. However this produced awful error messages (SFINAE was overkill for concat_varargs), and it required almost everything to live in the header files.

2)
```
#define THROW_EXCEPTION(e_type, msg) throw e_type(__FILE__, __LINE__, __func__, msg)

// ex: THROW_EXCEPTION(DuplicateResourceException, "can't have two camera servers ...");
```

I really liked this approach because it gave very nice error messages; however, I am hesitant to suggest that we use macros. (we might also need to do some macro magic so that the full path of `__FILE__` doesn't end up in the binary)

3) 
The simple
```
ViamException(const std::string& msg) : std::runtime_exception("ViamException: " + msg) {}
```

Ultimately, I landed on number 3 because number 1 didn't end up providing as many benefits as I wanted and we can do number 2 later if we want to give more detailed error messages.

Important Notes:
1) Some of the exception types are not being used right now. I still created them because they may be useful in the future (and because they line up roughly with the python sdk)
2) Conversion to and from error codes is a bit awkward. Wasn't sure about the best way to do this.


src/viam/sdk/config/resource.cpp
```diff
@@ -18,11 +19,7 @@ namespace viam {
 namespace sdk {
 
 Name ResourceConfig::resource_name() {
-    try {
-        this->fix_api();
-    } catch (std::string err) {  // NOLINT
-        throw err;
-    }
```
zaporter-work, line 25: Does this do something that I don't understand?

src/viam/sdk/rpc/dial.cpp
```diff
@@ -66,9 +68,7 @@ std::shared_ptr<ViamChannel> ViamChannel::dial(const char* uri,
     char* socket_path = ::dial(uri, payload, opts.allows_insecure_downgrade(), ptr);
     if (socket_path == NULL) {
         free_rust_runtime(ptr);
-        // TODO(RSDK-1742) Replace throwing of strings with throwing of
-        // runtime_error
-        throw "Unable to establish connecting path!";
```
zaporter-work, line 71: Ticket was marked as completed

src/viam/sdk/config/resource.cpp
stuqdog: Probably not! Looks like the `try/catch` is pretty much meaningless there, and catching a `str` is just an artifact of poor initial design decisions. Makes sense to remove.

src/viam/sdk/common/exception.hpp
```diff
@@ -0,0 +1,68 @@
...
+#include <stdexcept>
+#include <string>
+
+namespace viam {
+namespace sdk {
+
+enum class ViamErrorCode : uint8_t {
+    // Used to indicate no error
+    ok = 0,
+    generic = 1,
```
stuqdog, line 15: What does `generic` mean here? I think `unknown` is more typical as a catch-all from what I've seen.

src/viam/sdk/common/exception.cpp
```diff
@@ -0,0 +1,69 @@
...
+    : std::runtime_error("ViamException: " + what), _error_code(ViamErrorCode::generic){};
+
+ViamException::ViamException(const std::string& type,
+                             const ViamErrorCode& code,
+                             const std::string& what)
+    : std::runtime_error("ViamException(" + type + "): " + what), _error_code(code){};
+
+ViamException::~ViamException() = default;
+ViamException::ViamException(const ::grpc::Status& status_code)
+
```
stuqdog, line 16: (nit) why is this newline here?

src/viam/sdk/common/exception.cpp
```diff
@@ -0,0 +1,69 @@
...
+    : ViamException(status_code.error_message() + " " + status_code.error_details()) {}
+
+ViamException ViamException::from_viam_error_code(ViamErrorCode code) {
+    switch (code) {
+        case ViamErrorCode::ok: {
+            return ViamException(
+                "Ran ViamException::from_viam_error_code on an \'ok\' value. Check the code "
+                "value "
```
stuqdog, line 24: (nit) breaking these up into one word per line is a bit strange looking. Probably good to combine this and the following line.

src/viam/sdk/common/exception.hpp
```diff
@@ -0,0 +1,68 @@
...
+    explicit ViamException(const ::grpc::Status& status_code);
+
+    static ViamException from_viam_error_code(ViamErrorCode code);
+
+    virtual ~ViamException();
+
+    ViamErrorCode get_error_code() const noexcept;
+
+   private:
+    ViamErrorCode _error_code;
```
stuqdog, line 40: (q) why did you go with a `_` prefix? Elsewhere in the codebase we've been using a suffix pretty consistently.

src/viam/sdk/components/generic/generic.hpp
```diff
@@ -5,6 +5,7 @@
 
 #include <google/protobuf/descriptor.h>
 
+#include <viam/sdk/common/exception.hpp>
```
stuqdog, line 8: (q) why include this if it's never used in the header?

stuqdog (approved):
This is great, thanks for getting this done! LGTM, % questions/nits.

src/viam/sdk/common/exception.hpp
zaporter-work: Good idea, `unknown` is clearer.

src/viam/sdk/common/exception.hpp
zaporter-work: Because I accidentally got confused with python. No reason other than foolishness

src/viam/sdk/components/generic/generic.hpp
zaporter-work: No reason -- we should look into adding that tool that Andrew mentioned which ensures that headers match the code.

zaporter-work:
Closing due to 2 months of no activity. Happy to talk about or reopen this if there is renewed interest.

PR #101, opened 2023-05-11
RSDK-2096 - miscellaneous improvements

stuqdog:
#### Major Changes
Fix ProtoType conversions, write tests to confirm they work

#### Minor changes
Lots of flybys, hopefully self-evident but I tried to give some description in comments.


src/viam/sdk/common/proto_type.cpp
```diff
@@ -37,44 +37,45 @@ AttributeMap struct_to_map(Struct struct_) {
...
         map->emplace(key, value);
     }
     return map;
 }
 
-// float, string, struct, array, struct(map from string to any of the above)
-
-// NOLINTNEXTLINE(misc-no-recursion)
-ProtoType ProtoType::of_value(Value value) {
+ProtoType::ProtoType(const Value& value) {
```
stuqdog, line 46: (flyby) constructor instead of static method.

src/viam/sdk/common/proto_type.cpp
```diff
@@ -170,16 +172,21 @@ bool operator==(const ProtoType& lhs, const ProtoType& rhs) {
                 return false;
             }
             // NOLINTNEXTLINE(misc-no-recursion)
-            auto pred = [](auto lhs_map, auto rhs_map) {
-                return lhs_map.first == rhs_map.first && *lhs_map.second == *rhs_map.second;
-            };
-
-            return std::equal(lhs_map->begin(), lhs_map->end(), rhs_map->begin(), pred);
+            for (const auto& kv : *lhs_map) {
```
stuqdog, line 175: This map comparison was failing on larger maps because the iteration order seemed inconsistent. This works now.

src/viam/sdk/common/utils.hpp
```diff
@@ -16,15 +16,14 @@ namespace viam {
 namespace sdk {
 
 using viam::common::v1::ResourceName;
-const std::string COMPONENT = "component";
```
stuqdog, line 19: (flyby) global constants follow the google style guide for naming

src/viam/sdk/module/service.cpp
```diff
@@ -249,6 +249,7 @@ void ModuleService_::add_model_from_registry(std::shared_ptr<Server> server, API
     if (services.find(api) == services.end()) {
         add_api_from_registry(server, api);
     }
+    const std::lock_guard<std::mutex> lock(lock_);
```
stuqdog, line 252: (put lock below the call to `add_api_from_registry` since `add_api_from_registry` locks as well).

src/viam/sdk/config/resource.cpp
```diff
@@ -110,7 +110,7 @@ ResourceConfig ResourceConfig::from_proto(viam::app::v1::ComponentConfig proto_c
     }
 
     if (proto_cfg.has_frame()) {
-        LinkConfig lc = LinkConfig::from_proto(proto_cfg.frame());
+        resource.frame_ = LinkConfig::from_proto(proto_cfg.frame());
```
zaporter-work, line 113: Is this diff tested?

zaporter-work (approved):
Looks good. We should make sure that we use `kVarName` consistently if that is what we settle on.

src/viam/sdk/config/resource.cpp
stuqdog: Good question! I don't think this is actually tested anywhere at the moment. I'll make a ticket for adding those tests.

src/viam/sdk/module/service.cpp
stuqdog: Good question. I'm not too worried right now, since everything is synchronous and this is taking the lock immediately after `add_api_from_registry` returns. At a minimum this `lock` makes race conditions less likely at least. Might be worth thinking about long term though!

acmorrow: Potential bonus points if you make the `inlock` versions take a `const std::lock_guard<mutex>&` as a proof of work, but I don't find it that useful because it doesn't ensure that you have actually taken the right lock.

acmorrow (approved):
LGTM with some small comments and suggestions.

src/viam/sdk/module/service.cpp
```diff
@@ -229,8 +229,9 @@ void ModuleService_::close() {
     }
 }
 
-void ModuleService_::add_api_from_registry(std::shared_ptr<Server> server, API api) {
-    const std::lock_guard<std::mutex> lock(lock_);
+void ModuleService_::_add_api_from_registry_inlock(std::shared_ptr<Server> server,
```
acmorrow, line 232: I think convention in this repo is trailing `_` for private, so I'd call this `add_api_from_registry_inlock_`

acmorrow (approved):
LGTM with a tiny naming nit.

PR #102, opened 2023-05-11
Improve docs workflow & point to README

zaporter-work:
This PR does two main things:

1) Get rid of `etc/apidocmenu.md` and has doxygen point to the main `README.md` instead. This is consistent with the other SDKs and provides slightly better docs.

2) Creates a `./bin/build-docs.sh` file to improve ergonomics and has the `README.md` point there. I kept forgetting to make the `etc/docs/api/current` folder and have been mildly inconvenienced enough to automate the process.

Arguably, this should be two PRs but it was simple enough that I just did it in one. If people are against editing the README.md to point to this new script, I am happy to keep the README as it was (I have no strong opinions on this).


bin/build-docs.sh
```diff
@@ -0,0 +1,20 @@
...
+
+RELATIVE_OUTPUT="etc/docs/api/current"
+
+mkdir -p "$RELATIVE_OUTPUT"
+
+doxygen Doxyfile
+
+popd
+
+echo
```
zaporter-work, line 18: Extra echo to add a newline sep

PR #103, opened 2023-05-11
RSDK-3102 - Backport constructor changes to previous components

zaporter-work:
Ticket: https://viam.atlassian.net/browse/RSDK-3102

As we have created more components, we have changed how constructors are defined to prevent inlining (part of our styleguide) and to add explicit annotations to constructors when possible.

This brings those changes to the previous components.

### DIFF
--- a/src/viam/sdk/components/base/base.cpp
+++ b/src/viam/sdk/components/base/base.cpp
@@ -14,10 +14,8 @@
 namespace viam {
 namespace sdk {
 
-BaseClient::BaseClient(std::string name, std::shared_ptr<grpc::Channel> channel)
-    : Base(std::move(name)),
-      stub_(viam::component::base::v1::BaseService::NewStub(channel)),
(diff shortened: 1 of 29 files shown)


src/viam/sdk/components/base/base.cpp
```diff
@@ -14,10 +14,8 @@
 namespace viam {
 namespace sdk {
 
-BaseClient::BaseClient(std::string name, std::shared_ptr<grpc::Channel> channel)
-    : Base(std::move(name)),
-      stub_(viam::component::base::v1::BaseService::NewStub(channel)),
-      channel_(std::move(channel)){};
```
zaporter-work, line 20: This should not have been defined here in the first place.

acmorrow (approved):
I didn't read too closely because it looks largely mechanical, but I agree this is a good thing to do. LGTM.

PR #104, opened 2023-05-19
RSDK-2675 Add support for MLModelService to C++ SDK

acmorrow:
Marking this as a draft because there are still a few TODO items. In particular, the `extra` field in the `metadata` doesn't work, I think because there are issues in the `AttributeMap` type that are hopefully fixed by #101. Other notes:
- I still need to add a round-trip test for Inference.
- We need to vendor xtensor (header only library, so this is easy) and have CMake have options for using system vs vendored.
- The example mlmodel doesn't do anything interesting yet. It just returns a static bounding box and doesn't consult the inputs. I intend to actually make it integrate with the tensorflow lite model to perform effective detection.

Finally, note that the RDK vision service cannot currently make use of mlmodels implemented as modular resources with the C++ SDK due to https://viam.atlassian.net/browse/RSDK-3096. That'll need to be fixed before this can actually be used. I've managed to work around it by hacking up `attemptToBuildDetector` in the RDK. That's enough to get an end-to-end demo up that can the static bounding box on a transform camera in a robot, so I do have some confidence that this is basically working.

I'll update this from draft to a real PR once I've added the round-trip Inference test, but I'd definitely appreciate feedback on it in parallel.

### DIFF
--- a/src/viam/sdk/CMakeLists.txt
+++ b/src/viam/sdk/CMakeLists.txt
@@ -193,7 +193,7 @@ target_link_libraries(viamsdk
   PUBLIC ${VIAMCPPSDK_PROTOBUF_LIBRARIES}
   PRIVATE ${VIAMCPPSDK_GRPCXX_REFLECTION_LIBRARIES}
   PUBLIC Boost::headers
-  PRIVATE Boost::log
+  PUBLIC Boost::log
   PRIVATE viam_rust_utils
   PRIVATE Threads::Threads
 )
--- a/src/viam/examples/modules/example_module.cpp
+++ b/src/viam/examples/modules/example_module.cpp
@@ -3,6 +3,7 @@
 #include <pthread.h>
 #include <signal.h>
 
+#include <boost/log/trivial.hpp>
 #include <grpcpp/grpcpp.h>
 #include <grpcpp/server_context.h>
 
@@ -79,6 +80,11 @@ int main(int argc, char** argv) {
     if (argc < 2) {
         throw "need socket path as command line argument";
     }
+
+    // Use set_logger_severity_from_args to set the boost trivial logger's
+    // severity depending on commandline arguments.
+    set_logger_severity_from_args(argc, argv);
+    BOOST_LOG_TRIVIAL(debug) << "Starting module with debug level logging";
 
     // C++ modules must handle SIGINT and SIGTERM. Make sure to create a sigset
     // for SIGINT and SIGTERM that can be later awaited in a thread that cleanly
--- a/src/viam/sdk/common/utils.cpp
+++ b/src/viam/sdk/common/utils.cpp
@@ -5,6 +5,9 @@
 #include <vector>
 
 #include <boost/blank.hpp>
+#include <boost/log/core.hpp>
+#include <boost/log/expressions.hpp>
+#include <boost/log/trivial.hpp>
 #include <boost/optional/optional.hpp>
 #include <boost/variant/get.hpp>
 #include <boost/variant/variant.hpp>
@@ -80,6 +83,15 @@ google::protobuf::Duration to_proto(const std::chrono::microseconds& duration) {
     proto.set_nanos(static_cast<int32_t>(nanos.count()));
     proto.set_seconds(seconds.count());
     return proto;
+}
+
+void set_logger_severity_from_args(int argc, char** argv) {
+    if (argc >= 3 && strcmp(argv[2], "--log-level=debug") == 0) {
+        boost::log::core::get()->set_filter(boost::log::trivial::severity >=
+                                            boost::log::trivial::debug);
+        return;
+    }
+    boost::log::core::get()->set_filter(boost::log::trivial::severity >= boost::log::trivial::info);
 }
 
 }  // namespace sdk
--- a/src/viam/sdk/common/utils.hpp
+++ b/src/viam/sdk/common/utils.hpp
@@ -47,5 +47,15 @@ std::string bytes_to_string(std::vector<unsigned char> const& b);
 std::chrono::microseconds from_proto(const google::protobuf::Duration& proto);
 google::protobuf::Duration to_proto(const std::chrono::microseconds& duration);
 
+/// @brief Set the boost trivial logger's severity depending on args.
+/// @param argc The number of args.
+/// @param argv The commandline arguments to parse.
+///
+/// Sets the boost trivial logger's severity to debug if "--log-level=debug" is the
+/// the third argument. Sets severity to info otherwise. Useful for module
+/// implementations. See LogLevel documentation in the RDK for more information on
+/// how to start modules with a "log-level" commandline argument.
+void set_logger_severity_from_args(int argc, char** argv);
+
 }  // namespace sdk
 }  // namespace viam


CMakeLists.txt
```diff
@@ -121,7 +121,7 @@ endif()
 # info. This is probably the right build type for packaging and
 # release.
 if (NOT CMAKE_BUILD_TYPE)
-  set(CMAKE_BUILD_TYPE "RelWithDebInfo")
+  set(CMAKE_BUILD_TYPE "Debug")
```
acmorrow, line 124: Oops, I didn't mean to commit this. I'll back it out.

etc/docker/Dockerfile.debian.sid
```diff
@@ -22,6 +22,7 @@ RUN apt-get -y --no-install-recommends install \
     libgrpc++-dev \
     libprotobuf-dev \
     libssl-dev \
+    libxtensor-dev \
```
acmorrow, line 25: This is available in brew for macos. For `bookworm` this will be a builtin, but we will need to vendor xtensor for other platofrms.

src/viam/cmake/viamcppsdk_utils.cmake
```diff
@@ -2,13 +2,13 @@ function(viamcppsdk_add_boost_test SOURCE_FILE_NAME)
   get_filename_component(TEST_EXECUTABLE_NAME ${SOURCE_FILE_NAME} NAME_WE)
 
   add_executable(${TEST_EXECUTABLE_NAME} ${SOURCE_FILE_NAME})
-  target_link_libraries(${TEST_EXECUTABLE_NAME} viam-cpp-sdk::viamsdk)
+  target_link_libraries(${TEST_EXECUTABLE_NAME} viamsdk_test)
```
acmorrow, line 5: See below for details on `viamsdk-test`.

src/viam/sdk/CMakeLists.txt
```diff
@@ -79,18 +79,14 @@ target_sources(viamsdk
...
     rpc/server.cpp
+    services/mlmodel/client.cpp
+    services/mlmodel/mlmodel.cpp
+    services/mlmodel/private/proto.cpp
+    services/mlmodel/server.cpp
     services/service.cpp
     spatialmath/geometry.cpp
     spatialmath/orientation_types.cpp
     spatialmath/orientation.cpp
-    tests/test_utils.cpp
```
acmorrow, line 86: Mocks and tests didn't belong in the SDK library. I've moved them to their own library.

src/viam/sdk/CMakeLists.txt
```diff
@@ -132,18 +128,13 @@ target_sources(viamsdk
...
       ../../viam/sdk/rpc/dial.hpp
       ../../viam/sdk/rpc/server.hpp
+      ../../viam/sdk/services/mlmodel/client.hpp
+      ../../viam/sdk/services/mlmodel/mlmodel.hpp
+      ../../viam/sdk/services/mlmodel/server.hpp
       ../../viam/sdk/services/service.hpp
       ../../viam/sdk/spatialmath/geometry.hpp
       ../../viam/sdk/spatialmath/orientation_types.hpp
       ../../viam/sdk/spatialmath/orientation.hpp
-      ../../viam/sdk/tests/test_utils.hpp
```
acmorrow, line 139: Ditto (and even maybe moreso) for headers for mocks and tests.

src/viam/sdk/common/proto_type.hpp
```diff
@@ -8,6 +8,9 @@
 namespace viam {
 namespace sdk {
 
+class ProtoType;
+using AttributeMap = std::shared_ptr<std::unordered_map<std::string, std::shared_ptr<ProtoType>>>;
```
acmorrow, line 12: Drive-by. This typedef is more useful in this header than in `resource.hpp` and moving it here simplified things a bit. There is more work needed on `ProtoType`, some of which is in #101. I've resisted the urge to make those changes here in favor of waiting on that PR.

src/viam/sdk/resource/resource_manager.cpp
```diff
@@ -35,7 +35,8 @@ std::shared_ptr<Resource> ResourceManager::resource(std::string name) {
             return resources_.at(short_name);
         }
     }
-    throw std::runtime_error("Unable to find resource named " + name);
```
acmorrow, line 38: This was a contract violation with the existing style of code which expects to check the return value of `ResourceManager::resource` rather than dealing with an exception. We can reverse that decision but it will require updating all the existing callers of `ResourceManager::resource`.

src/viam/sdk/services/mlmodel/client.cpp
```diff
@@ -0,0 +1,158 @@
...
+        const auto where = output_fields.find(output.name);
+        // Ignore any outputs for which we don't have metadata, since
+        // we can't know what type they should decode to.
+        if (where != output_fields.end()) {
+            mlmodel_details::pb_value_to_tensor(
+                output, where->second, &tsav->storage, &tsav->views);
+        }
+    }
+
+    return {std::move(tsav), &tsav->views};
```
acmorrow, line 81: Tricky: this shared pointer keeps `tsav` alive but returns a shared_ptr which refers to the tensor views that `tsav` manages.

I'm pretty sure this works but I intend to write some tests to make sure.

src/viam/sdk/services/mlmodel/client.cpp
```diff
@@ -0,0 +1,158 @@
...
+        pb::Value& value = ib.first->second;
+        mlmodel_details::tensor_to_pb_value(kv.second, &value);
+    }
+
+    auto* const resp = pb::Arena::CreateMessage<mlpb::InferResponse>(&arena);
+
+    grpc::ClientContext ctx;
+    auto result = stub_->Infer(&ctx, *req, resp);
+
+    // TODO: Metadata caching.
```
acmorrow, line 66: This is fairly important because otherwise we need to make a metadata call for each infer call, and we don't want to need to do that. But it wasn't a requirement to have that working in order to move forward to review.

src/viam/examples/modules/example_mlmodel.cpp
```diff
@@ -0,0 +1,257 @@
...
+                     // shape
+                     {},
+
+                     // associated_files
+                     {},
+
+                     // extra
+                     {}
+
+                     // TODO, once PR 101
```
acmorrow, line 123: Per comment, the `extra` field cannot be dealt with until #101 lands.

src/viam/sdk/services/mlmodel/client.cpp
```diff
@@ -0,0 +1,158 @@
...
+                        default: {
+                            std::ostringstream message;
+                            message << "Failed to deserialize returned "
+                                       "Metadata.TensorInfo.File.label_type field with value `"
+                                    << af.label_type() << "` to one of the known label types";
+                            throw std::runtime_error(message.str());
+                        }
+                    }
+                }
+                // TODO: Currently, struct_to_map seems broken,
```
acmorrow, line 143: Waiting on #101

src/viam/sdk/services/mlmodel/mlmodel.cpp
```diff
@@ -0,0 +1,124 @@
...
+    return std::make_shared<MLModelServiceRegistration>(sd);
+}
+
+API MLModelService::dynamic_api() const {
+    return static_api();
+}
+
+boost::optional<enum MLModelService::tensor_info::data_type>
+MLModelService::tensor_info::string_to_data_type(const std::string& str) {
+    // TODO: I'm guessing at the string names here, because the scope
```
acmorrow, line 58: If we revise the scope per https://viam.atlassian.net/browse/RSDK-3161, then perhaps we can make the data types an enum instead. If not, hopefully we can at least update the existing scope to list the supported values.

src/viam/sdk/services/mlmodel/mlmodel.cpp
```diff
@@ -0,0 +1,124 @@
...
+    } else if (str == "float32") {
+        return data_type::k_float32;
+    } else if (str == "float64") {
+        return data_type::k_float64;
+    }
+    return {};
+}
+
+const char* MLModelService::tensor_info::data_type_to_string(const enum data_type data_type) {
+    // TODO: I'm guessing at the string names here, because the scope
```
acmorrow, line 85: As above.

src/viam/sdk/tests/CMakeLists.txt
```diff
@@ -13,14 +13,34 @@
 # limitations under the License.
 
 enable_testing()
+
 include(viamcppsdk_utils)
+add_library(viamsdk_test)
+
+target_sources(viamsdk_test
```
acmorrow, line 20: This is the new `viamsdk_test` library that contains mocks and test utilities so they don't end up in the SDK library.

src/viam/sdk/tests/mocks/mlmodel_mocks.hpp
```diff
@@ -0,0 +1,81 @@
...
+    std::shared_ptr<named_tensor_views> infer(const named_tensor_views& inputs) override;
+
+    MockMLModelService& metadata(struct metadata metadata);
+    struct metadata metadata() override;
+
+   private:
+    struct metadata metadata_;
+};
+
+#if 0
```
acmorrow, line 38: I believe this can be removed, it was copypasta from another mock but I think given Zack's new testing approach it isn't needed anymore.

src/viam/sdk/tests/test_mlmodel.cpp
```diff
@@ -0,0 +1,308 @@
...
+    server->Shutdown();
+}
+
+BOOST_AUTO_TEST_CASE(mock_metadata_grpc_roundtrip) {
+    auto mock = std::make_shared<MockMLModelService>();
+    mock->metadata(test_metadata);
+    client_server_test(mock, [](auto& client) { BOOST_TEST(test_metadata == client.metadata()); });
+}
+
+BOOST_AUTO_TEST_CASE(mock_infer_grpc_roundtrip) {
```
acmorrow, line 208: This test does not currently pass, I will push an update to comment it out.

etc/docker/Dockerfile.debian.sid
acmorrow: Also, we can _potentially_ drop the xtensor requirement and look into other options (maybe something already in boost). I originally picked xtensor because it was going to allow me to sew together multiple independent allocations under one tensor view with the `xchunked_array`. I thought I was going to need that at one point, but that was due to a misunderstanding of mine about how the tensor data was arriving over the wire. So we may be able to use something simpler. On the other hand, `xtensor` is pretty nice.

src/viam/sdk/resource/resource_manager.cpp
zaporter-work: [opt] I think I wrote the examples back when throwing was the default behavior for a missing resource. Might be a good idea to add nullptr checks for those (maybe in a different PR)

src/viam/sdk/services/mlmodel/mlmodel.hpp
```diff
@@ -0,0 +1,156 @@
...
+    using tensor_view_types =
+        boost::mpl::transform_view<base_types, make_tensor_view_<boost::mpl::placeholders::_1>>;
+
+    // Union the tensor views for the various base types.
+    using tensor_views = boost::make_variant_over<tensor_view_types>::type;
+
+    // Our parameters to and from the model come as named tensor_views.
+    using named_tensor_views = std::unordered_map<std::string, tensor_views>;
+
+    /// @brief Runs the model against the input tensors and returns inference results as tensosrs.
```
zaporter-work, line 101:
```suggestion
    /// @brief Runs the model against the input tensors and returns inference results as tensors
```

src/viam/sdk/services/mlmodel/client.hpp
```diff
@@ -0,0 +1,47 @@
...
+   public:
+    using service_type = viam::service::mlmodel::v1::MLModelService;
+
+    MLModelServiceClient(std::string name, std::shared_ptr<grpc::Channel> channel);
+
+    std::shared_ptr<named_tensor_views> infer(const named_tensor_views& inputs) override;
+    struct metadata metadata() override;
+
+   protected:
+    MLModelServiceClient(std::string name, std::unique_ptr<service_type::StubInterface> stub);
```
zaporter-work, line 39: [q] Where is this used? This reminds me of one of the old testing methods

src/viam/sdk/services/mlmodel/private/proto.cpp
```diff
@@ -0,0 +1,333 @@
...
+#include <boost/variant/get.hpp>
+
+namespace viam {
+namespace sdk {
+namespace mlmodel_details {
+
+namespace {
+
+template <typename T>
+::grpc::Status pb_value_to_tensor_t(const MLModelService::tensor_info& tensor_info,
```
acmorrow, line 28: Note that these methods (`pb_value_to_tensor_t` and `tensor_to_pb_`) do not yet have good test coverage. I wouldn't spend a ton of time worrying about correctness for them just yet, as I'd consider them very likely to have a few issues.

src/viam/sdk/services/mlmodel/mlmodel.hpp
acmorrow: Done

src/viam/sdk/services/mlmodel/client.hpp
acmorrow: It isn't. leftovers from before, and I've now removed it.

acmorrow: Oops, no. I take that back. It *is* used. I'll drop a comment on where.

acmorrow:
@stuqdog, @bhaney, and @zaporter-work (if you are interested) - I've promoted this to a real review as I have some increased confidence that the algorithms to convert to/from proto are more correct after adding some additional testing (and fixing the bugs the testing revealed). PTAL!

stuqdog (commented):
Initial comments from walkthrough. More thorough review to come.

src/viam/examples/modules/example_mlmodel.cpp
```diff
@@ -0,0 +1,246 @@
...
+                "C++ SDK Example MLModel - Faking EfficientDet Lite0 V1",
+
+                // type
+                "fake_tflite_detector",
+
+                // description
+                "Identify which of a known set of objects might be present and provide information "
+                "about their positions within the given image or a video stream.",
+
+                // `inputs`
```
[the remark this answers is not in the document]
acmorrow: Done.

src/viam/examples/modules/example_mlmodel.cpp
```diff
@@ -0,0 +1,246 @@
...
+                     "location",
+
+                     // description
+                     "The locations of the detected boxes.",
+
+                     // data_type
+                     tensor_info::data_types::k_float32,
+
+                     // shape
+                     {},
```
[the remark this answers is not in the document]
acmorrow: Added the following comment:
```
                  //
                  // TODO(RSDK-3033): Once tflite_cpu reports output
                  // shape metadata, fill this in.
```

src/viam/sdk/services/mlmodel/client.cpp
```diff
@@ -0,0 +1,159 @@
...
+                        case ::viam::service::mlmodel::v1::LABEL_TYPE_TENSOR_AXIS: {
+                            new_file.label_type = tensor_info::file::k_label_type_tensor_axis;
+                            break;
+                        }
+                        default: {
+                            std::ostringstream message;
+                            message << "Failed to deserialize returned "
+                                       "Metadata.TensorInfo.File.label_type field with value `"
+                                    << af.label_type() << "` to one of the known label types";
+                            throw std::runtime_error(message.str());
```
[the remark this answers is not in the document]
acmorrow: Not yet merged, but I added `// TODO: Throw viam exception once PR #100 merges.` to all new `throw` instances in the PR.

src/viam/sdk/services/mlmodel/mlmodel.hpp
```diff
@@ -0,0 +1,163 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#pragma once
+
+#include <boost/mpl/joint_view.hpp>
+#include <boost/mpl/list.hpp>
+#include <boost/mpl/transform_view.hpp>
+
```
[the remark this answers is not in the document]
acmorrow: Done

src/viam/sdk/services/mlmodel/server.cpp
```diff
@@ -0,0 +1,186 @@
...
+    // bubble up and have a higher layer deal with it? Perhaps all of
+    // our server side overrides should be `noexcept`.
+    const auto outputs = mlms->infer(inputs);
+    auto& pb_output_data_fields = *(response->mutable_output_data()->mutable_fields());
+    for (const auto& kv : *outputs) {
+        auto emplace_result = pb_output_data_fields.try_emplace(kv.first);
+        // This assert should be impossible: `outputs` is a map and we
+        // are iterating its unique keys, and our `InferResponse`
+        // should have empty output data fields.
+        assert(emplace_result.second);
```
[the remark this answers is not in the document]
acmorrow: Filed https://viam.atlassian.net/browse/RSDK-3285

src/viam/cmake/viamcppsdk_utils.cmake
acmorrow: We discussed this in person, resolving.

src/viam/sdk/services/mlmodel/client.cpp
tessavitabile: Is it safe to cache metadata? What if the model service has been reconfigured?

tessavitabile (commented):
Looks great! Added one question, but please don't wait for my sign-off.

src/viam/examples/modules/example_mlmodel.cpp
```diff
@@ -0,0 +1,240 @@
...
+                  "The locations of the detected boxes.",
+
+                  // `data_type`
+                  tensor_info::data_types::k_float32,
+
+                  // `shape`
+                  //
+                  // TODO(RSDK-3033): Once tflite_cpu reports output
+                  // shape metadata, fill this in.
+                  {},
```
bhaney, line 107: can't you fill this in anyway, since the API has the field available (even if the model can't do anything with it yet)?

src/viam/sdk/services/mlmodel/client.cpp
```diff
@@ -0,0 +1,158 @@
...
+
+struct tensor_storage_and_views {
+    mlmodel_details::tensor_storage storage;
+    MLModelService::named_tensor_views views;
+};
+
+}  // namespace
+
+MLModelServiceClient::MLModelServiceClient(std::string name, std::shared_ptr<grpc::Channel> channel)
+    : MLModelServiceClient(std::move(name), service_type::NewStub(channel)) {
```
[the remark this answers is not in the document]
zaporter-work: Got it

src/viam/sdk/services/mlmodel/client.cpp
```diff
@@ -0,0 +1,159 @@
...
+
+    auto tsav = std::make_shared<tensor_storage_and_views>();
+    const auto& output_fields = resp->output_data().fields();
+    for (const auto& output : md.outputs) {
+        const auto where = output_fields.find(output.name);
+        // Ignore any outputs for which we don't have metadata, since
+        // we can't know what type they should decode to.
+        if (where != output_fields.end()) {
+            mlmodel_details::pb_value_to_tensor(
+                output, where->second, &tsav->storage, &tsav->views);
```
bhaney, line 77: This makes sense to me, and also makes me glad you're doing the proof of concept to try to eliminate having to rely on the metadata for the output type

bhaney (approved):
I think all of this makes sense, and given how we are probably going to extend the proto with more specified fields, a lot of how infer is packed/unpacked will become easier as well.

src/viam/sdk/services/mlmodel/client.cpp
acmorrow: Yes, this is the concern I brought up last week in the meeting. The explanation I received was that it should be OK to cache, because if an mlmodel service is reconfigured then components which depend on it should also be reconfigured, and so the client would be re-created - effectively dropping the cache. If that logic isn't sound, then some time will need to be invested in considering solutions. For now, we fetch the metadata on each inference, which isn't great, and may also not be safe across reconfigurations if those are actually a problem, since you could race between obtaining the metadata and a reconfiguration.

src/viam/examples/modules/example_mlmodel.cpp
acmorrow: I've rebased above #101 and added handling for the `extra` field.

src/viam/sdk/tests/mocks/mlmodel_mocks.hpp
acmorrow: This has been removed.

src/viam/sdk/tests/test_mlmodel.cpp
acmorrow: This test has been implemented and now passses.

acmorrow:
Updated, the infer grpc round trip test is now implemented and passes. There are various other bug fixes and improvements since the prior round of review.

src/viam/sdk/services/mlmodel/client.cpp
tessavitabile: This is true for the vision service, when it depends on an ML model service. But this won't be true for an external client using the C++ SDK.

acmorrow: That's a very good point which I had not considered. A few ways I see to make it work would be to 1) require fetching metadata on every inference as we are doing now, or 2) move only the necessary subset of metadata into the inference response (e.g. the `shape` info in the proposed flat tensors scope is already a move in this direction), or 3) including a version field in the metadata response and returning the metadata version in the inference response,  which would allow clients to invalidate and refresh the cache. I suspect that 3 is the right answer since 1 is probably unacceptable for performance reasons, and 2 feels vague. The downside to three is that it places requirements on servers to manage the metadata version properly, and on clients to validate and refresh. Definitely interested if you have any other ideas how we might approach the issue.

Meanwhile, I will update the comment to say "Investigate" and fold in some of this discussion.

acmorrow: Filed https://viam.atlassian.net/browse/RSDK-3298, and I'll update the comment.

src/viam/sdk/tests/mocks/mlmodel_mocks.cpp
```diff
@@ -0,0 +1,40 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#include <viam/sdk/tests/mocks/mlmodel_mocks.hpp>
+
+namespace viam {
+namespace sdktests {
+
+MockMLModelService& MockMLModelService::set_infer_handler(infer_handler handler) {
```
stuqdog, line 20: (q) It doesn't look to me like the return value of this is used in the tests. Curious why you have it return the ptr instead of being void.

src/viam/sdk/services/mlmodel/client.cpp
```diff
@@ -0,0 +1,159 @@
...
+                auto& ti = target.back();
+                ti.name = std::move(*s.mutable_name());
+                ti.description = std::move(*s.mutable_description());
+                auto data_type = MLModelService::tensor_info::string_to_data_type(s.data_type());
+                if (!data_type) {
+                    std::ostringstream message;
+                    message << "Failed to deserialize returned Metadata.TensorInfo.data_type field "
+                               "with value `"
+                            << s.data_type() << "` to one of the known tensor data types";
+                    // TODO: Throw viam exception once PR #100 merges.
```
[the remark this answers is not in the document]
acmorrow: Indeed it is, sorry for the delay on that.

src/viam/sdk/services/mlmodel/private/proto.cpp
```diff
@@ -0,0 +1,363 @@
...
+    // Seed our shape vector as starting with one unknown dimension.
+    // We will populate it while walking the protobuf. Also create a
+    // stack of Value pointers which we will use to manage our
+    // walk. Start a depth counter so we know the index of the
+    // dimension we are iterating.
+    typename MLModelService::tensor_view<T>::shape_type shape{};
+    std::stack<const ::google::protobuf::Value*> vs{{&pb}};
+    size_t depth = 0;
+
+    while (!vs.empty()) {
```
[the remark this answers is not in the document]
acmorrow: I suspect that this loop will be short-lived, as we hope to replace Struct with flat tensors per https://viam.atlassian.net/browse/RSDK-3161. If that doesn't work out for some reason, I will make sure to come back through and revisit these loops. I'm actually not entirely pleased with them anyway and I suspect they can be simplified / improved. But I didn't want to invest further time in improving them if we are going to remove them soon. Which I truly hope we can.

src/viam/sdk/services/mlmodel/private/proto.cpp
```diff
@@ -0,0 +1,363 @@
...
+        if (!vs.top()) {
+            vs.pop();
+            --depth;
```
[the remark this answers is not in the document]
acmorrow: The decrement at zero wouldn't be UB because unsigned integers wrap. So you would end up with a huge positive value. If you used that to index into `shape` that'd definitely be undefined behavior. In practice you would almost certainly observe a crash. That said, I'm not too worried about it getting out of sync because there is only one increment and one decrement, and they are paired fairly well with the marking of the top of the stack as nullptr. See also my comment above about the future of these loops.

src/viam/sdk/services/mlmodel/private/proto.cpp
```diff
@@ -0,0 +1,363 @@
...
+        }
+        value_->set_allocated_list_value(lvs.top().release());
+        lvs.pop();
+        return ::grpc::Status();
+    }
+
+    // A similar tricky bit of work, but for the special case of
+    // `byte` tensors where we must honor golang protobuf convention
+    // of storing byte arrays as Base64 encoded strings.
+    ::grpc::Status tensor_to_pb_(const MLModelService::tensor_view<std::uint8_t>& tensor) const {
```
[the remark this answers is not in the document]
acmorrow: I actually had a bunch of difficulty in this exact space where I was trying to mix templates and overloads and it wasn't working. Let me try one more time. I share your distaste for the special casing via names.

src/viam/sdk/services/mlmodel/server.cpp
```diff
@@ -0,0 +1,191 @@
...
+    // TODO: Should we handle exceptions here? Or is it ok to let them
+    // bubble up and have a higher layer deal with it? Perhaps all of
+    // our server side overrides should be `noexcept`.
+    //
+    // NOTE: Exceptions from here will result in an empty reply. We do
+    // need to handle them and convert into a failing reply status.
+    const auto outputs = mlms->infer(inputs);
+    auto& pb_output_data_fields = *(response->mutable_output_data()->mutable_fields());
+    for (const auto& kv : *outputs) {
+        auto emplace_result = pb_output_data_fields.try_emplace(kv.first);
```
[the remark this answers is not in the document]
acmorrow: I'm going to TODO this for now, since I will need to come back through and validate this entire change against bullseye / focal which will catch this, but I can't do that until I've made this work against vendored xtensor. I'd rather fix up any older proto version incompatibilities in one go.

src/viam/sdk/tests/mocks/mlmodel_mocks.cpp
acmorrow: Just because I like setters that chain:

`foo.set_x(42).set_y(true)`.

So I tend to write them that way, but in this case I didn't end up using it. I can remove it if you aren't in favor.

src/viam/sdk/services/mlmodel/client.cpp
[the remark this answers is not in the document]
acmorrow: Yes, that works. Done.

src/viam/sdk/services/mlmodel/private/proto.cpp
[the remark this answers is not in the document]
acmorrow: Yeah, thanks for the push. There was indeed a simpler way.

src/viam/sdk/services/mlmodel/client.cpp
acmorrow: Yes. I'm hopeful that with the type and shape in the tensor itself we can avoid needing metadata at all for decoding structs. We still might want it to validate against the expected shape though.

src/viam/examples/modules/example_mlmodel.cpp
acmorrow: @bhaney - Can you elaborate a bit on what you are proposing? I'm not quite getting it.

CMakeLists.txt
```diff
@@ -278,6 +278,8 @@ add_custom_target(
 set(VIAMCPPSDK_BOOST_VERSION_MINIMUM 1.71)
 set(VIAMCPPSDK_GRPC_VERSION_MINIMUM 1.30.2)
 set(VIAMCPPSDK_PROTOBUF_VERSION_MINIMUM 3.12.4)
+set(VIAMCPPSDK_XTENSOR_VERSION_MINIMUM 0.24.6)
+set(VIAMCPPSDK_XTENSOR_VERSION_MINIMUM 0.7.5)
```
acmorrow, line 282: This should say XTL

acmorrow: Also these versions need to be downgraded to match the current version available in sid.

src/viam/sdk/services/mlmodel/client.cpp
```diff
@@ -0,0 +1,159 @@
...
+
+    // Invoke the stub
+    grpc::ClientContext ctx;
+    viam::service::mlmodel::v1::MetadataResponse resp;
+    const auto stub_result = stub_->Metadata(&ctx, req, &resp);
+
+    struct metadata result;
+    if (resp.has_metadata()) {
+        auto& metadata_pb = *resp.mutable_metadata();
+        result.name = std::move(*metadata_pb.mutable_name());
```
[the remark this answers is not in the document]
stuqdog: That makes sense, thanks for explaining!

src/viam/sdk/services/mlmodel/client.cpp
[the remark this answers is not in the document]
stuqdog: No worries!

src/viam/sdk/services/mlmodel/private/proto.cpp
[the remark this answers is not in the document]
stuqdog: Sounds good to me, will resolve this convo.

zaporter-work (approved):
Not able to do a deep dive, but surface-level, it looks good to me!

acmorrow:
Thanks all for taking a look at this. There is one outstanding bit of work that needs to be handled before I can land this, because as it is right now it won't build on anything but MacOS and debian sid, both of which have xtensor packaged. But for our older platforms like Focal and Bullseye it will fail to build. The answer is for the SDK to have a vendored copy of xtensor which it can use if the system lacks it. That'll take some CMake-ing to get set up right. Alternatively, I could do a small amount of work to make supporting the MLModelService in the SDK conditional on having xtensor. Then we could merge this more or less as-is and tackle vendoring xtensor as follow on work. Please let me know which you would prefer.

stuqdog:
> That'll take some CMake-ing to get set up right. Alternatively, I could do a small amount of work to make supporting the MLModelService in the SDK conditional on having xtensor.

How much CMake work do you expect this to be? It feels a little code-smelly to have complicated build system work happening in this PR so if it's non-trivial I think it probably makes most sense to ship with the workaround and then file a separate ticket to update CMake. But, I defer to your sense of how complicated the vendoring fix will be.

acmorrow:
Hey reviewers I didn't mean to close this, I made an egregious git error of some sort and reset the remote branch to main. I will either re-open this or make a new one.

acmorrow:
Re-opened as #111, but I don't think any reviewers on here except @stuqdog need to be involved over there since everyone has already approved the work so far. The only changes that will happen in #111 will be for managing the xtensor dependency.

PR #109, opened 2023-05-25
RSDK-2725 Add set_logger_severity_from_args

benjirewis:
[RSDK-2725](https://viam.atlassian.net/browse/RSDK-2725)

Adds a new utility function `set_logger_severity_from_args` that allows users to set the boost trivial logger's severity to `debug` or `info` if `--log-level=debug` was the third command line argument.

Tested that `BOOST_LOG_TRIVIAL(debug)` statements are printed when server is started with -debug or module is configured with log_level: debug. Tested that `BOOST_LOG_TRIVIAL(debug)` statements are not printed when server is started with -debug but module is configured with `log_level: warn`.

[RSDK-2725]: https://viam.atlassian.net/browse/RSDK-2725?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/viam/sdk/CMakeLists.txt
+++ b/src/viam/sdk/CMakeLists.txt
@@ -193,7 +193,7 @@ target_link_libraries(viamsdk
   PUBLIC ${VIAMCPPSDK_PROTOBUF_LIBRARIES}
   PRIVATE ${VIAMCPPSDK_GRPCXX_REFLECTION_LIBRARIES}
   PUBLIC Boost::headers
-  PRIVATE Boost::log
+  PUBLIC Boost::log
   PRIVATE viam_rust_utils
   PRIVATE Threads::Threads
 )
--- a/src/viam/sdk/common/utils.cpp
+++ b/src/viam/sdk/common/utils.cpp
@@ -5,6 +5,9 @@
 #include <vector>
 
 #include <boost/blank.hpp>
+#include <boost/log/core.hpp>
+#include <boost/log/expressions.hpp>
+#include <boost/log/trivial.hpp>
 #include <boost/optional/optional.hpp>
 #include <boost/variant/get.hpp>
 #include <boost/variant/variant.hpp>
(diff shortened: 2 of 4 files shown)


src/viam/sdk/CMakeLists.txt
```diff
@@ -193,7 +193,7 @@ target_link_libraries(viamsdk
   PUBLIC ${VIAMCPPSDK_PROTOBUF_LIBRARIES}
   PRIVATE ${VIAMCPPSDK_GRPCXX_REFLECTION_LIBRARIES}
   PUBLIC Boost::headers
-  PRIVATE Boost::log
+  PUBLIC Boost::log
```
benjirewis, line 196: Interestingly, external code (like the example module code) fails to compile unless we make the boost log library `PUBLIC` here.

src/viam/sdk/common/utils.cpp
```diff
@@ -82,5 +85,14 @@ google::protobuf::Duration to_proto(const std::chrono::microseconds& duration) {
     return proto;
 }
 
+void set_logger_severity_from_args(int argc, char** argv) {
+    if (argc >= 3 && strcmp(argv[2], "--log-level=debug") == 0) {
```
stuqdog, line 89: This seems pretty brittle. Specifically, the requirement that `log-level` _must_ be the third arg. Could we instead filter through with [find](https://en.cppreference.com/w/cpp/algorithm/find)?

I could be convinced to leave it as is if you feel strongly about maintaining parity with `Go` on this.

src/viam/sdk/common/utils.hpp
```diff
@@ -47,5 +47,15 @@ std::string bytes_to_string(std::vector<unsigned char> const& b);
 std::chrono::microseconds from_proto(const google::protobuf::Duration& proto);
 google::protobuf::Duration to_proto(const std::chrono::microseconds& duration);
 
+/// @brief Set the boost trivial logger's severity depending on args.
+/// @param argc The number of args.
+/// @param argv The commandline arguments to parse.
+///
+/// Sets the boost trivial logger's severity to debug if "--log-level=debug" is the
+/// the third argument. Sets severity to info otherwise. Useful for module
```
stuqdog, line 55: (opt) if we end up updating this func, we should update comment here to remove the "third arg" restriction.

stuqdog (changes requested):

src/viam/sdk/common/utils.cpp
benjirewis: > Could we instead filter through with [find](https://en.cppreference.com/w/cpp/algorithm/find)?

Certainly could! 

Python and Go do have equally brittle functions. My thought here is that at some point we'll add more exhaustive logging with many levels in the RDK and SDKs. This function will then probably change to parse for `--log-level` in any arg position, and we'll have more complex switch logic to map the string argument to a logging level. I was thinking we would make the function less brittle at that point in time.

stuqdog: Got it. Happy to leave this for a future date if you'd rather make all those improvements at once :)

benjirewis: Sounds good 😃 would prefer to leave this as-is then for now.

cheukt (approved):

PR #110, opened 2023-05-25
RSDK-3312 - fix docs, stop throwing strings

stuqdog:
#### Major changes
Remove `<t><t>` tags from `modules` page in documentation
Stop throwing/catching strings anywhere, consistently throw exceptions
Make catches `const&`
Clarify documentation on `validate` method

### DIFF
--- a/src/viam/sdk/module/service.cpp
+++ b/src/viam/sdk/module/service.cpp
@@ -119,7 +119,7 @@ std::shared_ptr<Resource> ModuleService_::get_parent_resource(Name name) {
     // if the type isn't reconfigurable by default, replace it
     try {
         res->stop();
-    } catch (const std::exception& err) {  // NOLINT
+    } catch (const std::exception& err) {
         BOOST_LOG_TRIVIAL(error) << "unable to stop resource: " << err.what();
     }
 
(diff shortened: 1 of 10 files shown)


src/viam/sdk/registry/registry.hpp
```diff
@@ -95,6 +95,8 @@ class ModelRegistration {
     std::function<std::shared_ptr<Resource>(Dependencies, ResourceConfig)> construct_resource;
 
     /// @brief Validates a resource config.
+    /// @return a list of the resource's implicit dependencies.
```
stuqdog, line 98: @bhaney I opted for only updating the documentation here, since it's both in code and our cpp.viam.dev documentation, and I didn't want to have duplicative comments in two locations within the same file. Let me know if you'd like it above as well!

src/viam/sdk/module/service.cpp
```diff
@@ -111,14 +113,14 @@ ::grpc::Status ModuleService_::ReconfigureResource(
...
-    } catch (std::exception& exc) {
+    } catch (const std::exception& exc) {
     }
 
     // if the type isn't reconfigurable by default, replace it
     try {
         res->stop();
-    } catch (std::string err) {  // NOLINT
-        BOOST_LOG_TRIVIAL(error) << "unable to stop resource: " << err;
+    } catch (const std::exception& err) {  // NOLINT
```
benjirewis, line 122: Is this `NOLINT` still needed?

stuqdog: Good catch! No it is not.

PR #113, opened 2023-06-07
RSDK-3402 - GetGeometries implemented in camera and CPP SDK now building

jckras:


src/viam/sdk/spatialmath/geometry.hpp
```diff
@@ -23,7 +23,48 @@ class GeometryConfig {
...
+    void set_geometry_type(GeometryType type) {
+        geometry_type_ = type;
+    }
+    void set_orientation_config(OrientationConfig config) {
+        orientation_config_ = config;
+    }
+    void set_label(std::string label) {
+        label_ = label;
+    }
+    std::vector<double> get_coordinates(){
```
stuqdog, line 50: returning a vector here is a little brittle. There's nothing programmatically enforcing that it be the right length. I think we should probably instead return a tuple to enforce the number of values, or (if you're up for a slightly larger refactor) make a `coordinates` struct and a `pose` struct, and have the argument for the setter/return for the getter be those structs instead. I leave it up to you :)

src/viam/sdk/tests/mocks/generic_mocks.cpp
stuqdog, line 1: I don't see that we're actually testing the `get_geometries` methods from generic or camera anywhere. We'll want to update the test files (`test_camera.cpp` and `test_generic_component.cpp` to make sure that all this logic works correctly :)

stuqdog (changes requested):
Hey Julie! Thanks for doing this. There are some small changes that are needed here and there and we'll need to update our tests, but overall this looks great!

src/viam/sdk/components/generic/generic.hpp
```diff
@@ -48,6 +48,10 @@ class Generic : public Component {
     /// @return The result of the executed command.
     virtual AttributeMap do_command(AttributeMap command) = 0;
 
+    /// @brief A `GeometryConfig` specifies the format of geometries associated with the component.
```
stuqdog, line 51: (minor) see above comment re: this description in `camera.hpp`

src/viam/sdk/spatialmath/geometry.cpp
```diff
@@ -94,7 +111,62 @@ viam::common::v1::Geometry GeometryConfig::to_proto() const {
             return geometry_;
         }
     }
-};
+}
+void GeometryConfig::set_coordinates(GeometryConfig::coordinates coord) {
```
stuqdog, line 115: Since we've created a struct for coordinates that can be accessed, it seems like a good change to `GeometryConfig` would be to replace `x_`, `y_`, and `z_` with a single `coordinates` struct. Same for the `pose` struct

src/viam/sdk/spatialmath/geometry.cpp
```diff
@@ -94,7 +111,62 @@ viam::common::v1::Geometry GeometryConfig::to_proto() const {
...
+OrientationConfig GeometryConfig::get_orientation_config() const {
+    return orientation_config_;
+}
+
+std::string GeometryConfig::get_label() const {
+    return label_;
+}
+
+bool operator==(const GeometryConfig& lhs, const GeometryConfig& rhs) {
+    if (lhs.geometry_type_ == GeometryType::sphere && lhs.r_ != rhs.r_) {
```
stuqdog, line 163: might be good to add a quick comment explaining the reason for this particular bit of logic.

src/viam/sdk/spatialmath/orientation.cpp
```diff
@@ -8,6 +8,7 @@
 
 #include <viam/api/app/v1/robot.pb.h>
 
+#include "orientation.hpp"
```
stuqdog, line 11: (nit) just for consistency's sake, can we go with angle brace syntax rather than quotation marks?

stuqdog (changes requested):
A couple very small requests, but overall this is looking really good, thanks so much!

src/viam/sdk/spatialmath/geometry.cpp
```diff
@@ -38,8 +38,17 @@ viam::common::v1::Pose GeometryConfig::pose_proto() const {
     return pose;
 }
 
-GeometryConfig GeometryConfig::from_proto(viam::common::v1::Geometry proto) {
+GeometryConfig GeometryConfig::from_proto(const viam::common::v1::Geometry& proto) {
     GeometryConfig cfg;
+    auto& pose = proto.center();
```
stuqdog, line 43: Oh, it looks like we'll need to explicitly declare this as `const` or the clang-tidy will get mad.

cheukt:
removing myself since I'll be out and my C++ is weak

src/viam/sdk/components/camera/camera.hpp
```diff
@@ -131,6 +132,10 @@ class Camera : public Component {
     /// @return The requested `point_cloud`.
     virtual point_cloud get_point_cloud(std::string mime_type) = 0;
 
+    /// @brief A `GeometryConfig` specifies the format of geometries associated with the component.
```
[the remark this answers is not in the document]
jckras: Returns geometries associated with the calling camera or Returns `GeometryConfigs` associated with the calling camera to be tagged to the class?

src/viam/sdk/spatialmath/geometry.cpp
jckras: Would it be okay if the struct was public rather than private?

stuqdog: The values inside the struct can (and idiomatically should) be public. The field within the GeometryConfig class should probably remain private with a getter/setter as necessary.

src/viam/sdk/components/camera/camera.hpp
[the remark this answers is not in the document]
stuqdog: The latter, tagging to the class is good. Thanks for flagging :)

src/viam/sdk/spatialmath/geometry.cpp
jckras: another thing to consider here, once we change it to a struct will I need to update the functions in pose_proto() such as set_x(x_), which I believe are proto functions?

stuqdog: I think we should be able to just do `pose.set_x(coordinates.x)` instead of `pose.set_x(x_)`

jckras: I have made the changes of adding the struct for coordinates and pose and the code looks much neater! Thanks for the suggestion. I also made the other small fixes. Unfortunately I am running into some github problems because my local code is "behind" the main branch. For extra precaution I will wait until Monday to push the final changes!

stuqdog (commented):
Want to make some small changes (discussed offline) but otherwise looks great!

src/viam/sdk/spatialmath/orientation.cpp
```diff
@@ -145,6 +145,5 @@ proto::Orientation OrientationConfig::to_proto() const {
         }
     }
 }
-
```
stuqdog, line 148: (minor) can we add this newline back? Just for consistency with the rest of the SDK :)

stuqdog (changes requested):
A couple very minor additional comments and then I think this is ready to merge!

stuqdog (approved):
LGTM!

PR #114, opened 2023-06-12
RSDK-3335 Add an example tflite modular resource to the C++ SDK

acmorrow:
This PR adds a new tflite mlmodel example modular resource to the C++ SDK. I've validated that I'm able to run the effective detection person detector model with this setup, but not much else. I've attempted to heavily comment the example so that it provides enough guideposts that someone else could probably use it as a template to integrate with another inference engine.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -40,7 +40,7 @@ project(viam-cpp-sdk
   VERSION 0.0.0
   DESCRIPTION "Viam Robotics C++ SDK"
   HOMEPAGE_URL https://github.com/viamrobotics/viam-cpp-sdk
-  LANGUAGES CXX
+  LANGUAGES C CXX
 )
 
 
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -40,7 +40,7 @@ project(viam-cpp-sdk
   VERSION 0.0.0
   DESCRIPTION "Viam Robotics C++ SDK"
   HOMEPAGE_URL https://github.com/viamrobotics/viam-cpp-sdk
-  LANGUAGES C CXX
+  LANGUAGES CXX
 )
 
 
--- a/etc/docker/Dockerfile.debian.bookworm
+++ b/etc/docker/Dockerfile.debian.bookworm
@@ -1,0 +1,44 @@
+FROM debian:bookworm
+
+ENV HOME /root
+ARG DEBIAN_FRONTEND=noninteractive
+
+RUN apt-get update
+
+RUN apt-get -y dist-upgrade
+
+RUN apt-get -y --no-install-recommends install \
+    build-essential \
+    ca-certificates \
+    cmake \
+    curl \
+    doxygen \
+    g++ \
+    gdb \
+    git \
+    gnupg \
+    gpg \
+    less \
+    libabsl-dev \
+    libboost-all-dev \
+    libgrpc++-dev \
+    libprotobuf-dev \
+    libssl-dev \
+    libxtensor-dev \
+    ninja-build \
+    pkg-config \
+    protobuf-compiler-grpc \
+    software-properties-common \
+    sudo \
+    wget
+
+RUN bash -c 'wget -O - https://apt.llvm.org/llvm-snapshot.gpg.key|apt-key add -'
+RUN apt-add-repository -y 'deb http://apt.llvm.org/bookworm/ llvm-toolchain-bookworm-15 main'
+# NOTE: For reasons I cannot explain, the first invocation of apt-add-repository fails
+# but the second works fine. I suspect a Debian bug
+RUN apt-add-repository -y 'deb http://apt.llvm.org/bookworm/ llvm-toolchain-bookworm-15 main'
+RUN apt-get update
+
+RUN apt-get -y --no-install-recommends install -t llvm-toolchain-bookworm-15 \
+    clang-15 \
+    clang-tidy-15
--- a/src/viam/api/CMakeLists.txt
+++ b/src/viam/api/CMakeLists.txt
@@ -138,7 +138,7 @@ if (VIAMCPPSDK_USE_DYNAMIC_PROTOS)
 
     COMMAND ${BUF_COMMAND} generate buf.build/viamrobotics/goutils --template buf.gen.yaml
     COMMAND ${BUF_COMMAND} generate buf.build/googleapis/googleapis --template buf.gen.yaml --path google/rpc --path google/api
-    COMMAND ${BUF_COMMAND} generate "https://github.com/viamrobotics/api#format=git,ref=0f18896259f5a317129370560769f7d2f08b14d7,subdir=proto/viam" --template buf.gen.yaml --path proto/viam/common,proto/viam/component,proto/viam/robot,proto/viam/service,proto/viam/module,proto/viam/tagger,proto/viam/app/v1
+    COMMAND ${BUF_COMMAND} generate buf.build/viamrobotics/api --template buf.gen.yaml --path common,component,robot,service,module,tagger,app/v1
     MAIN_DEPENDENCY buf.gen.yaml
   )
 
--- a/src/viam/api/CMakeLists.txt
+++ b/src/viam/api/CMakeLists.txt
@@ -138,7 +138,7 @@ if (VIAMCPPSDK_USE_DYNAMIC_PROTOS)
 
     COMMAND ${BUF_COMMAND} generate buf.build/viamrobotics/goutils --template buf.gen.yaml
     COMMAND ${BUF_COMMAND} generate buf.build/googleapis/googleapis --template buf.gen.yaml --path google/rpc --path google/api
-    COMMAND ${BUF_COMMAND} generate buf.build/viamrobotics/api --template buf.gen.yaml --path common,component,robot,service,module,tagger,app/v1
+    COMMAND ${BUF_COMMAND} generate "https://github.com/viamrobotics/api#format=git,ref=0f18896259f5a317129370560769f7d2f08b14d7,subdir=proto/viam" --template buf.gen.yaml --path proto/viam/common,proto/viam/component,proto/viam/robot,proto/viam/service,proto/viam/module,proto/viam/tagger,proto/viam/app/v1
     MAIN_DEPENDENCY buf.gen.yaml
   )
 
--- a/src/viam/examples/modules/CMakeLists.txt
+++ b/src/viam/examples/modules/CMakeLists.txt
@@ -43,14 +43,9 @@ install(
   COMPONENT examples
 )
 
-
-include_directories(/opt/homebrew/include)
-get_directory_property(includes INCLUDE_DIRECTORIES)
-set(CMAKE_REQUIRED_INCLUDES ${includes})
-CHECK_INCLUDE_FILE("tensorflow/lite/c/c_api.h" VIAMCPPSDK_HAVE_TFLITE_C_API)
-
-if (VIAMCPPSDK_HAVE_TFLITE_C_API)
-
+find_path(TFLITE_INCLUDE_PATH "tensorflow/lite/c/c_api.h")
+find_library(TFLITE_LIBRARY "tensorflowlite_c")
+if (EXISTS ${TFLITE_INCLUDE_PATH} AND EXISTS ${TFLITE_LIBRARY})
   add_executable(mlmodelservice_tflite
     mlmodelservice_tflite.cpp
   )
@@ -66,5 +61,6 @@ if (VIAMCPPSDK_HAVE_TFLITE_C_API)
     COMPONENT examples
   )
 else()
-  message("XXX ACM")
+  message(WARNING "Failed to find tflite C headers/library: tflite example module will not be built")
+  message(WARNING "If you have tflite in a non-standard location, add -DCMAKE_PREFIX_PATH=/path/to/tflite to your CMake invocation")
 endif()
--- a/src/viam/examples/modules/CMakeLists.txt
+++ b/src/viam/examples/modules/CMakeLists.txt
@@ -12,8 +12,6 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-include(CheckIncludeFile)
-
 add_executable(example_module
   example_module.cpp
 )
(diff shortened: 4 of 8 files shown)


CMakeLists.txt
```diff
@@ -40,7 +40,7 @@ project(viam-cpp-sdk
   VERSION 0.0.0
   DESCRIPTION "Viam Robotics C++ SDK"
   HOMEPAGE_URL https://github.com/viamrobotics/viam-cpp-sdk
-  LANGUAGES CXX
+  LANGUAGES C CXX
```
acmorrow, line 43: I think this can be backed out. I'd needed it at one point while I was experimenting with different ways to make a soft dependency on tflite, but I don't think the mechanism I settled on requires it.

src/viam/api/CMakeLists.txt
```diff
@@ -138,7 +138,7 @@ if (VIAMCPPSDK_USE_DYNAMIC_PROTOS)
 
     COMMAND ${BUF_COMMAND} generate buf.build/viamrobotics/goutils --template buf.gen.yaml
     COMMAND ${BUF_COMMAND} generate buf.build/googleapis/googleapis --template buf.gen.yaml --path google/rpc --path google/api
-    COMMAND ${BUF_COMMAND} generate buf.build/viamrobotics/api --template buf.gen.yaml --path common,component,robot,service,module,tagger,app/v1
+    COMMAND ${BUF_COMMAND} generate "https://github.com/viamrobotics/api#format=git,ref=0f18896259f5a317129370560769f7d2f08b14d7,subdir=proto/viam" --template buf.gen.yaml --path proto/viam/common,proto/viam/component,proto/viam/robot,proto/viam/service,proto/viam/module,proto/viam/tagger,proto/viam/app/v1
```
acmorrow, line 141: Workaround for broken codegen until #113 merges. @zaporter-work had suggested we just merge this if it was going to be a while so that others didn't need to manually apply this in order to be able to work. If #113 merges before this I will remove it.

src/viam/examples/modules/CMakeLists.txt
```diff
@@ -12,6 +12,8 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+include(CheckIncludeFile)
```
acmorrow, line 15: This can also be backed out now that I'm using `find_path` / `find_library` below.

src/viam/examples/modules/CMakeLists.txt
```diff
@@ -40,3 +42,25 @@ install(
   TARGETS example_mlmodel
   COMPONENT examples
 )
+
+find_path(TFLITE_INCLUDE_PATH "tensorflow/lite/c/c_api.h")
```
acmorrow, line 46: In theory we could move this to a new `FindTensorflowLite.cmake` in our modules directory, but since I don't know for how long this module will remain in the C++ SDK, that effort didn't seem justified.

src/viam/sdk/common/proto_type.hpp
```diff
@@ -30,13 +62,13 @@ class ProtoType {
     friend bool operator==(const ProtoType& lhs, const ProtoType& rhs);
 
     template <typename T>
-    T* get() {
-        return boost::get<T>(&proto_type_);
+    auto get() {
```
acmorrow, line 65: Potentially this could be `decltype(auto)` but in practice I don't think it matters here.

src/viam/examples/modules/mlmodelservice_tflite.cpp
```diff
@@ -0,0 +1,759 @@
...
+#include <viam/sdk/rpc/server.hpp>
+#include <viam/sdk/services/mlmodel/mlmodel.hpp>
+#include <viam/sdk/services/mlmodel/server.hpp>
+
+namespace {
+
+namespace vsdk = ::viam::sdk;
+constexpr char service_name[] = "mlmodelservice_tflite";
+
+class MLModelServiceTFLite : public vsdk::MLModelService {
```
acmorrow, line 28: This should have a class level comment that gives an overview and also documents the configuration parameters.

src/viam/examples/modules/mlmodelservice_tflite.cpp
```diff
@@ -0,0 +1,759 @@
...
+    // Block the signals we intend to wait for synchronously.
+    sigset_t sigset;
+    sigemptyset(&sigset);
+    sigaddset(&sigset, SIGINT);
+    sigaddset(&sigset, SIGTERM);
+    pthread_sigmask(SIG_BLOCK, &sigset, NULL);
+
+    // Create a new model registration for the service.
+    auto module_registration = std::make_shared<vsdk::ModelRegistration>(
+        // TODO: Document me.
```
acmorrow, line 691: @stuqdog - I'd like some more details on what this field does / means.

src/viam/api/CMakeLists.txt
acmorrow: I saw! I have already backed this out locally.

src/viam/examples/modules/mlmodelservice_tflite.cpp
```diff
@@ -0,0 +1,759 @@
...
+            // reconfigurations and so other invocations wait on a new
+            // state.
+            std::unique_lock<std::mutex> lock(state_lock_);
+            state_ready_.wait(lock, [this]() { return state_ != nullptr; });
+            swap(state_, state);
+        }
+        try {
+            state = reconfigure_(std::move(dependencies), std::move(configuration));
+        } catch (...) {
+            // If we failed to reconfigure, restore the original state and unblock waiters.
```
tessavitabile, line 70: If we failed to reconfigure, the resource should behave as if it is closed and not allow requests.

src/viam/examples/modules/mlmodelservice_tflite.cpp
```diff
@@ -0,0 +1,759 @@
...
+        // not be higher level protection against invocation during
+        // reconfiguration. Keep all state in a shared_ptr managed
+        // block, and allow client invocations to act against current
+        // state while a new configuration is built, then swap in the
+        // new state. State which is in use by existing invocations
+        // will remain valid until the clients drain. If
+        // reconfiguration fails, fall back to the old state, though
+        // in practice the component is likely to get restarted.
+
+        // Swap out the state_ member with nullptr. Existing infocations
```
tessavitabile, line 55: Trailing comment

tessavitabile (commented):
Looks great! Added a couple comments. No need to wait for my sign-off.

src/viam/examples/modules/mlmodelservice_tflite.cpp
acmorrow: Great. I wasn't sure what was expected. I think I can make this somewhat simpler if I can allow the service to become closed on a failed reconfiguration.

etc/docker/Dockerfile.debian.bookworm
```diff
@@ -0,0 +1,44 @@
+FROM debian:bookworm
```
acmorrow, line 1: This is a drive-by: bookworm has been released so I copied the `sid` `Dockerfile` to `bookworm` and updated it as needed.

acmorrow (commented):
Updated, please take another look. I cleaned up some comments, added some more, and reworked the reconfiguration logic to put the service into a stopped state on a failed reconfiguration.

src/viam/sdk/services/mlmodel/client.cpp
```diff
@@ -48,6 +48,9 @@ std::shared_ptr<MLModelService::named_tensor_views> MLModelServiceClient::infer(
     auto& mutable_input_data = *req->mutable_input_data();
     auto& mutable_input_data_fields = *mutable_input_data.mutable_fields();
 
+    // TODO: Currently, this doesn't validate that we are passing the
```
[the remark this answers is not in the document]
bhaney: I agree that after the flat tensors are added, this will not be necessary- the type check can always be added in later if needed or if flat tensors turn out not to be viable.

bhaney (commented):
So far from what I can tell, this looks good and infer() is doing everything right - I'm going to take one more pass at this tomorrow morning now having digested everything and then should be able to give full approval!

src/viam/examples/modules/mlmodelservice_tflite.cpp
```diff
@@ -0,0 +1,809 @@
...
+            }
+        }
+
+        // Configuration parsing / extraction is complete. Move on to
+        // building the actual model with the provided information.
+
+        // Try to load the provided `model_path`. The TFLite API
+        // declares that if you use `TfLiteModelCreateFromFile` that
+        // the file must remain unaltered during execution. That's not
+        // a guarantee I'm willing to provide, so instead we read the
```
stuqdog, line 364: (nit) s/I'm/we're? Just thinking from the perspective of a third party user expecting an authoritative Viam voice in the example.

src/viam/examples/modules/mlmodelservice_tflite.cpp
stuqdog: Per offline conversation, I think it's an artifact, model/resourceAPI should give all that's necessary.

stuqdog (approved):
I can't speak super well to the accuracy of some of the mlmodel logic, but generally looks good to me! Left a couple comments about comments, no need to re-request review though.

src/viam/examples/modules/mlmodelservice_tflite.cpp
```diff
@@ -0,0 +1,809 @@
...
+                std::ostringstream buffer;
+                buffer << service_name
+                       << ": Value for field `num_threads` is not a positive integer: "
+                       << *num_threads_double;
+                throw std::invalid_argument(buffer.str());
+            }
+
+            state->interpreter_options.reset(TfLiteInterpreterOptionsCreate());
+            TfLiteInterpreterOptionsSetNumThreads(state->interpreter_options.get(),
+                                                  static_cast<int32_t>(*num_threads_double));
```
bhaney, line 421: Should we set a limit to number of threads someone can request? In Go, we warn the user that we limited their threads to maxCPUs/2 if they requested a larger amount.

bhaney (approved):
Just really one comment about a limit of num_threads - otherwise the meat of the methods all look good to me!

src/viam/examples/modules/mlmodelservice_tflite.cpp
```diff
@@ -0,0 +1,759 @@
+#include <pthread.h>
+#include <signal.h>
+
```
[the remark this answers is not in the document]
acmorrow: I guess I think of them as "system" includes rather than std includes, so I grouped them. I'll fold them in with the standard library includes.

src/viam/examples/modules/mlmodelservice_tflite.cpp
```diff
@@ -0,0 +1,759 @@
...
+    ~MLModelServiceTFLite() final {
+        // All invocations arrive via gRPC, so we know we are idle
+        // here. It should be safe to tear down all state
+        // automatically without needing to wait for anything more to
+        // drain.
+    }
+
+    void reconfigure(vsdk::Dependencies dependencies, vsdk::ResourceConfig configuration) final {
+        // Care needs to be taken during reconfiguration. There may
```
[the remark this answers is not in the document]
acmorrow: Fair. At the time I wrote it I wasn't sure it was needed. Changed to `The framework does not offer protection`

src/viam/examples/modules/mlmodelservice_tflite.cpp
```diff
@@ -0,0 +1,759 @@
...
+    struct metadata metadata() final {
+        // Just return a copy of our metadata from leased state.
+        return lease_state_()->metadata;
+    }
+
+   private:
+    struct state;
+
+    std::shared_ptr<state> lease_state_() {
+        // Wait for our state to be valid and then an incrementing
```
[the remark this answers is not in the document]
acmorrow: I forgot a verb there. Changed to:
```
        // Wait for our state to be valid or stopped and then an
        // obtain a shared_ptr to state if valid, incrementing the
        // refcount, or throws if the service is stopped. We don't
        // need to deal with interruption or shutdown because the gRPC
        // layer will drain requests during shutdown, so it shouldn't
        // be possible for callers to get stuck here.
```

src/viam/examples/modules/mlmodelservice_tflite.cpp
```diff
@@ -0,0 +1,759 @@
...
+
+        // We serialize access to the interpreter. We use a
+        // unique_lock because we will move the lock into the shared
+        // state that we return, allowing the higher level to effect a
+        // direct copy out of the tflite buffers while the interpreter
+        // is still locked.
+        std::unique_lock<std::mutex> lock(state->interpreter_mutex);
+
+        // We need at least one input in order to do
+        // inference. Potentially, we need all of them, so this check
```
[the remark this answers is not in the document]
acmorrow: I've made the check exact and updated the comment.

src/viam/examples/modules/mlmodelservice_tflite.cpp
acmorrow: I made the comment stronger.

src/viam/examples/modules/mlmodelservice_tflite.cpp
```diff
@@ -0,0 +1,809 @@
...
+
+        // Walk the inputs, and copy the data from each of the input
+        // tensor views we were given into the associated tflite input
+        // tensor buffer.
+        for (const auto& kv : inputs) {
+            const auto where = state->input_tensor_indices_by_name.find(kv.first);
+            if (where == state->input_tensor_indices_by_name.end()) {
+                std::ostringstream buffer;
+                buffer << service_name << ": No found index for input tensor name `" << kv.first
+                       << "`";
```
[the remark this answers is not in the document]
acmorrow: Sure, done.

acmorrow:
@bhaney and @stuqdog - Thanks for the reviews and approvals! I believe I've addressed all open comments, with one standing question about whether we want a ticket for the `ResourceType` change, which would let me clean up that comment with a `TODO(RSDK-ABCD): `.

src/viam/examples/modules/mlmodelservice_tflite.cpp
bhaney: Sounds good to me! I'm good with leaving things as is

PR #116, opened 2023-06-15
RSDK-3336 Add support for easily building against a local API repo

acmorrow:
- Reformats the options declarations. That's optional (hah!)
- Defines a new option: `VIAMCPPSDK_USE_LOCAL_PROTOS` which causes proto generation to be performed against proto definitions as currently defined in a sibling directory named `api`.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -93,7 +93,7 @@ option(VIAMCPPSDK_USE_DYNAMIC_PROTOS "Regenerate protos as part of building (net
 # - `VIAMCPPSDK_USE_LOCAL_PROTOS`
 #
 # Builds against local proto definitions, as opposed to those in the
-# buf registry. This flag assumes that there is a `buf.workspace` file
+# buf registry. This flag assumes that there is a `buf.work.yaml` file
 # in the parent directory of the C++ SDK source directory, and that
 # the API repo is in a sibling directory to the C++ SDK source, with
 # the directory name `api`. This flag requires
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -45,8 +45,10 @@ project(viam-cpp-sdk
 
 
 # Configure cmake-level options:
+
+# - `BUILD_SHARED_LIBS`
 #
-# - `BUILD_SHARED_LIBS`: Enabled by default so that we produce a
+# Enabled by default so that we produce a
 # modern SDK, this option can be set to `OFF` to build a static
 # library. Note that this may result in non-functional examples.
 #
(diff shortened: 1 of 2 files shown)


src/viam/api/CMakeLists.txt
```diff
@@ -1,5 +1,31 @@
 set(PROTO_GEN_DIR ${CMAKE_CURRENT_SOURCE_DIR})
 
+set(BUF_PROTO_SOURCE buf.build/viamrobotics/api)
+set(BUF_PROTO_COMPONENTS common component robot service module tagger app/v1)
```
acmorrow, line 4: I'm going to verticalize and alphabetize this list.

CMakeLists.txt
```diff
@@ -45,55 +45,85 @@ project(viam-cpp-sdk
...
+#
 option(VIAMCPPSDK_USE_DYNAMIC_PROTOS "Regenerate protos as part of building (network access may be required)" OFF)
 
 
-# # - `VIAMCPPSDK_USE_WALL_WERROR`: This causes the SDK's internal
-# code to compile with `-Wall` and `-Werror` flags.
+# - `VIAMCPPSDK_USE_LOCAL_PROTOS`
+#
+# Builds against local proto definitions, as opposed to those in the
+# buf registry. This flag assumes that there is a `buf.workspace` file
```
stuqdog, line 96: @acmorrow I know you're generally opposed to `.gitignore` proliferation, but I wonder if it's worth adding `buf.work.yaml` to the `.gitignore` in this case. I'd be surprised if we ever want to check one in, but I could certainly imagine that other repos would want to check one in such that adding it to a user's `.gitignore` is a little bit annoying.

stuqdog (approved):
LGTM!

CMakeLists.txt
acmorrow: The `buf.work.yaml` file goes in the parent directory of the SDK, so there isn't anything to git ignore.

src/viam/api/CMakeLists.txt
acmorrow: This is done in latest.

acmorrow:
@stuqdog - Thanks for the quick review. If I get lucky with a green test run I'll squash this in, otherwise could you please take care of it?

CMakeLists.txt
stuqdog: Ah, so it is! Disregard :)

stuqdog:
> @stuqdog - Thanks for the quick review. If I get lucky with a green test run I'll squash this in, otherwise could you please take care of it?

You got it, merging now.

PR #122, opened 2023-06-27
RSDK-4196 Add support for flat tensors in mlmodel

acmorrow:
Updates the C++ SDK to support the changes from https://github.com/viamrobotics/api/pull/279.

### DIFF
--- a/src/viam/sdk/services/mlmodel/client.cpp
+++ b/src/viam/sdk/services/mlmodel/client.cpp
@@ -23,10 +23,7 @@ namespace sdk {
 
 namespace {
 
-struct tensor_storage_and_views {
-    mlmodel_details::tensor_storage storage;
-    MLModelService::named_tensor_views views;
-};
+constexpr bool kUseFlatTensors = false;
 
 }  // namespace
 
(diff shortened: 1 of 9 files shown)


src/viam/sdk/services/mlmodel/client.cpp
```diff
@@ -23,10 +23,7 @@ namespace sdk {
 
 namespace {
 
-struct tensor_storage_and_views {
-    mlmodel_details::tensor_storage storage;
-    MLModelService::named_tensor_views views;
-};
+constexpr bool kUseFlatTensors = true;
```
acmorrow, line 26: I've opted to leave the existing struct based implementation in place so I can easily A/B test. All code selected by `kUseFlatTensors = false` would be removable if we decided to make a breaking change.

src/viam/sdk/services/mlmodel/private/proto.cpp
```diff
@@ -12,6 +12,9 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+#include <memory>
```
acmorrow, line 15: I've moved these down where they belong.

src/viam/sdk/services/mlmodel/private/proto.cpp
```diff
@@ -338,6 +341,145 @@ class tensor_to_pb_value_visitor : public boost::static_visitor<::grpc::Status>
...
+
+template <typename T>
+MLModelService::tensor_views make_sdk_tensor_from_api_tensor_t(const T* data,
+                                                               std::size_t size,
+                                                               std::vector<std::size_t>&& shape,
+                                                               tensor_storage* ts) {
+    if (!data || (size == 0) || shape.empty()) {
+        std::ostringstream message;
+        message << "Empty or zero length data or shape";
+        throw std::invalid_argument(message.str());
```
stuqdog, line 438: (q) just for my own clarity, what is `message` doing here? Naively it looks like we're adding a string and then immediately unpacking that string.

stuqdog (approved):
LGTM!

src/viam/sdk/services/mlmodel/private/proto.cpp
acmorrow: @stuqdog - Merely copy paste from other places where I was adding some state to the string, but here I'm not. I'll remove it.

bhaney (approved):
looked like there wasn't much to change, but you also mentioned how you already wrote a lot with flat-tensors in mind. LGTM!

PR #128, opened 2023-07-18
RSDK-4006 Fixes from working on mlmodel tutorial

acmorrow:
- Fixes three places where I accidentally C++17'd. I didn't catch it because on macOS our brew absl dependency forces us into C++17 mode.
- Adds a missing header for `std::condition_variable` which was needed with libstdc++ but not libc++, so I didn't catch it on macOS builds.
- Fixes a typo in the function-level catch handler for `main` (`excetion` to `exception`).
- Adds a handler for thrown `std::string` since SDK does that.
- Adds an opt-in to build the tflite module (see next)
- Uses FetchContent to build tflite from source to satisfy the tflite dependency for the module. Otherwise, we were trying to get it from system and it just didn't work anywhere. Since this significantly adds to the build time for the SDK, guard it under the flag above.

### DIFF
--- a/src/viam/examples/mlmodel/example_audio_classification_client.cpp
+++ b/src/viam/examples/mlmodel/example_audio_classification_client.cpp
@@ -242,7 +242,7 @@ int main(int argc, char* argv[]) try {
         // first signal is just silence, and the second is just noise.
         const std::vector<float> silence(15600, 0.0);
         const std::vector<float> noise = [&silence]() {
-            std::vector temp{silence};
+            std::vector<float> temp{silence};
             std::random_device seed;
             std::uniform_real_distribution<float> dist{-1.0, 1.0};
             std::generate(begin(temp), end(temp), [&dist, e = std::mt19937(seed())]() mutable {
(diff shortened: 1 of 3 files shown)


src/viam/examples/mlmodel/example_audio_classification_client.cpp
```diff
@@ -387,8 +387,11 @@ int main(int argc, char* argv[]) try {
     std::cout << argv[0] << ": "
               << "Failed: a std::exception was thrown: `" << ex.what() << "``" << std::endl;
     return EXIT_FAILURE;
+} catch (const std::string& ex) {
```
andf-viam, line 390: comment typo: `hardcdoding` on line 360 above.

src/viam/examples/modules/example_mlmodelservice_tflite.cpp
```diff
@@ -126,7 +127,7 @@ class MLModelServiceTFLite : public vsdk::MLModelService {
         // release the lock, and then notify any callers waiting on
         // reconfiguration to complete.
         {
-            std::scoped_lock<std::mutex> lock(state_lock_);
+            std::lock_guard<std::mutex> lock(state_lock_);
```
andf-viam, line 130: comment typo: `constuction` line 586 below

andf-viam (approved):
Thanks for these changes! This covers everything we ran into in testing I think, nice work!

src/viam/examples/mlmodel/example_audio_classification_client.cpp
acmorrow: Done

stuqdog (approved):
LGTM!

PR #129, opened 2023-07-18
Rsdk-4053 patch - create metadata object

bhaney:
Since we are now returning a general "response metadata" message from proto, this will ensure that future additions to the metadata will not be a breaking change.

### DIFF
--- a/src/viam/sdk/components/camera/camera.cpp
+++ b/src/viam/sdk/components/camera/camera.cpp
@@ -93,25 +93,6 @@ std::string Camera::format_to_MIME_string(viam::component::camera::v1::Format fo
     return viam::component::camera::v1::FORMAT_UNSPECIFIED;
 }
 
-std::chrono::time_point<long long, std::chrono::nanoseconds> Camera::timestamp_to_time_pt(
-    const google::protobuf::Timestamp& timestamp) {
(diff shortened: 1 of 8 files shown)


src/viam/sdk/components/camera/camera.cpp
```diff
@@ -145,7 +126,7 @@ Camera::image_collection Camera::from_proto(viam::component::camera::v1::GetImag
         images.push_back(raw_image);
     }
     image_collection.images = std::move(images);
-    image_collection.captured_at = timestamp_to_time_pt(proto.response_metadata().captured_at());
+    image_collection.metadata = ::viam::sdk::from_proto(proto.response_metadata());
```
stuqdog, line 129: we're already in the `viam::sdk` namespace here so no need to use it as a prefix.

stuqdog (changes requested):
A couple small changes, generally looks good! We should coordinate though since I have a PR touching the same code.

src/viam/sdk/components/camera/camera.cpp
bhaney: done

stuqdog (approved):
LGTM!

PR #130, opened 2023-07-19
RSDK-4071 - add higher precision timestamps to capture nanos on mac

stuqdog:


bhaney (approved):
cool! it's funny how protobuf Timestamp goes up to nano but the system_clock doesn't. Thanks for catching this fix, how did you find it? Feels like a subtle thing!

clintpurser (approved):

PR #132, opened 2023-08-02
RSDK-2069 - Fix proto generation

stuqdog:


src/viam/api/CMakeLists.txt
```diff
@@ -87,11 +87,6 @@ if (VIAMCPPSDK_USE_DYNAMIC_PROTOS)
       config/buf.gen.remote.plugin.yaml.in
       buf.gen.yaml
     )
-  elseif ((NOT VIAMCPPSDK_OFFLINE_PROTO_GENERATION) AND (VIAMCPPSDK_GRPCXX_VERSION VERSION_GREATER_EQUAL 1.41.1))
```
jckras, line 90: I remember having to turn these flags on in the command line to avoid problems in github when testing the submitted code and they helped with debugging. Why are we getting rid of them?

jckras (approved):
Added a question for my own understanding, LGTM

acmorrow (approved):
LGTM per earlier discussion.

PR #134, opened 2023-08-03
RSDK-4335 - Default to dynamic proto generation

stuqdog:
#### Major Changes
Flips default behavior of dynamic proto generation and offline proto generation.


stuqdog:
@acmorrow note that I updated defaults for `OFFLINE_PROTO_GENERATION` in addition to `USE_DYNAMIC_PROTOS`. Both flags have been necessary for me to compile, so I erred on the side of simplifying the cmake invocation as much as possible. Happy to chat if you had a preference against default enabling offline proto generation.

acmorrow:
> @acmorrow note that I updated defaults for `OFFLINE_PROTO_GENERATION` in addition to `USE_DYNAMIC_PROTOS`. Both flags have been necessary for me to compile, so I erred on the side of simplifying the cmake invocation as much as possible. Happy to chat if you had a preference against default enabling offline proto generation.

That seems reasonable to me given that without `OFFLINE=ON` you are likely to fail due to the missing protobuf version binding. I somewhat hope that once that is sorted out that we can revert `OFFLINE` to default `OFF`. The problem with offline mode is that it makes you dependent on the local tooling which is of unknown sanity. Can you add a comment on the ticket for fixing the missing version binding that we should evaluate restoring `OFFLINE=OFF` as the default as part of that work?

acmorrow (approved):
LGTM with a request to fold some details about reverting the default change for `OFFLINE` into an existing ticket.

PR #142, opened 2023-08-21
RSDK-4485 - Add extra

stuqdog:
#### Major Changes
- Adds `extra` param where it was missing and ensures consistent usage of `extra`
- Adds missing `GetGeometries`
- Adds `GetProperties` to `Base`

#### Minor Changes
- Add const correctness to `AttributeMap`s
- Some variable name correcting

### DIFF
--- a/src/viam/sdk/services/mlmodel/mlmodel.hpp
+++ b/src/viam/sdk/services/mlmodel/mlmodel.hpp
@@ -109,7 +109,9 @@ class MLModelService : public Service {
     ///
     /// @returns The results of the inference as a shared pointer to `named_tensor_views`. The data
     /// viewed by the views is only valid for the lifetime of the returned shared_ptr.
-    virtual std::shared_ptr<named_tensor_views> infer(const named_tensor_views& inputs) = 0;
+    inline std::shared_ptr<named_tensor_views> infer(const named_tensor_views& inputs) {
+        return infer(inputs, {});
+    }
 
     /// @brief Runs the model against the input tensors and returns inference results as tensors.
     ///
@@ -169,7 +171,9 @@ class MLModelService : public Service {
     };
 
     /// @brief Returns metadata describing the inputs and outputs of the model
-    virtual struct metadata metadata() = 0;
+    inline struct metadata metadata() {
+        return metadata({});
+    }
 
     /// @brief Returns metadata describing the inputs and outputs of the model
     ///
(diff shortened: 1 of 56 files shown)


stuqdog:
@maximpertsov @purplenicole730 this PR is a lot of lines of review (sorry about that!) but it's largely mechanical. I think it'll take a lot less time to review than the line count suggests. If you find that isn't the case, please let me know early and I'll try to fix it!

maximpertsov (approved):

src/viam/sdk/services/mlmodel/mlmodel.hpp
```diff
@@ -111,6 +111,17 @@ class MLModelService : public Service {
...
 
+    /// @brief Runs the model against the input tensors and returns inference results as tensors.
+    ///
+    /// @param `inputs`: The input tensors on which to run inference.
+    ///
+    /// @param `extra`: Any additional arguments to the method.
+    ///
+    /// @returns The results of the inference as a shared pointer to `named_tensor_views`. The data
+    /// viewed by the views is only valid for the lifetime of the returned shared_ptr.
+    virtual std::shared_ptr<named_tensor_views> infer(const named_tensor_views& inputs,
```
acmorrow, line 122: - I think the tflite example and audio client will need an update.

- I don't love that this requires all callers to now provide an extra argument when 99% of the time it will go unused. It is generally not a good idea to mix defaulted parameters with virtual functions because you are combining static and dynamic binding. But you could consider adding a non-virtual overload:
```
inline std::shared_ptr<named_tensor_views> infer(const named_tensor_views& inputs) {
    return infer(inputs, {});
}

virtual std::shared_ptr<named_tensor_views> infer(const named_tensor_views& inputs, const AttributeMap& extra);
```

Or just decide that you aren't that worried about adding the defaults on a virtual and just give `extra` a defaulted `={}`.

Same thoughts, really, for all the API methods that have the new `extra` parameter.

stuqdog: (discussed offline, will de-virtualize the `extra`less method overloads)

stuqdog:
@acmorrow fyi made updates we discussed and added `extra` to methods in the example mlmodel service. will hold off on merging until sometime tomorrow to make space to think on a cleaner solution to our `extra` overload problem.

purplenicole730 (approved):

PR #143, opened 2023-08-28
MovementSensor

zaporter-work:
Howdy!

I needed MovementSensor bindings for a personal project so I created them. 

I don't have time to polish this up and bring it through a PR, but I thought I would share the code in case someone wanted to take it from here. If everyone is busy, that is fine too! Just thought I should share.

There are a few things to figure out (and the linter is complaining about a few missing `const`s), but the general structure is there and the tests pass on my machine...

Weirdest part of this PR is that the file is named `movementsensor/movementsensor.hpp`.

Pros of `movementsensor.hpp`:
- rdk has `movementsensor.go`
- api has `movementsensor.proto`
- protos are generated as `movementsensor`, not `movement_sensor`

Cons of `movementsensor.hpp`:
- `movement_sensor.hpp` would be easier to read and would make more sense
- Python calls the file `movement_sensor.py`
- Ts calls the file `movement-sensor.ts`

### DIFF
--- a/src/viam/sdk/components/movementsensor/movementsensor.hpp
+++ b/src/viam/sdk/components/movementsensor/movementsensor.hpp
@@ -1,0 +1,210 @@
+/// @file components/movementsensor/movementsensor.hpp
+///
+/// @brief Defines a `MovementSensor` component.
+#pragma once
+
+#include <string>
+
+#include <viam/api/component/movementsensor/v1/movementsensor.pb.h>
+
+#include <viam/sdk/common/linear_algebra.hpp>
+#include <viam/sdk/common/proto_type.hpp>
+#include <viam/sdk/common/utils.hpp>
+#include <viam/sdk/config/resource.hpp>
+#include <viam/sdk/registry/registry.hpp>
+#include <viam/sdk/resource/resource_manager.hpp>
+#include <viam/sdk/spatialmath/orientation.hpp>
+
+namespace viam {
+namespace sdk {
+
+/// @defgroup MovementSensor Classes related to the MovementSensor component.
+
+/// @class MovementSensorRegistration
+/// @brief Defines a `ResourceRegistration` for the `MovementSensor` component.
+/// @ingroup MovementSensor
+class MovementSensorRegistration : public ResourceRegistration {
+   public:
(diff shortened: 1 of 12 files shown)


src/viam/sdk/components/movementsensor/movementsensor.hpp
```diff
@@ -0,0 +1,210 @@
...
+        double value;
+    };
+
+    // @struct orientation
+    // @brief MovementSensor is the only user of viam::common::v1::Orientation right now.
+    // Unfortunately, that is not the same as viam::app::v1::Orientation (which is what spatialmath
+    // uses).
+    struct orientation {
+        double o_x, o_y, o_z, theta;
+        // TODO(pre-merge) move the to/from protos in here. Nicer that way
```
zaporter-work, line 58: Maybe? I saw you do this for some other methods? I kept the current way with everything in the component..

Not sure what the plan is

src/viam/sdk/components/movementsensor/server.cpp
```diff
@@ -0,0 +1,304 @@
...
+
+    AttributeMap extra;
+    if (request->has_extra()) {
+        extra = struct_to_map(request->extra());
+    }
+
+    std::unordered_map<std::string, float> result = movementsensor->get_accuracy(extra);
+    response->mutable_accuracy()->empty();
+    for (const auto& i : result) {
+        response->mutable_accuracy()->emplace(i.first, i.second);
```
zaporter-work, line 216: Strangely this like works on my laptop but does not in CI. Different proto versions maybe?
``` 
/tmp/src/viam/sdk/components/movementsensor/server.cpp:216:39: error: no member named 'emplace' in 'google::protobuf::Map<std::basic_string<char>, float>' [clang-diagnostic-error]
        response->mutable_accuracy()->emplace(i.first, i.second);
```

acmorrow: Yes, almost certainly different proto version. I remember encountering this same issue when working on the MLModel stuff and had to stop using `emplace`.

zaporter-work:
Implemented in https://github.com/viamrobotics/viam-cpp-sdk/pull/171

PR #146, opened 2023-09-11
RSDK-4760 consume .a from rust-utils

abe-winter:
## what changed
- download the `.a` instead of the `.so` for libviam_rust_utils
- I did this because the `.so` ends up in a nested directory and therefore adds complexity to module bundling; with the `.a`, there's nothing to bundle
- I think this makes the distributable size for a C++ module smaller
## incidental changes
- added sstream to examples/modules/simple to fix compile error; not sure why this worked before + is failing now
- convenience change to show lint diff in CI when it fails
## optional configuration
- if we want to preserve the old behavior, I can put the file extension behind a config flag
## possible extra testing
- ~:apple: this works on x86 + arm linux per [this module-example-cpp CI run](https://github.com/viamrobotics/module-example-cpp/actions/runs/6150750042), but I didn't try this on a mac~ -> works on mac now

### DIFF
--- a/src/viam/sdk/CMakeLists.txt
+++ b/src/viam/sdk/CMakeLists.txt
@@ -202,6 +202,11 @@ target_link_libraries(viamsdk
 )
 
 
+if (APPLE)
+  target_link_libraries(viamsdk PUBLIC "-framework Security")
+endif()
+
+
 install(
   TARGETS viamsdk
   EXPORT viamsdk
--- a/.github/workflows/linter.yml
+++ b/.github/workflows/linter.yml
@@ -26,5 +26,6 @@ jobs:
           if [ -n "$GEN_DIFF" ]; then
               echo 'linting resulted in changes not in git' 1>&2
               git status
+              git diff
               exit 1
           fi
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -225,9 +225,9 @@ find_package(Threads REQUIRED)
 # TODO: When this is removed, also remove the
 # `target_link_directories` call down in src/CMakeLists.txt, as it
 # will no longer be neede.
-set(viam_rust_utils_file ${CMAKE_CURRENT_BINARY_DIR}/${CMAKE_SHARED_LIBRARY_PREFIX}viam_rust_utils${CMAKE_SHARED_LIBRARY_SUFFIX})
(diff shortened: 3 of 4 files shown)


benjirewis:
I can no longer build/install the C++ SDK after this change; let's sync tomorrow and I can show you the errors I'm seeing. I'm on an M2 Mac.

stuqdog:
> I can no longer build/install the C++ SDK after this change; let's sync tomorrow and I can show you the errors I'm seeing. I'm on an M2 Mac.

I'm on an M1 mac, also unable to build with this change.

abe-winter:
oof I should have tested on mac

I see a bunch of `security_framework::` undefined symbols with `make viamsdk` on mac M1, I'll debug

abe-winter:
I made the static linkage platform-specific: it's now dylib on mac, static everywhere else

This fixes the build issue on mac. In some ways it's kicking the can down the road (like don't we want static rust everywhere eventually). It's not great that bundling a module will be different on mac than linux (you'd need to add the viam-rust `.dylib`, probably in its relative path).

Can dig more into the mac link issue if either of you think this is high-pri; my guess is the fix would end up being one of:
- add static linkage for the security-framework library
- or adding dynamic linkage to the security-framework library in the viamsdk target

benjirewis (approved):
Unclear on the priority of building rust statically on all platforms; seems like it would be pretty much necessary as we move toward a more modular Viam. This seems fine for now, though, if OSX is gonna be a real pain. I'll definitely defer to @stuqdog though. Thanks for looking into this @abe-winter.

abe-winter:
oops turned out the fix was a one-liner to add `-framework` on macos

`make viamsdk` works on macos with static `libviam_rust_utils.a`

benjirewis:
Still LGTM; can build on M2 mac and I see:

`Installing: /Users/benji.rewis/Code/viam-cpp-sdk/build/install/lib/libviam_rust_utils.a`.

src/viam/sdk/CMakeLists.txt
```diff
@@ -202,6 +202,11 @@ target_link_libraries(viamsdk
 )
 
 
+if (APPLE)
```
acmorrow, line 205: @abe-winter - I was out on PTO when this review happened, but this looks a little questionable to me. Can you explain why it was required as part of this change?

PR #161, opened 2023-10-11
RSDK-4679 - Add from_dm_from_extra util function

cheukt:
grabs the value of `fromDataManagement` from the extra dict

### DIFF
--- a/src/viam/sdk/common/utils.cpp
+++ b/src/viam/sdk/common/utils.cpp
@@ -134,13 +134,15 @@ bool from_dm_from_extra(const AttributeMap& extra) {
     if (!extra) {
         return false;
     }
-    if (auto pos = extra->find("fromDataManagement"); pos != extra->end()) {
+    auto pos = extra->find("fromDataManagement");
+    if (pos != extra->end()) {
         ProtoType value = *(pos->second);
         const bool* boolValue = value.get<bool>();
-        return boolValue ? *boolValue : false;
-    } else {
-        return false;
-    };
+        if (boolValue) {
+            return *boolValue;
+        }
+    }
+    return false;
 }
 
 }  // namespace sdk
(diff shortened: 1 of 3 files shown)


dgottlieb (approved):

src/viam/sdk/common/utils.cpp
```diff
@@ -130,5 +130,18 @@ bool operator==(const response_metadata& lhs, const response_metadata& rhs) {
     return lhs.captured_at == rhs.captured_at;
 }
 
+bool from_dm_from_extra(const AttributeMap& extra) {
+    if (!extra) {
+        return false;
+    }
+    if (auto pos = extra->find("fromDataManagement"); pos != extra->end()) {
```
acmorrow, line 137: I think this syntax is C++17 specific but the driver targets C++14.

src/viam/sdk/common/utils.cpp
```diff
@@ -130,5 +130,18 @@ bool operator==(const response_metadata& lhs, const response_metadata& rhs) {
...
+        ProtoType value = *(pos->second);
+        const bool* boolValue = value.get<bool>();
+        return boolValue ? *boolValue : false;
+    } else {
+        return false;
```
acmorrow, line 142: This function has three explicit `return false` statements, but I think it only needs one:

```
if (extra) {
    if (auto pos ...) {
        ...
        if (boolValue) { return *boolValue; }
    }
}
return false;
```

src/viam/sdk/common/utils.cpp
cheukt: thanks for the catch! fixed

src/viam/sdk/common/utils.cpp
cheukt: changed the structure a little, but I do prefer using early returns to reduce nesting and make it easier to read

acmorrow (approved):
LGTM

stuqdog (approved):

PR #164, opened 2023-10-17
RSDK-5194 - set client_ctx auth to a placeholder value

stuqdog:
#### Major changes
Set a placeholder `authority` value for gRPC client contexts.

#### What why? Please explain.
For reasons that aren't entirely clear to me, the `authority` value in the http headers sent to rust from C++ is set to the socket path to the rust runtime, e.g. `/tmp/proxy-XXXXXX.sock`. This is not a valid uri, and is causing rust to send back a `RST_STREAM` message to C++ rather than sending the request on to the server.

In python (and presumably in linux though I had some difficulty testing this), the header value was set to some valid localhost value (`127.0.0.1:50051`). It's not entirely clear to me where or why this is happening, but fortunately the authority value in the headers is not at all relevant for our use case. By simply setting the authority in the client context to some dummy value that is a syntactically valid uri, we no longer send invalid uris to rust and so are able to process calls correctly.

### DIFF
--- a/src/viam/examples/dial/example_dial.cpp
+++ b/src/viam/examples/dial/example_dial.cpp
@@ -63,7 +63,11 @@ int main() {
 
     // ensure we can create clients to the robot
     auto gc = robot->resource_by_name<GenericClient>("generic1");
-    std::cout << "got generic client named " << gc->name() << std::endl;
+    if (gc) {
+        std::cout << "got generic client named " << gc->name() << std::endl;
+    }
+
+    robot->close();
 
     return EXIT_SUCCESS;
 }
--- a/src/viam/sdk/common/utils.cpp
+++ b/src/viam/sdk/common/utils.cpp
@@ -11,6 +11,7 @@
 #include <boost/optional/optional.hpp>
 #include <boost/variant/get.hpp>
 #include <boost/variant/variant.hpp>
+#include <grpcpp/client_context.h>
 
 #include <viam/api/common/v1/common.pb.h>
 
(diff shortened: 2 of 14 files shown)


src/viam/examples/dial/example_dial.cpp
```diff
@@ -63,7 +63,11 @@ int main() {
 
     // ensure we can create clients to the robot
     auto gc = robot->resource_by_name<GenericClient>("generic1");
-    std::cout << "got generic client named " << gc->name() << std::endl;
+    if (gc) {
```
stuqdog, line 66: (flyby) add a guardrail here just in case the robot someone tries to use doesn't have a generic named `generic1`, which causes a segfault (and therefore a failure to close the robot successfully).

src/viam/sdk/common/utils.cpp
```diff
@@ -130,6 +131,14 @@ bool operator==(const response_metadata& lhs, const response_metadata& rhs) {
     return lhs.captured_at == rhs.captured_at;
 }
 
+// the authority is sometimes set to an invalid uri on mac, causing `rust-utils` to fail
+// to process gRPC requests. Fortunately we don't particularly care what the authority is
+// set to within the ctx once a channel is built, so we can just set a placeholder value
+// to appease rust and continue to rely on the gRPC channel to process requests correctly.
+void set_client_ctx_authority(grpc::ClientContext& ctx) {
```
stuqdog, line 138: ideally this would create a ctx and return it so we wouldn't have to still create a ctx at every call site and then pass that ctx into this function, but the gRPC library explicitly disallows use of the copy and assignment constructors for a `ClientContext`, so I think we're stuck with this.

benjirewis (commented):
Nice! Just a nit.

src/viam/sdk/common/utils.cpp
```diff
@@ -130,6 +131,14 @@ bool operator==(const response_metadata& lhs, const response_metadata& rhs) {
     return lhs.captured_at == rhs.captured_at;
 }
 
+// the authority is sometimes set to an invalid uri on mac, causing `rust-utils` to fail
+// to process gRPC requests. Fortunately we don't particularly care what the authority is
+// set to within the ctx once a channel is built, so we can just set a placeholder value
+// to appease rust and continue to rely on the gRPC channel to process requests correctly.
```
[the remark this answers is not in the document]
stuqdog: Yep good call! Done.

benjirewis (approved):
Nice! This allows me to build and successfully run the example dial on my mac and linux machines.

PR #166, opened 2023-10-23
RSDK-5364 RSDK-5195 RSDK-3534

benjirewis:
Updates pin for `api` buf build commit. Cherry-picks commit from #155 to get @hexbabe's changes for a new `timeout` dial option (makes the C++ SDK compatible with the latest version of `rust-utils`). Adds new `obstacle_detector` type and uses it within motion configurations.

### DIFF
--- a/src/viam/api/CMakeLists.txt
+++ b/src/viam/api/CMakeLists.txt
@@ -1,7 +1,6 @@
 set(PROTO_GEN_DIR ${CMAKE_CURRENT_SOURCE_DIR})
 
-# TODO (RSDK-5364): update code to comply with up-to-date proto, then stop pinning to this commit
-set(BUF_PROTO_SOURCE buf.build/viamrobotics/api:2109757288234ffc8248121f60437cdd)
+set(BUF_PROTO_SOURCE buf.build/viamrobotics/api)
 set(BUF_PROTO_COMPONENTS
   app/v1
   common
--- a/src/viam/api/CMakeLists.txt
+++ b/src/viam/api/CMakeLists.txt
@@ -1,6 +1,7 @@
 set(PROTO_GEN_DIR ${CMAKE_CURRENT_SOURCE_DIR})
 
-set(BUF_PROTO_SOURCE buf.build/viamrobotics/api)
+# TODO (RSDK-5441): modify this pin through an automated process
+set(BUF_PROTO_SOURCE buf.build/viamrobotics/api:57108c60d0da8f46df147a496845bdf2160facd6)
 set(BUF_PROTO_COMPONENTS
   app/v1
   common
--- a/src/viam/api/CMakeLists.txt
+++ b/src/viam/api/CMakeLists.txt
@@ -1,7 +1,7 @@
 set(PROTO_GEN_DIR ${CMAKE_CURRENT_SOURCE_DIR})
 
 # TODO (RSDK-5441): modify this pin through an automated process
-set(BUF_PROTO_SOURCE buf.build/viamrobotics/api:57108c60d0da8f46df147a496845bdf2160facd6)
+set(BUF_PROTO_SOURCE buf.build/viamrobotics/api:v0.1.216)
 set(BUF_PROTO_COMPONENTS
   app/v1
   common
--- a/src/viam/sdk/rpc/dial.cpp
+++ b/src/viam/sdk/rpc/dial.cpp
@@ -20,7 +20,7 @@ extern "C" char* dial(const char* uri,
                       const char* type,
                       const char* payload,
                       bool allow_insecure,
-                      const float timeout,
+                      float timeout,
                       void* ptr);
 namespace viam {
 namespace sdk {
@@ -57,7 +57,7 @@ void DialOptions::set_entity(boost::optional<std::string> entity) {
     auth_entity_ = entity;
 }
 
-void DialOptions::set_timeout(const float timeout) {
+void DialOptions::set_timeout(float timeout) {
     timeout_ = timeout;
 }
 
@@ -69,7 +69,7 @@ const boost::optional<Credentials>& DialOptions::credentials() const {
     return credentials_;
 }
 
-const float DialOptions::timeout() const {
+float DialOptions::timeout() const {
     return timeout_;
 }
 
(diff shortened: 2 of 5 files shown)


src/viam/api/CMakeLists.txt
```diff
@@ -1,7 +1,7 @@
 set(PROTO_GEN_DIR ${CMAKE_CURRENT_SOURCE_DIR})
 
-# TODO (RSDK-5364): update code to comply with up-to-date proto, then stop pinning to this commit
-set(BUF_PROTO_SOURCE buf.build/viamrobotics/api:2109757288234ffc8248121f60437cdd)
+# TODO (RSDK-5441): modify this pin through an automated process
```
benjirewis, line 3: New `TODO` of [RSDK-5441](https://viam.atlassian.net/browse/RSDK-5441) here. This PR just updates our API to include https://github.com/viamrobotics/api/commit/57108c60d0da8f46df147a496845bdf2160facd6, which means the `viam::robot::v1::Status` class will now contain a `last_reconfigured` field. I investigated implementing [RSDK-4574](https://viam.atlassian.net/browse/RSDK-4573) (returning native C++ types from the robot API). Adding a native `status` type brought up a lot of questions (is `last_reconfigured` a `std::chrono::time_point`? is the inner `status` field an `AttributeMap`? how must we extend the mock robot service?) that I think I'll punt on until the C++ usability epic.

cheukt (approved):

src/viam/sdk/rpc/dial.hpp
```diff
@@ -41,15 +41,20 @@ class Credentials {
 class DialOptions {
    public:
     DialOptions()
-        : auth_entity_(boost::none), credentials_(boost::none), allow_insecure_downgrade_(false) {}
+        : auth_entity_(boost::none),
```
acmorrow, line 44: This constructor could be improved. The initializations of `auth_entity_` and `credentials_` with `boost::none` are unnecessary, since that is the state of a default initialized `boost::optional`. And the initializers of `allow_insecure_downgrade_` and `timeout_` would be better done as non-static data member initializers below, which would have the added benefit of moving the text explaining the choice of `20.0` next to its application:

```
...
bool allow_insecure_downgrade_ = false;
...
float timeout_ = 20.0f;
```

That'd make it possible to define the constructor as `DialOptions::DialOptions() = default` in `dial.cpp` which would also de-inline the constructor, which is probably a good idea.

src/viam/sdk/rpc/dial.hpp
```diff
@@ -41,15 +41,20 @@ class Credentials {
...
 
     const boost::optional<Credentials>& credentials() const;
     const boost::optional<std::string>& entity() const;
     bool allows_insecure_downgrade() const;
+    float timeout() const;
 
     void set_entity(boost::optional<std::string> entity);
     void set_credentials(boost::optional<Credentials> creds);
     void set_allow_insecure_downgrade(bool allow);
+    void set_timeout(float timeout);
```
acmorrow, line 57: Consider using something from `std::chrono` to represent timeouts.

src/viam/api/CMakeLists.txt
```diff
@@ -1,7 +1,7 @@
...
+set(BUF_PROTO_SOURCE buf.build/viamrobotics/api:57108c60d0da8f46df147a496845bdf2160facd6)
```
acmorrow, line 4: Is there a tag this could bind to instead of a sha?

acmorrow (commented):
I have some minor thoughts on this, feel free to kick any of these comments / ideas forward to the useability work.

stuqdog (approved):
lgtm modulo Drew's comments!

src/viam/sdk/rpc/dial.cpp
```diff
@@ -56,6 +59,10 @@ void DialOptions::set_entity(boost::optional<std::string> entity) {
     auth_entity_ = entity;
 }
 
+void DialOptions::set_timeout(std::chrono::seconds timeout) {
```
acmorrow, line 62: Just as a good practice, either 1) take the parameter as `const&`, or 2) continue to take it as a value but then `std::move` it into `timeout_`. Yes, it doesn't matter for this particular type and use-case much, but it is just a good habit to get into.

acmorrow, line 62: Seconds are fairly coarse, and since they are denominated in integers you couldn't, as this is written, express a 2.5 second timeout. I'd consider either:
1) Going to `std::milliseconds` OR
2) Taking `std::chrono::duration<float>` which implicitly represents seconds but can take a non-integer value.

src/viam/sdk/rpc/dial.cpp
```diff
@@ -64,6 +71,10 @@ const boost::optional<Credentials>& DialOptions::credentials() const {
     return credentials_;
 }
 
+std::chrono::seconds DialOptions::timeout() const {
```
acmorrow, line 74: This could return by `const &`

src/viam/sdk/rpc/dial.hpp
```diff
@@ -63,7 +64,11 @@ class DialOptions {
 
     /// @brief Allows the RPC connection to be downgraded to an insecure connection if detected.
     /// This is only used when credentials are not present.
-    bool allow_insecure_downgrade_;
+    bool allow_insecure_downgrade_ = false;
+
+    /// @brief Duration before the dial connection times out
+    /// Set to 20sec to match _defaultOfferDeadline in goutils/rpc/wrtc_call_queue.go
+    std::chrono::seconds timeout_ = std::chrono::seconds(20);
```
acmorrow, line 71: I think you can even just write this like `std::chrono::seconds timeout_{20};`

acmorrow (approved):
LGTM, all things mentioned here are entirely optional.

src/viam/sdk/rpc/dial.hpp
benjirewis: Yep!

src/viam/sdk/rpc/dial.cpp
benjirewis: Done.

src/viam/sdk/rpc/dial.cpp
benjirewis: Fair point; I chose option 2. At the rust level (and I believe also now in Python), the timeout value is a float representing seconds. I think we want to allow non-integer values, but maintain the unit across the SDKs where possible.

benjirewis: Ah good point. Chose option 2, as I've seen that `std::move` pattern quite a bit throughout the SDK.

benjirewis (commented):
Appreciate all the C++ knowledge, @acmorrow, and do not hesitate to suggest what may seem obvious to you in my future C++ PRs! Definitely still a beginner in the language, and the suggestions are usually non-obvious to me and very informative.

acmorrow (approved):
LGTM

PR #167, opened 2023-10-25
Fix complex module client

benjirewis:
Prints resources more cleanly in complex client, example dial and API key example dial. Uses const references in for loops for all three (my linter was complaining, and using const refs where possible in for-each loops seems idiomatic). Removes erroneous `namespace` field from `mybase` complex module example JSON (namespace should be `rdk` which is the default). Registers gizmo and summation APIs in complex module client. Adds missing newlines to complex module client ouput. Output now looks like:

```
Resources
	name: base1 (type:component subtype:base)
	name: motor1 (type:component subtype:motor)
	name: gizmo1 (type:component subtype:gizmo)
	name: mysum1 (type:service subtype:summation)
	name: builtin (type:service subtype:data_manager)
	name: builtin (type:service subtype:motion)
	name: builtin (type:service subtype:sensors)
	name: motor2 (type:component subtype:motor)
gizmo1 do_one returned: 1
gizmo1 do_one_client_stream returned: 1
gizmo1 do_two returned: arg1=false
gizmo1 do_one_server_stream returned: 
	1
	0
	1
	0
gizmo1 do_one_bidi_stream returned: 
	0
	0
summation1 sum of numbers [0, 10) is: 45
base1 is not moving
```

### DIFF
--- a/src/viam/examples/modules/complex/client.cpp
+++ b/src/viam/examples/modules/complex/client.cpp
@@ -78,11 +78,11 @@ int main() {
     // Exercise Summation methods.
     auto sc = robot->resource_by_name<SummationClient>("mysum1");
     if (!sc) {
-        std::cerr << "could not get 'summation1' resource from robot" << std::endl;
+        std::cerr << "could not get 'mysum1' resource from robot" << std::endl;
         return EXIT_FAILURE;
     }
     double sum = sc->sum({0, 1, 2, 3, 4, 5, 6, 7, 8, 9});
-    std::cout << "summation1 sum of numbers [0, 10) is: " << sum << std::endl;
+    std::cout << "mysum1 sum of numbers [0, 10) is: " << sum << std::endl;
 
     // Exercise a Base method.
     auto bc = robot->resource_by_name<BaseClient>("base1");
--- a/src/viam/examples/modules/complex/client.cpp
+++ b/src/viam/examples/modules/complex/client.cpp
@@ -47,7 +47,7 @@ int main() {
     // Print resources.
     std::cout << "Resources" << std::endl;
     std::vector<ResourceName>* resource_names = robot->resource_names();
-    for (ResourceName& resource : *resource_names) {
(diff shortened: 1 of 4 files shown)


src/viam/examples/modules/complex/client.cpp
```diff
@@ -38,14 +37,19 @@ int main() {
     std::string address(uri);
     Options options(1, opts);
 
+    // Register custom gizmo and summation API so robot client can access resources
+    // of that type from the server.
+    Registry::register_resource(Gizmo::static_api(), Gizmo::resource_registration());
+    Registry::register_resource(Summation::static_api(), Summation::resource_registration());
```
benjirewis, line 43: Users have to manually register custom APIs in their client _as well_ as their servers (not registering here means that the client resource manager cannot emplace a gizmo or summation client). This is normally not necessary as resource registration for regular APIs happens at runtime. I've written a note in https://viam.atlassian.net/browse/RSDK-5206 to investigate alternatives to this strategy.

src/viam/examples/modules/complex/client.cpp
```diff
@@ -55,24 +59,24 @@ int main() {
...
+    std::cout << "gizmo1 do_one_bidi_stream returned: " << std::endl;
     for (bool ret : do_one_bidi_stream_ret) {
         std::cout << '\t' << ret << std::endl;
     }
 
     // Exercise Summation methods.
-    auto sc = robot->resource_by_name<SummationClient>("summation1");
+    auto sc = robot->resource_by_name<SummationClient>("mysum1");
     if (!sc) {
         std::cerr << "could not get 'summation1' resource from robot" << std::endl;
```
stuqdog, line 81: `s/summation1/mysum1`?

stuqdog (approved):
Nice lgtm!

src/viam/examples/modules/complex/client.cpp
benjirewis: 😮‍💨 oof thank you!

PR #171, opened 2023-11-01
RSDK-4713 RSDK-5626 MovementSensor

abe-winter:
## What changed
- implement `MovementSensor` component (mostly copying from [Zack's branch](https://github.com/viamrobotics/viam-cpp-sdk/pull/143))
- incidental: fix issue in handler map where only 1st model implementing an API is registered
- add (privately for now) `ResourceRegistration2` a one-liner to implement a resource registration -- this is in movementsensor.cpp for now but could be dropped into registry.hpp if useful in the future


src/viam/sdk/registry/registry.hpp
```diff
@@ -58,6 +58,36 @@ class ResourceRegistration {
...
+
+    static std::shared_ptr<ResourceRegistration> resource_registration() {
+        const google::protobuf::DescriptorPool* p =
+            google::protobuf::DescriptorPool::generated_pool();
+        const google::protobuf::ServiceDescriptor* sd =
+            p->FindServiceByName(MyProtoService::service_full_name());
+        if (!sd) {
+            throw std::runtime_error("Unable to get service descriptor");
+        }
+        return std::make_shared<MyResourceReg>(sd);
```
acmorrow, line 87: I wonder if you could get rid of the CRTP if this wasn't a static member of the template, but a free function. Something like:

```
template <typename ClientType, typename ServerType, typename ProtoServiceType>
std::shared_ptr<ResourceRegistration> make_resource_registration() {
    // ...
    return std::make_shared<ResourceRegistration2<ClientType, ServerType, ProtoServiceType>>(sd);
}
```

Or perhaps it could be a member function of `Registry`:

```
class Registry {
public:
...
template<typename T, typename TClient, typename TServer>
void register_resource() { ... };
```

And then those init blocks would look like:

```
namespace {
bool init() {
    Registry::register_resource<MovementSensor, MovementSensorClient, MovementSensorServer>();
    return true;
};

// NOLINTNEXTLINE(cert-err58-cpp)
const bool inited = init();
}  // namespace

File: src/viam/sdk/registry/registry.hpp [hunk 2]
```diff
@@ -58,6 +58,36 @@ class ResourceRegistration {
     const google::protobuf::ServiceDescriptor* service_descriptor_;
 };
 
+/// @brief subclass of ResourceRegistration with templatized defaults
+template <typename MyResourceClient,
+          typename MyResourceServer,
+          typename MyProtoService,
+          typename MyResourceReg>
+class ResourceRegistration2 : public ResourceRegistration {
```
acmorrow, line 66: Suggest making this class `final`.

src/viam/sdk/registry/registry.hpp
```diff
@@ -58,6 +58,36 @@ class ResourceRegistration {
...
+   public:
+    using ResourceRegistration::ResourceRegistration;
+    std::shared_ptr<ResourceServer> create_resource_server(
```
acmorrow, line 69: This function should be `override`.

src/viam/sdk/registry/registry.hpp
```diff
@@ -58,6 +58,36 @@ class ResourceRegistration {
...
+        std::shared_ptr<ResourceManager> manager) {
+        return std::make_shared<MyResourceServer>(manager);
+    }
+
+    std::shared_ptr<Resource> create_rpc_client(std::string name,
```
acmorrow, line 74: This function should be `override`.

src/viam/sdk/components/movementsensor/client.hpp
```diff
@@ -0,0 +1,80 @@
...
+        grpc::ClientContext ctx;
+        *request.mutable_name() = self->name();
+        *request.mutable_extra() = map_to_struct(extra);
+        throw_status((*(self->stub_).*method)(&ctx, request, &resp));
+    }
+    std::unique_ptr<Stub> stub_;
+    std::shared_ptr<grpc::Channel> channel_;
+};
+
+class MovementSensorRegistration
```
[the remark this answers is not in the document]
acmorrow: I also think this belongs in the `movementsensor.hpp` (or, better, `movementsensor.cpp` if you can ).

src/viam/sdk/registry/registry.hpp
```diff
@@ -58,6 +58,36 @@ class ResourceRegistration {
     const google::protobuf::ServiceDescriptor* service_descriptor_;
 };
 
+/// @brief subclass of ResourceRegistration with templatized defaults
+template <typename MyResourceClient,
+          typename MyResourceServer,
+          typename MyProtoService,
```
acmorrow, line 64: The `MyProtoService` type parameter could probably go away if all `ResourceServer` types had a required `proto_service_type` typedef. They all inherit from that type so it would make sense:

```
class MLModelServiceServer : public ResourceServer,
                             public ::viam::service::mlmodel::v1::MLModelService::Service {
   public:
       using proto_service_type = ::viam::service::mlmodel::v1::MLModelService::Service;
   ...
};
```

Then you could just say `p->FindServiceByName(MyResourceServer::proto_service_type::service_full_name());`, and then this template would have only two types: the Client and the Server (if you get rid of the CRTP).

Of course, that would require attaching that typedef to all server classes.

src/viam/sdk/components/movementsensor/movementsensor.cpp
```diff
@@ -0,0 +1,123 @@
...
+#include <viam/sdk/common/utils.hpp>
+#include <viam/sdk/components/movementsensor/client.hpp>
+#include <viam/sdk/components/movementsensor/server.hpp>
+#include <viam/sdk/registry/registry.hpp>
+#include <viam/sdk/resource/resource.hpp>
+
+namespace viam {
+namespace sdk {
+
+std::shared_ptr<ResourceRegistration> MovementSensor::resource_registration() {
```
acmorrow, line 17: With the new templated type, does this function still have value? Is it intrinsically useful to be able to obtain the `ResourceRegistration` for every type? Or could this be inlined into the `init` call in each `.cpp`:

```
namespace {
bool init() {
    Registry::register_resource(MovementSensor::static_api(),
                                MovementSensorRegistration::resource_registration());
    return true;
};

// NOLINTNEXTLINE(cert-err58-cpp)
const bool inited = init();
}  // namespace
```

Or, if you take my suggestion to eliminate the CRTP:

```
namespace {
bool init() {
    Registry::register_resource(
        MovementSensor::static_api(),
        make_resource_registration<MovementSensorClient, MovementSensorServer>());
    return true;
};

// NOLINTNEXTLINE(cert-err58-cpp)
const bool inited = init();
}  // namespace
```

acmorrow (commented):
I love a good boilerplate elimination, and I have lots of thoughts. Feel free to take or leave what you find useful.

One thing I would suggest, is to split this into two reviews. One to just add the movement sensor under the standard registration mechanics, and then a follow-up to refactor registration if that ends up being considered worthwhile.

acmorrow:
Also, @stuqdog, I know there are some plans afoot to eliminate the static initialization. There have also been some thoughts about making a cleaner separation between the client facing types and the server (e.g. module supporting) types. The current registration model requires that one part of the code (the registration type) know of both the Server and Client types, because there must be one `ResourceRegistration` subtype that can produce both types. That might be another thing to consider refactoring if/when the auto-registration is addressed.

src/viam/sdk/registry/registry.hpp
abe-winter: done in 61408bfe

src/viam/sdk/registry/registry.hpp
abe-winter: done in 61408bfe

src/viam/sdk/registry/registry.hpp
```diff
@@ -58,6 +58,36 @@ class ResourceRegistration {
     const google::protobuf::ServiceDescriptor* service_descriptor_;
 };
 
+/// @brief subclass of ResourceRegistration with templatized defaults
+template <typename MyResourceClient,
```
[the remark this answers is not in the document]
abe-winter: done in c5957649

src/viam/sdk/components/movementsensor/client.hpp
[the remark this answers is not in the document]
abe-winter: originally had it in client.hpp to avoid circular import

but you're right + it doesn't need to be in a header -- moved to movementsensor.cpp in a1c30361, tested with my sample module; works

src/viam/sdk/components/movementsensor/movementsensor.cpp
abe-winter: removed wrapper in 58685273

src/viam/sdk/components/movement_sensor/movement_sensor.hpp
```diff
@@ -0,0 +1,164 @@
...
+    /// @struct compassheading
+    struct compassheading {
+        /// A number from 0-359 where 0 is North, 90 is East, 180 is South, and 270 is   West
+        double value;
+    };
+
+    // @struct orientation
+    // @brief MovementSensor is the only user of viam::common::v1::Orientation right now.
+    // Unfortunately, that is not the same as viam::app::v1::Orientation (which is what spatialmath
+    // uses).
```
zaporter-work, line 40: Polish to be the actual comment

src/viam/sdk/components/movement_sensor/movement_sensor.hpp
```diff
@@ -0,0 +1,164 @@
...
+///
+/// This acts as an abstract base class to be inherited from by any drivers representing
+/// specific movementsensor implementations. This class cannot be used on its own.
+class MovementSensor : public Component {
+   public:
+    /// @struct compassheading
+    struct compassheading {
+        /// A number from 0-359 where 0 is North, 90 is East, 180 is South, and 270 is   West
+        double value;
+    };
```
zaporter-work, line 35: Not hyped that this lives on the movement sensor object, but probably fine

src/viam/sdk/components/movement_sensor/client.cpp
```diff
@@ -0,0 +1,121 @@
...
+    : MovementSensor(std::move(name)),
+      stub_(viam::component::movementsensor::v1::MovementSensorService::NewStub(channel)),
+      channel_(std::move(channel)){};
+
+/// helper to throw on !ok grpc status
+void throw_status(const ::grpc::Status& status) {
+    if (!status.ok()) {
+        throw std::runtime_error(status.error_message());
+    }
+}
```
zaporter-work, line 31: :+1:

zaporter-work (approved):
Thanks!

src/viam/sdk/components/movement_sensor/movement_sensor.hpp
abe-winter: fixed in cd45668

src/viam/sdk/components/movement_sensor/movement_sensor.hpp
abe-winter: want me to change to typedef?
```suggestion
    /// A number from 0-359 where 0 is North, 90 is East, 180 is South, and 270 is West
    typedef double compassheading;
```

src/viam/sdk/components/movement_sensor/client.cpp
```diff
@@ -0,0 +1,121 @@
...
+namespace viam {
+namespace sdk {
+
+MovementSensorClient::MovementSensorClient(std::string name, std::shared_ptr<grpc::Channel> channel)
+    : MovementSensor(std::move(name)),
+      stub_(viam::component::movementsensor::v1::MovementSensorService::NewStub(channel)),
+      channel_(std::move(channel)){};
+
+/// helper to throw on !ok grpc status
+void throw_status(const ::grpc::Status& status) {
```
acmorrow, line 27: Just a note, but one problem with functions like this is that it means that the throw location is always this function, and not the place where the error happened. That's why these are often implemented via macros.

acmorrow:
I only really looked at the resource registration changes, so my LGTM is limited to that. Overall, I think resource registration needs to be refactored and I expect that the final result will look more like a template member on `Registry` which does the work that `ResourceRegistration2` does now.

stuqdog (approved):
Nice! A couple small comments/questions but otherwise looks great, thanks for doing this. No need to treat my re-review as blocking.

benjirewis:
Potentially an unhelpful too-many-cooks situation if I review here, and my review is mostly qs 😆. I'll eventually follow up offline about those qs instead.

src/viam/sdk/components/movement_sensor/client.cpp
```diff
@@ -0,0 +1,121 @@
...
+
+    const google::protobuf::Struct proto_command = map_to_struct(command);
+    *request.mutable_command() = proto_command;
+    *request.mutable_name() = this->name();
+
+    throw_status(stub_->DoCommand(&ctx, request, &response));
+    return struct_to_map(response.result());
+}
+
+// TODO this one didn't have a throw. is that intentional?
```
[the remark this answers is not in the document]
abe-winter: oops I just meant the old PR I copied doesn't have a throw, but it looks like there are 6 other client.cpp with get_geometries and they all don't throw here

src/viam/sdk/components/movement_sensor/client.cpp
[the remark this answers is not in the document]
abe-winter: moved to utils in 3d47fa0, made a macro per andrew's comment about location, renamed THROW_NOT_OK

abe-winter: made macro in 3d47fa0

src/viam/sdk/components/movement_sensor/movement_sensor.cpp
```diff
@@ -0,0 +1,159 @@
...
+namespace viam {
+namespace sdk {
+
+// todo: move me somewhere central
+/// @brief subclass of ResourceRegistration with templatized defaults
+template <typename ResourceClientT,
+          typename ResourceServerT,
+          typename ProtoServiceT,
+          typename ResourceRegT>
+class ResourceRegistration2 : public ResourceRegistration {
```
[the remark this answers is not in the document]
abe-winter: I"m assuming this comment was for the MovementSensorRegistration class? done in 7dfe377, made final + moved to the anonymous namespace w/ init

src/viam/sdk/common/utils.hpp
```diff
@@ -89,5 +89,8 @@ void set_logger_severity_from_args(int argc, char** argv);
 /// @param extra The extra AttributeMap.
 bool from_dm_from_extra(const AttributeMap& extra);
 
+/// helper to throw on !ok grpc::Status
+#define THROW_NOT_OK(status) if (!status.ok()) { throw std::runtime_error(status.error_message()); }
```
acmorrow, line 93: Somehow my comment didn't make it onto the review:

> Let's not put this in a common location just yet. We have no facilities to ensure macro hygiene right now. I'd rather address this as part of the (I believe upcoming) work to refactor error handling and exceptions throughout the codebase. Fine to push it down into the file that uses it for now.

src/viam/sdk/components/movement_sensor/client.hpp
```diff
@@ -0,0 +1,71 @@
...
+    using MovementSensor::get_linear_acceleration;
+    using MovementSensor::get_linear_velocity;
+    using MovementSensor::get_orientation;
+    using MovementSensor::get_position;
+    using MovementSensor::get_properties;
+
+    typedef viam::component::movementsensor::v1::MovementSensorService::StubInterface Stub;
+
+   private:
+    /// template to wrap a stub. todo: move to base class
```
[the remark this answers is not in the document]
abe-winter: ah I didn't realize how much variation there is in the setup across different components. removed todo

src/viam/sdk/components/movement_sensor/movement_sensor.cpp
```diff
@@ -0,0 +1,159 @@
...
+#include <viam/sdk/components/movement_sensor/client.hpp>
+#include <viam/sdk/components/movement_sensor/server.hpp>
+#include <viam/sdk/registry/registry.hpp>
+#include <viam/sdk/resource/resource.hpp>
+
+namespace viam {
+namespace sdk {
+
+// todo: move me somewhere central
+/// @brief subclass of ResourceRegistration with templatized defaults
```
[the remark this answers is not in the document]
abe-winter: removed docstring in 7dfe3774

src/viam/sdk/components/movement_sensor/server.hpp
```diff
@@ -0,0 +1,111 @@
...
+        ::grpc::Status status;
+        AttributeMap extra;
+        std::shared_ptr<Cls> movementsensor;
+
+        bool ok() const {
+            return status.ok();
+        }
+    };
+
+    // todo: move to base
```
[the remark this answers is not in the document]
abe-winter: my overreach, removed todo

src/viam/sdk/components/movement_sensor/server.cpp
```diff
@@ -0,0 +1,197 @@
...
+
+    return ::grpc::Status();
+}
+
+void MovementSensorServer::register_server(std::shared_ptr<Server> server) {
+    server->register_service(this);
+}
+
+}  // namespace sdk
+}  // namespace viam
```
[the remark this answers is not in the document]
abe-winter: fixed in 0830a3d

src/viam/sdk/components/movement_sensor/server.hpp
```diff
@@ -0,0 +1,111 @@
...
+    /// wrapper template for boilerplate
```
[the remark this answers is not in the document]
abe-winter: fixed 0830a3d

src/viam/sdk/common/utils.hpp
abe-winter: removed wrapper for now 8673dc7

src/viam/sdk/components/movement_sensor/movement_sensor.hpp
zaporter-work: I don't have a  strong opinion. typedef is probably better but either is fine

PR #172, opened 2023-11-01
RSDK-5441 - Pin buf versions through an automated process

cheukt:
uses a buf.lock file to do the pinning
for some reason updating the api causes tests to fail, so did not do it this pr.
something about https://github.com/viamrobotics/api/commit/8fc476f421f79d8ea1a8efdffdb670c3559e8b6b doesn't work, even though we have no sensors

an example pr is https://github.com/viamrobotics/viam-cpp-sdk/pull/170


.github/workflows/update_protos.yml
```diff
@@ -31,18 +31,19 @@ jobs:
           cmakeVersion: 3.25.2
           ninjaVersion: 1.11.1
       - uses: bufbuild/buf-setup-action@v1.9.0
-        with:
-          github_token: ${{ secrets.GITHUB_TOKEN }}
```
benjirewis, line 35: Was `github_token`, the below `repo-token`, and the below `BUF_TOKEN` already unnecessary? Or made unnecessary by what you're doing here?

src/viam/api/CMakeLists.txt
```diff
@@ -1,7 +1,23 @@
...
-set(BUF_PROTO_SOURCE buf.build/viamrobotics/api:v0.1.216)
+
+if (VIAMCPPSDK_USE_DYNAMIC_PROTOS)
+  # reads in buf.lock and grabs the commit hashes supported by this version of C++ SDK
+  if (NOT EXISTS ${CMAKE_CURRENT_SOURCE_DIR}/buf.lock)
+    message(FATAL_ERROR "No `buf.lock` file was found in the src/viam/api directory")
+  endif()
+  file(READ ${CMAKE_CURRENT_SOURCE_DIR}/buf.lock lockfile)
+
+  string(REGEX MATCH "googleapis[ \t\r\n]*commit: ([A-Za-z0-9_]*)" _ ${lockfile})
```
benjirewis, line 11: Awesome; I hadn't seen builtin cmake regex matching before.

.github/workflows/update_protos.yml
cheukt: I believe they've been unnecessary for a bit, we just haven't removed it

benjirewis (approved):
LGTM thanks! 🧑‍🔧

stuqdog (approved):
awesome lgtm, thanks!

PR #174, opened 2023-11-06
RSDK-794 Sensor wrapper

benjirewis:
[RSDK-794](https://viam.atlassian.net/browse/RSDK-794)

Implements wrappers for sensor API.

(cc @cheukt and @acmorrow).

[RSDK-794]: https://viam.atlassian.net/browse/RSDK-794?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/viam/sdk/components/movement_sensor/movement_sensor.hpp
+++ b/src/viam/sdk/components/movement_sensor/movement_sensor.hpp
@@ -21,7 +21,7 @@ namespace sdk {
 
 /// @defgroup MovementSensor Classes related to the MovementSensor component.
 
-/// @class MovementSensor movementsensor.hpp "components/movement_sensor/movement_sensor.hpp"
+/// @class MovementSensor movement_sensor.hpp "components/movement_sensor/movement_sensor.hpp"
 /// @ingroup MovementSensor
 ///
 /// This acts as an abstract base class to be inherited from by any drivers representing
--- a/src/viam/sdk/components/movement_sensor/movement_sensor.hpp
+++ b/src/viam/sdk/components/movement_sensor/movement_sensor.hpp
@@ -1,4 +1,4 @@
-/// @file components/movementsensor/movementsensor.hpp
+/// @file components/movement_sensor/movement_sensor.hpp
 ///
 /// @brief Defines a `MovementSensor` component.
 #pragma once
@@ -21,7 +21,7 @@ namespace sdk {
 
 /// @defgroup MovementSensor Classes related to the MovementSensor component.
 
-/// @class MovementSensor movementsensor.hpp "components/movementsensor/movementsensor.hpp"
+/// @class MovementSensor movementsensor.hpp "components/movement_sensor/movement_sensor.hpp"
 /// @ingroup MovementSensor
 ///
 /// This acts as an abstract base class to be inherited from by any drivers representing
--- a/src/viam/sdk/components/movement_sensor/server.cpp
+++ b/src/viam/sdk/components/movement_sensor/server.cpp
@@ -17,9 +17,10 @@ MovementSensorServer::MovementSensorServer()
 MovementSensorServer::MovementSensorServer(std::shared_ptr<ResourceManager> manager)
     : ResourceServer(manager){};
 
-::grpc::Status MovementSensorServer::GetLinearVelocity(::grpc::ServerContext* context,
-                                                       const GetLinearVelocityRequest* request,
-                                                       GetLinearVelocityResponse* response) {
+::grpc::Status MovementSensorServer::GetLinearVelocity(
+    ::grpc::ServerContext* context,
+    const GetLinearVelocityRequest* request,
+    GetLinearVelocityResponse* response) noexcept {
     return make_service_helper<MovementSensor>("MovementSensorServer::GetLinearVelocity",
                                                this,
                                                request)([&](auto& helper, auto& movementsensor) {
@@ -28,9 +29,10 @@ MovementSensorServer::MovementSensorServer(std::shared_ptr<ResourceManager> mana
     });
 }
 
-::grpc::Status MovementSensorServer::GetAngularVelocity(::grpc::ServerContext* context,
-                                                        const GetAngularVelocityRequest* request,
-                                                        GetAngularVelocityResponse* response) {
+::grpc::Status MovementSensorServer::GetAngularVelocity(
+    ::grpc::ServerContext* context,
+    const GetAngularVelocityRequest* request,
+    GetAngularVelocityResponse* response) noexcept {
     return make_service_helper<MovementSensor>("MovementSensorServer::GetAngularVelocity",
                                                this,
                                                request)([&](auto& helper, auto& movementsensor) {
@@ -39,9 +41,10 @@ MovementSensorServer::MovementSensorServer(std::shared_ptr<ResourceManager> mana
     });
 }
 
-::grpc::Status MovementSensorServer::GetCompassHeading(::grpc::ServerContext* context,
-                                                       const GetCompassHeadingRequest* request,
-                                                       GetCompassHeadingResponse* response) {
+::grpc::Status MovementSensorServer::GetCompassHeading(
+    ::grpc::ServerContext* context,
+    const GetCompassHeadingRequest* request,
+    GetCompassHeadingResponse* response) noexcept {
     return make_service_helper<MovementSensor>("MovementSensorServer::GetCompassHeading",
                                                this,
                                                request)([&](auto& helper, auto& movementsensor) {
@@ -53,7 +56,7 @@ MovementSensorServer::MovementSensorServer(std::shared_ptr<ResourceManager> mana
 
 ::grpc::Status MovementSensorServer::GetOrientation(::grpc::ServerContext* context,
                                                     const GetOrientationRequest* request,
-                                                    GetOrientationResponse* response) {
+                                                    GetOrientationResponse* response) noexcept {
     return make_service_helper<MovementSensor>("MovementSensorServer::GetOrientation",
                                                this,
                                                request)([&](auto& helper, auto& movementsensor) {
@@ -65,7 +68,7 @@ MovementSensorServer::MovementSensorServer(std::shared_ptr<ResourceManager> mana
 
 ::grpc::Status MovementSensorServer::GetPosition(::grpc::ServerContext* context,
                                                  const GetPositionRequest* request,
-                                                 GetPositionResponse* response) {
+                                                 GetPositionResponse* response) noexcept {
     return make_service_helper<MovementSensor>("MovementSensorServer::GetPosition", this, request)(
         [&](auto& helper, auto& movementsensor) {
             const MovementSensor::position result = movementsensor->get_position(helper.getExtra());
@@ -76,7 +79,7 @@ MovementSensorServer::MovementSensorServer(std::shared_ptr<ResourceManager> mana
 
 ::grpc::Status MovementSensorServer::GetProperties(::grpc::ServerContext* context,
                                                    const GetPropertiesRequest* request,
-                                                   GetPropertiesResponse* response) {
+                                                   GetPropertiesResponse* response) noexcept {
     return make_service_helper<MovementSensor>("MovementSensorServer::GetProperties",
                                                this,
                                                request)([&](auto& helper, auto& movementsensor) {
@@ -92,11 +95,10 @@ MovementSensorServer::MovementSensorServer(std::shared_ptr<ResourceManager> mana
 
 ::grpc::Status MovementSensorServer::GetAccuracy(::grpc::ServerContext* context,
                                                  const GetAccuracyRequest* request,
-                                                 GetAccuracyResponse* response) {
+                                                 GetAccuracyResponse* response) noexcept {
     return make_service_helper<MovementSensor>("MovementSensorServer::GetAccuracy", this, request)(
         [&](auto& helper, auto& movementsensor) {
-            const std::unordered_map<std::string, float> result =
-                movementsensor->get_accuracy(helper.getExtra());
+            const auto result = movementsensor->get_accuracy(helper.getExtra());
             for (const auto& i : result) {
                 response->mutable_accuracy()->insert({i.first, i.second});
             }
@@ -106,7 +108,7 @@ MovementSensorServer::MovementSensorServer(std::shared_ptr<ResourceManager> mana
 ::grpc::Status MovementSensorServer::GetLinearAcceleration(
     ::grpc::ServerContext* context,
     const GetLinearAccelerationRequest* request,
-    GetLinearAccelerationResponse* response) {
+    GetLinearAccelerationResponse* response) noexcept {
     return make_service_helper<MovementSensor>("MovementSensorServer::GetLinearAcceleration",
                                                this,
                                                request)([&](auto& helper, auto& movementsensor) {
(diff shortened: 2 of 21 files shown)


src/viam/sdk/registry/registry.hpp
```diff
@@ -58,6 +58,37 @@ class ResourceRegistration {
     const google::protobuf::ServiceDescriptor* service_descriptor_;
 };
 
+// TODO(RSDK-3030): Potentially make ResourceRegistration2 the one and only
+// form of resource registration.
+template <typename ResourceClientT,
+          typename ResourceServerT,
+          typename ProtoServiceT,
+          typename ResourceRegT>
+class ResourceRegistration2 : public ResourceRegistration {
```
benjirewis, line 67: I threw `ResourceRegistration2` into `registry.hpp` thinking that was a good enough "central place" for now. I think we can reconsider this class and its placement in RSDK-3030.

src/viam/sdk/components/sensor/client.hpp
```diff
@@ -0,0 +1,52 @@
...
+    AttributeMap get_readings(const AttributeMap& extra) override;
+    AttributeMap do_command(const AttributeMap& command) override;
+    std::vector<GeometryConfig> get_geometries(const AttributeMap& extra) override;
+
+    typedef viam::component::sensor::v1::SensorService::StubInterface Stub;
+
+   private:
+    // template to wrap a stub
+    template <typename Request, typename Response, typename Cls>
+    void stub_wrapper(Cls* self,
```
benjirewis, line 32: I copied the `stub_wrapper` and `server_wrapper` templated methods from @abe-winter's movement sensor wrapper implementation. I stopped short of moving them to some central location and employing them for every resource.

src/viam/sdk/tests/test_sensor.cpp
```diff
@@ -0,0 +1,100 @@
...
+
+        AttributeMap command = fake_map();
+        AttributeMap result_map = client.do_command(command);
+
+        ProtoType expected_pt = *(expected->at(std::string("test")));
+        ProtoType result_pt = *(result_map->at(std::string("test")));
+        BOOST_CHECK(result_pt == expected_pt);
+    });
+}
+
```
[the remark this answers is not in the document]
benjirewis: Added. Also added a `test_get_geometries` to `test_movement_sensor.cpp`.

src/viam/sdk/tests/mocks/mock_sensor.cpp
```diff
@@ -0,0 +1,34 @@
...
+
+using namespace viam::sdk;
+
+AttributeMap MockSensor::get_readings(const AttributeMap& extra) {
+    return fake_map();
+};
+AttributeMap MockSensor::do_command(const AttributeMap& command) {
+    return command;
+};
+std::vector<sdk::GeometryConfig> MockSensor::get_geometries(const AttributeMap& extra) {
```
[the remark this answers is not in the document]
benjirewis: Ah good call out. Returned `fake_geometries` here and in `MockMovementSensor`.

src/viam/sdk/components/sensor/client.hpp
```diff
@@ -0,0 +1,52 @@
...
+/// @class SensorClient
+/// @brief gRPC client implementation of a `Sensor` component.
+/// @ingroup Sensor
+class SensorClient : public Sensor {
+   public:
+    SensorClient(std::string name, std::shared_ptr<grpc::Channel> channel);
+    AttributeMap get_readings(const AttributeMap& extra) override;
+    AttributeMap do_command(const AttributeMap& command) override;
+    std::vector<GeometryConfig> get_geometries(const AttributeMap& extra) override;
+
```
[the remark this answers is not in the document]
benjirewis: Added; thanks! Couldn't take your suggestion because clang reverses the order (alphabetical).

benjirewis (commented):
Also fixed some "typos" in movement sensor files.

src/viam/sdk/components/sensor/server.hpp
```diff
@@ -0,0 +1,75 @@
...
+        std::shared_ptr<Cls> sensor;
+
+        bool ok() const {
+            return status.ok();
+        }
+    };
+
+    // wrapper template for boilerplate
+    template <typename Cls, typename Req>
+    ServerWrapperResult<Cls> server_wrapper(const Req* request) {
```
acmorrow, line 54: Same thoughts on `server_wrapper`: I think if this is going to become a pattern it should have a closer review and be hoisted out. I'll follow up with some more detailed thoughts.

src/viam/sdk/registry/registry.hpp
acmorrow: To my way of looking at it RSDK-3030 is somewhat independent: it is about the practice of using file-scoped static initializers to make registrations happen, rather than the API of the Registry. But I'm willing to to stretch the scope a little. It is good that the TODO(RSDK-3030) is here, but can we please add a note or update the description of RSDK-3030 so that it includes revisiting this class? I suspect a large part of this can be boiled away into a template method on `Registry` itself, with the majority, if not all, of the details hidden inside `registry.cpp`.

acmorrow (commented):
Overall, this looks fine to me, but I have reservations about the current design of`server_wrapper` and `client_stub` and I don't want to see them go viral by duplication. I'll follow up with some suggestions about how I think they could be improved such that they could be factored out in this review. I missed my chance on the prior review but I didn't mind much because they were just in one place.

src/viam/sdk/tests/test_sensor.cpp
```diff
@@ -0,0 +1,109 @@
...
+#include <vector>
+
+#include <boost/qvm/all.hpp>
+#include <boost/test/included/unit_test.hpp>
+
+#include <viam/api/common/v1/common.pb.h>
+#include <viam/api/component/sensor/v1/sensor.grpc.pb.h>
+#include <viam/api/component/sensor/v1/sensor.pb.h>
+
+#include "viam/sdk/spatialmath/geometry.hpp"
```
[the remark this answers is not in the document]
benjirewis: Auto -included by my IDE and a relative include was used; switched.

src/viam/sdk/registry/registry.hpp
benjirewis: Good point; updated the description of https://viam.atlassian.net/browse/RSDK-3030.

src/viam/sdk/components/sensor/client.hpp
benjirewis: Sounds good; I'll wait for your feedback before proceeding, then. Thanks for thinking about it, would be great to standardize these in an idiomatic way now before we implement more wrappers (although only the power sensor is in my TODO for now).

src/viam/sdk/components/movement_sensor/movement_sensor.hpp
```diff
@@ -21,7 +21,7 @@ namespace sdk {
 
 /// @defgroup MovementSensor Classes related to the MovementSensor component.
 
-/// @class MovementSensor movementsensor.hpp "components/movementsensor/movementsensor.hpp"
+/// @class MovementSensor movementsensor.hpp "components/movement_sensor/movement_sensor.hpp"
```
stuqdog, line 24: `s/movementsensor.hpp/movement_sensor.hpp`

stuqdog (approved):

src/viam/sdk/components/sensor/server.hpp
acmorrow: OK, I spent a little time thinking about this one and I have something that I think will work pretty OK. It solves two issues: removing some of the boilerplate on server method implementation entry, and getting rid of the need for ugly funtion-scoped try blocks for every server method.

First, we add a little convenience method to `ResourceManager`:

```diff
     /// @throws `std::runtime_error` if the desired resource does not exist.
     std::shared_ptr<Resource> resource(const std::string& name);

+    template<typename T>
+    std::shared_ptr<T> resource(const std::string& name) {
+        return std::dynamic_pointer_cast<T>(resource(name));
+    }
+
```
There is no ambiguity here because the template will never be preferred.

Then we write a little wrapper class that takes an arbitrary callable and a factory function for it:

```diff
+template <typename ServiceType, typename RequestType>
+class RequestWrapper {
+   public:
+    RequestWrapper(ResourceServer* rs, RequestType* request) noexcept
+        : rs_(rs), request_(request){};
+
+    template <typename Callable>
+    ::grpc::Status operator()(Callable&& callable) noexcept try {
+        if (!request_) {
+            return {::grpc::INVALID_ARGUMENT, "TODO ERROR MESSAGE"};
+        }
+        const auto resource = rs_->resource_manager()->resource<ServiceType>(request_->n
ame());
+        if (!resource) {
+            return {::grpc::UNKNOWN, "TODO ERROR MESSAGE"};
+        }
+        AttributeMap extra;
+        if (request_->has_extra()) {
+            extra = struct_to_map(request_->extra());
+        }
+        return std::forward<Callable>(callable)(resource, extra);
+    } catch (const std::exception& xcp) {
+        return {grpc::INTERNAL, xcp.what()};
+    } catch (...) {
+        return {grpc::INTERNAL, "TODO ERROR MESSAGE"};
+    }
+
+   private:
+    ResourceServer* rs_;
+    RequestType* request_;
+};
+
+template <typename ServiceType, typename RequestType>
+auto make_request_wrapper(ResourceServer* rs, RequestType* request) {
+    return RequestWrapper<ServiceType, RequestType>{rs, request};
+}
+
```

Note that to produce good error messages, somehow the class / function as a string should be a parameter or a template arg! That's not done here.

Then you can remove all of the on-entry boilerplate (I'm using the `MLModelServiceServer` here as a demo:

```diff
 ::grpc::Status MLModelServiceServer::Infer(
     ::grpc::ServerContext* context,
     const ::viam::service::mlmodel::v1::InferRequest* request,
-    ::viam::service::mlmodel::v1::InferResponse* response) noexcept try {
-    if (!request) {
-        return {::grpc::StatusCode::INVALID_ARGUMENT, "Called [Infer] without a request"
};
-    };
-
-    std::shared_ptr<Resource> rb = resource_manager()->resource(request->name());
-
-    if (!rb) {
-        return {::grpc::UNKNOWN, "resource not found: " + request->name()};
-    }
-
-    std::shared_ptr<MLModelService> mlms = std::dynamic_pointer_cast<MLModelService>(rb)
;
-
-    if (!request->has_input_tensors()) {
-        return {::grpc::StatusCode::INVALID_ARGUMENT, "Called [Infer] with no inputs"};
-    }
-
-    AttributeMap extra;
-    if (request->has_extra()) {
-        extra = struct_to_map(request->extra());
-    }
```

And replace it with the existing interesating part of the body pushed into a lambda passed to the wrapper:

```diff
+    ::viam::service::mlmodel::v1::InferResponse* response) noexcept {
+    return make_request_wrapper<MLModelService>(
+        this, request)([&](auto mlms, const auto& extra) -> ::grpc::Status {
+        if (!request->has_input_tensors()) {
```
And then the whole function-scoped-try catch handler drops out:

```diff
- } catch (...) {
-    // TODO(RSDK-3556): This is pretty bad. Investigate ways to
-    // simplify this, or decide against having clang-tidy enforce
-    // noexcept, or decide that letting gRPC deal with thrown
-    // exceptions is OK.
-    try {
-        try {
-            throw;
-        } catch (const std::exception& ex) {
-            return ::grpc::Status(grpc::INTERNAL,
-                                  std::string("[Infer]: Failed with exception: ") + ex.what());
-        } catch (...) {
-            return ::grpc::Status(grpc::INTERNAL, "[Infer]: Failed with an unknown exception");
-    } catch (...) {
-        std::abort();
-    }
```

Though I expect that when you put real error messages into the wrapper you will need to reintroduce something similar. That's better though: at least the horrible thing is now in only one place.

Overall, this seems like a pretty decent improvement on the server side to me. I think there are a few things that could be improved beyond what's already here:

- As noted, providing more context into the wrapper about what class and method we are in. So, passing `"MLModelServiceServer::Infer"` as either a template param or argument to the wrapper factory / ctor would be needed.
- The handling around the Status return is still clunky. If the contained lambda returned void and communicated all errors with exceptions, then the wrapper could just call the `Callback` and return an ok Status, and handle all exceptions. But getting there would require re-writing all the handlers. I think it is possible to use SFINAE to make a wrapper which detects whether the return type of the callback is void or Status, and deals with it appropriately. That'd allow us to incrementally migrate: you could adopt the wrapper without re-writing the code that currently returns Status, but new code could just return `void` and throw errors.
- Similarly, I dislike that the wrapper only works with `Request` types that have an `extra` parameter. I think it is also possible to SFINAE our way out of that and detect whether or not the `Request` type has an extra param and call a differently signatured callback.

I'll try to post an update soon with details on how to do both of those SFINAE tricks.

I'm also going to undertake the same investigation for the client side. Please hold off on merging this until I've written up my thoughts on that.

acmorrow: Please see #178 for a fairly complete PoC implementation of this.

src/viam/sdk/components/movement_sensor/server.cpp
```diff
@@ -19,174 +20,124 @@ MovementSensorServer::MovementSensorServer(std::shared_ptr<ResourceManager> mana
...
                                                  const GetAccuracyRequest* request,
                                                  GetAccuracyResponse* response) {
-    const auto status = server_wrapper<MovementSensor>(request);
-    if (!status.ok()) {
-        return status.status;
-    }
-
-    std::unordered_map<std::string, float> result =
-        status.movementsensor->get_accuracy(status.extra);
-    response->mutable_accuracy()->empty();
```
benjirewis, line 120: I removed this line; I don't think it did anything (my guess is it was an attempt to "empty" the `mutable_accuracy` map, but `empty` just returns a boolean describing if the map _is_ empty that we were ignoring.

benjirewis (commented):
Used @acmorrow's `ServiceHelper` class for both `SensorServer` and `MovementSensorServer`. Removed custom `ServerWrapperResult` structs and `server_wrapper` methods.

src/viam/sdk/components/sensor/server.hpp
```diff
@@ -0,0 +1,42 @@
...
+/// @ingroup Sensor
+class SensorServer : public ResourceServer,
+                     public viam::component::sensor::v1::SensorService::Service {
+   public:
+    SensorServer();
+    explicit SensorServer(std::shared_ptr<ResourceManager> manager);
+
+    ::grpc::Status GetReadings(::grpc::ServerContext* context,
+                               const GetReadingsRequest* request,
+                               GetReadingsResponse* response) override;
```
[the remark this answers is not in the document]
benjirewis: Oops! Thank you; done.

src/viam/sdk/components/sensor/server.cpp
```diff
@@ -0,0 +1,56 @@
...
+SensorServer::SensorServer(std::shared_ptr<ResourceManager> manager) : ResourceServer(manager){};
+
+::grpc::Status SensorServer::GetReadings(::grpc::ServerContext* context,
+                                         const GetReadingsRequest* request,
+                                         GetReadingsResponse* response) {
+    return make_service_helper<Sensor>(
+        "SensorServer::GetReadings", this, request)([&](auto& helper, auto& sensor) {
+        const AttributeMap result = sensor->get_readings(helper.getExtra());
+        for (const auto& r : *result) {
+            response->mutable_readings()->insert({std::move(r.first), r.second->proto_value()});
```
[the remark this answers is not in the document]
benjirewis: Removed.

src/viam/sdk/components/sensor/client.hpp
benjirewis: SGTM!

acmorrow (approved):
LGTM for my part (wrappers and such).

PR #175, opened 2023-11-09
RSDK-5681 Do not inject complex module stuff into viamsdk_test

benjirewis:
https://viam.atlassian.net/browse/RSDK-5681

Only adds gizmo and summation API source files to `test_complex_module` binary (copy-pasted from the `viamcppsdk_add_boost_test` cmake function). Only adds the generated complex module files as an include directory to the `test_complex_module` binary.

@acmorrow found an issue where I was injecting some complex module stuff into `viamsdk_test` needlessly. This was breaking the encapsulation of the tests.

### DIFF
--- a/src/viam/examples/modules/complex/CMakeLists.txt
+++ b/src/viam/examples/modules/complex/CMakeLists.txt
@@ -90,60 +90,47 @@ add_custom_target(
   DEPENDS ${MODULE_PROTO_GEN_DIR}/gizmo.grpc.pb.cc
 )
 
-add_executable(complex_module)
-# We have to include the generated proto files in order to #include the files.
-target_include_directories(complex_module
+# Create an OBJECT library for the generated output files. Include the whole
+# generated directory so libraries that link to complex_module_sources
+# transitively include all generated files. Also link to the viamsdk library
+# so following executables do not have to.
+add_library(complex_module_sources
+  OBJECT
+    gizmo/api.cpp
+    summation/api.cpp
+    ${MODULE_PROTO_OUTPUT_FILES}
+)
+target_include_directories(complex_module_sources
   PUBLIC
     ${MODULE_PROTO_GEN_DIR}
 )
+target_link_libraries(complex_module_sources
+  viam-cpp-sdk::viamsdk
+)
+
+add_executable(complex_module)
 target_sources(complex_module
   PRIVATE
     main.cpp
     base/impl.cpp
-    base/impl.hpp
     gizmo/impl.cpp
-    gizmo/impl.hpp
-    gizmo/api.cpp
-    gizmo/api.hpp
-    summation/api.cpp
-    summation/api.hpp
     summation/impl.cpp
-    summation/impl.hpp
-
-    ${MODULE_PROTO_OUTPUT_FILES}
 )
-
-add_executable(complex_module_client)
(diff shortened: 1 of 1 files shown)


src/viam/examples/modules/complex/CMakeLists.txt
```diff
@@ -154,21 +154,31 @@ install(
 # `test_complex_module.cpp` into the Viam C++ SDK testing suite.
 
 enable_testing()
-target_sources(viamsdk_test
+add_executable(test_complex_module test_complex_module.cpp)
+target_link_libraries(test_complex_module viamsdk_test)
+target_sources(test_complex_module
   PRIVATE
     gizmo/impl.cpp
-    gizmo/impl.hpp
```
acmorrow, line 160: FYI you can make this same change up above for `complex_module` itself - you don't need to list the header files in the `target_sources`. Same for `complex_module_client`.

src/viam/examples/modules/complex/CMakeLists.txt
```diff
@@ -154,21 +154,31 @@ install(
...
 
     ${MODULE_PROTO_OUTPUT_FILES}
 )
-target_include_directories(viamsdk_test
+target_include_directories(test_complex_module
   PUBLIC
     ${MODULE_PROTO_GEN_DIR}
 )
-viamcppsdk_add_boost_test(test_complex_module.cpp)
+
```
acmorrow, line 172: What prevents you from continuing to use `viamcppsdk_add_boost_test` here, and then calling `target_sources`, `target_include_directories` and `target_link_libraries` on the resulting executable to add in what you need? I believe those calls are all addidative:

```
viamcppsdk_add_boost_test(test_complex_module.cpp)

target_sources(test_complex_module ...)
target_include_directories(test_complex_module ...)
target_link_libraries(test_complex_module...)
```

Also, I note that there are now three places where the generated sources are listed as `target_sources`: once in the module, once in the client, and once in the test.

But I think you can use the CMake "object libraries" to eliminate that: https://cmake.org/cmake/help/latest/manual/cmake-buildsystem.7.html#object-libraries.

You would want to declare two: one for the client side sources and one for the server. Then you can name those in `target_link_libraries`.

I think you could even then use `target_include_directories` on the object library to attach `${MODULE_PROTO_GEN_DIR}` as `PUBLIC` and then it would transitively propagate to the libraries that link to the object libraries.

Can you give that a try?

acmorrow (changes requested):

stuqdog (commented):
Seems reasonable to me but I'd like to see how @acmorrow's feedback is integrated.

src/viam/examples/modules/complex/CMakeLists.txt
benjirewis: Good point; done.

src/viam/examples/modules/complex/CMakeLists.txt
acmorrow: You did, I was imagining that the sources decomposed cleanly into client-y and server-y groups, but if that's not the case (and looking more closely I don't think it is), then there is no need for two `OBJECT` libraries and what you have done here is fine.

acmorrow (approved):
Looks great!

stuqdog (approved):
lgtm, thanks!

PR #176, opened 2023-11-09
RSDK-4425 Power sensor wrappers

benjirewis:
https://viam.atlassian.net/browse/RSDK-4425

Adds wrappers/tests for the power sensor.

(cc @acmorrow)

### DIFF
--- a/src/viam/sdk/components/power_sensor/client.hpp
+++ b/src/viam/sdk/components/power_sensor/client.hpp
@@ -1,0 +1,60 @@
+/// @file components/power_sensor/client.hpp
+///
+/// @brief Implements a gRPC client for the `PowerSensor` component.
+#pragma once
+
+#include <grpcpp/channel.h>
+
+#include <viam/api/component/powersensor/v1/powersensor.grpc.pb.h>
+
+#include <viam/sdk/components/power_sensor/power_sensor.hpp>
+#include <viam/sdk/components/power_sensor/server.hpp>
+#include <viam/sdk/config/resource.hpp>
+#include <viam/sdk/robot/client.hpp>
+
+namespace viam {
+namespace sdk {
+
+/// @class PowerSensorClient
+/// @brief gRPC client implementation of a `PowerSensor` component.
+/// @ingroup PowerSensor
+class PowerSensorClient : public PowerSensor {
+   public:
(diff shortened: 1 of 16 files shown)


benjirewis:
Will leave this is in draft until #174 is merged and we decide on a clearer path forward for server and client wrappers for basic gRPC handler methods.

src/viam/sdk/components/power_sensor/client.hpp
```diff
@@ -0,0 +1,60 @@
...
+    using PowerSensor::get_power;
+    using PowerSensor::get_readings;
+    using PowerSensor::get_voltage;
+
+    typedef viam::component::powersensor::v1::PowerSensorService::StubInterface Stub;
+
+   private:
+    // template to wrap a stub
+    template <typename Request, typename Response, typename Cls>
+    void stub_wrapper(Cls* self,
```
benjirewis, line 40: Not sure if we wanted to merge #180 before this; happy to remove this and use the `ClientHelper` if so. I'm thinking of using https://viam.atlassian.net/browse/RSDK-3556 to propagate `ClientHelper` and `ServiceHelper` everywhere else.

acmorrow: I don't think it matters which way they merge. If you merge this before #180, then I will convert the `stub_wrappers` added in this review over to `ClientHelper` calls. If it goes the other way, you can take it in this review before merge.

I think RSDK-3556 is a fine ticket to use to track the conversion, and I agree that it should be done with as few fly-by changes as possible. In particular, I'd recommend explicitly calling out any such fly-by changes with preemptive comments on the PR so that reviewers are aware of them. The temptation to skim such a large and mechanical review is high.

acmorrow (commented):
I don't have much to add on this review, and I'm happy with any sequencing w.r.t. to the ClientHelpers. No need to wait for an approval from me.

acmorrow (approved):

stuqdog (approved):
lgtm!

src/viam/sdk/components/power_sensor/client.hpp
benjirewis: Since I'm OOO tomorrow I'll merge this PR before #180.

PR #178, opened 2023-11-10
RSDK-794 PoC ServiceHelper to simplify service method implementations

acmorrow:
This is my PoC response to the discussion in the Sensor review: https://github.com/viamrobotics/viam-cpp-sdk/pull/174/files#r1386948048

There is a lot to understand in the `ServiceHelper` and `ServiceHelperBase` classes, but I think the boilerplate reduction that it enables (which can be seen most clearly in the `encoder` server changes) is pretty compelling.

I changed the `MLModelServiceServer` class to show how the `ServiceHelper` works in the case where the RPC handler needs to return custom errors and therefore the lambda traffics in `Status` returns, and also did the `EncoderServer` case to show that it also works when there are no `Status` returns and the lambda therefore returns void.

Handling both cases is dealt with through some SFINAE magic.

If this sort of approach is appealing (or at least acceptable) to those on the review list, I'll continue this PR with a helper for client side boilerplate as well, as a response to the `stub_wrapper` also in #174.

Various reviewers on this:
- @benjirewis since it is in response to #174.
- @stuqdog as project owners
- @abe-winter since he wrote the original `stub_wrapper` and `ServerWrapper` boilerplate reduction concepts.
- @zaporter-work since I think he will be interested in this w.r.t. his error handling thoughts.

Please let me know your thoughts.

### DIFF
--- a/src/viam/sdk/CMakeLists.txt
+++ b/src/viam/sdk/CMakeLists.txt
@@ -46,6 +46,7 @@ target_sources(viamsdk
     common/linear_algebra.cpp
     common/pose.cpp
     common/proto_type.cpp
+    common/service_helper.cpp
     common/utils.cpp
     common/world_state.cpp
     components/base/base.cpp
--- a/src/viam/sdk/common/service_helper.cpp
+++ b/src/viam/sdk/common/service_helper.cpp
@@ -21,9 +21,9 @@ namespace sdk {
 ::grpc::Status ServiceHelperBase::failNoResource(const std::string& name) const noexcept try {
     std::ostringstream stream;
     stream << "Failed to find resource `" << name << "`";
-    return fail(::grpc::INVALID_ARGUMENT, stream.str().c_str());
+    return fail(::grpc::NOT_FOUND, stream.str().c_str());
 } catch (...) {
-    return fail(::grpc::INVALID_ARGUMENT, "Failed to find resource");
+    return fail(::grpc::NOT_FOUND, "Failed to find resource");
 }
 
 ::grpc::Status ServiceHelperBase::failStdException(const std::exception& xcp) const noexcept try {
--- a/src/viam/sdk/common/service_helper.cpp
+++ b/src/viam/sdk/common/service_helper.cpp
@@ -1,0 +1,40 @@
+#include <viam/sdk/common/service_helper.hpp>
+
+namespace viam {
+namespace sdk {
+
+::grpc::Status ServiceHelperBase::fail(::grpc::StatusCode code, const char* message) const noexcept
+    try {
+    std::ostringstream stream;
+    stream << '[' << method_ << "]: " << message;
+    return {code, stream.str()};
+} catch (...) {
+    return {code, message};
+}
+
+::grpc::Status ServiceHelperBase::failNoRequest() const noexcept {
+    return fail(::grpc::INVALID_ARGUMENT, "Called without a `request` object");
+}
+
+::grpc::Status ServiceHelperBase::failNoResource(const std::string& name) const noexcept try {
+    std::ostringstream stream;
+    stream << "Failed to find resource `" << name << "`";
+    return fail(::grpc::INVALID_ARGUMENT, stream.str().c_str());
+} catch (...) {
+    return fail(::grpc::INVALID_ARGUMENT, "Failed to find resource");
+}
+
+::grpc::Status ServiceHelperBase::failStdException(const std::exception& xcp) const noexcept try {
+    std::ostringstream stream;
+    stream << "Failed with a std::exception: " << xcp.what();
+    return fail(::grpc::INTERNAL, stream.str().c_str());
+} catch (...) {
+    return fail(::grpc::INTERNAL, "Failed with a std::exception: <unknown>");
+}
+
+::grpc::Status ServiceHelperBase::failUnknownException() const noexcept {
+    return fail(::grpc::INTERNAL, "Failed with an unknown exception");
+}
+
+}  // namespace sdk
+}  // namespace viam
--- a/src/viam/sdk/common/service_helper.cpp
+++ b/src/viam/sdk/common/service_helper.cpp
@@ -1,5 +1,7 @@
 #include <viam/sdk/common/service_helper.hpp>
 
+#include <sstream>
+
 namespace viam {
 namespace sdk {
 
--- a/src/viam/sdk/common/service_helper.hpp
+++ b/src/viam/sdk/common/service_helper.hpp
@@ -30,7 +30,7 @@ template <typename ServiceType, typename RequestType>
 class ServiceHelper : public ServiceHelperBase {
    public:
     ServiceHelper(const char* method, ResourceServer* rs, RequestType* request) noexcept
-        : ServiceHelperBase(method), rs_{rs}, request_{request} {};
+        : ServiceHelperBase{method}, rs_{rs}, request_{request} {};
 
     template <typename Callable>
     ::grpc::Status operator()(Callable&& callable) const noexcept try {
--- a/src/viam/sdk/common/service_helper.hpp
+++ b/src/viam/sdk/common/service_helper.hpp
@@ -1,0 +1,93 @@
+#pragma once
+
+#include <type_traits>
+
+#include <viam/sdk/resource/resource_server_base.hpp>
+
+namespace viam {
+namespace sdk {
+
+class ServiceHelperBase {
+   public:
+    ::grpc::Status fail(::grpc::StatusCode code, const char* message) const noexcept;
+
+    ::grpc::Status failNoRequest() const noexcept;
+
+    ::grpc::Status failNoResource(const std::string& name) const noexcept;
+
+    ::grpc::Status failStdException(const std::exception& xcp) const noexcept;
+
+    ::grpc::Status failUnknownException() const noexcept;
+
+   protected:
+    explicit ServiceHelperBase(const char* method) noexcept : method_{method} {}
+
+   private:
+    const char* method_;
+};
+
+template <typename ServiceType, typename RequestType>
+class ServiceHelper : public ServiceHelperBase {
+   public:
+    ServiceHelper(const char* method, ResourceServer* rs, RequestType* request) noexcept
+        : ServiceHelperBase(method), rs_{rs}, request_{request} {};
+
+    template <typename Callable>
+    ::grpc::Status operator()(Callable&& callable) const noexcept try {
+        if (!request_) {
+            return failNoRequest();
+        }
+        const auto resource = rs_->resource_manager()->resource<ServiceType>(request_->name());
+        if (!resource) {
+            return failNoResource(request_->name());
+        }
+        return invoke_(std::forward<Callable>(callable), std::move(resource));
+    } catch (const std::exception& xcp) {
+        return failStdException(xcp);
+    } catch (...) {
+        return failUnknownException();
+    }
+
+    auto getExtra() const {
+        return request_->has_extra() ? struct_to_map(request_->extra()) : AttributeMap{};
+    }
+
+   private:
+    template <typename Callable, typename... Args>
+    using is_void_result = std::is_void<std::result_of_t<Callable(Args...)>>;
+
+    // Implementation of `invoke_` for a Callable returning non-void,
+    // presumably an error return, which we return as a
+    // ::grpc::Status.
+    template <typename Callable,
+              typename ResourcePtrType,
+              std::enable_if_t<!is_void_result<Callable, ServiceHelper&, ResourcePtrType&&>::value,
+                               bool> = true>
+    ::grpc::Status invoke_(Callable&& callable, ResourcePtrType&& resource) const {
+        return std::forward<Callable>(callable)(*this, std::forward<ResourcePtrType>(resource));
+    }
+
+    // Implementation of `invoke_` for a Callable returning void,
+    // which is therefore either non-failing or communicates errors by
+    // throwing exceptions. We return an OK status automatically.
+    template <typename Callable,
+              typename ResourcePtrType,
+              std::enable_if_t<is_void_result<Callable, ServiceHelper&, ResourcePtrType&&>::value,
+                               bool> = true>
+    ::grpc::Status invoke_(Callable&& callable, ResourcePtrType&& resource) const {
+        std::forward<Callable>(callable)(*this, std::forward<ResourcePtrType>(resource));
+        return {};
+    }
+
+    const char* method_;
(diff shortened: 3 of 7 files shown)


src/viam/sdk/common/service_helper.hpp
```diff
@@ -0,0 +1,92 @@
...
+   public:
+    ::grpc::Status fail(::grpc::StatusCode code, const char* message) const noexcept;
+
+    ::grpc::Status failNoRequest() const noexcept;
+
+    ::grpc::Status failNoResource(const std::string& name) const noexcept;
+
+    ::grpc::Status failStdException(const std::exception& xcp) const noexcept;
+
+    ::grpc::Status failUnknownException() const noexcept;
```
acmorrow, line 20: These functions could be made `protected` instead since they are really only used in the implementation of `ServiceHelper::operator()`, but I didn't see much need to hide them away.

src/viam/sdk/common/service_helper.hpp
```diff
@@ -0,0 +1,92 @@
...
+            return failNoResource(request_->name());
+        }
+        return invoke_(std::forward<Callable>(callable), std::move(resource));
+    } catch (const std::exception& xcp) {
+        return failStdException(xcp);
+    } catch (...) {
+        return failUnknownException();
+    }
+
+    auto getExtra() const {
```
acmorrow, line 51: Another way to go about this would be to use SFINAE again to have the Callable be invoked with `extra`, but only if `RequestType` defines `has_request`. But it felt like the template metaprogramming complexity wasn't worth it.

src/viam/sdk/services/mlmodel/server.cpp
```diff
@@ -34,184 +35,134 @@ void MLModelServiceServer::register_server(std::shared_ptr<Server> server) {
...
-            // we can't validate the type info.
-            //
-            // TODO: Should this be an error? For now we just don't decode
-            // those tensors.
-            continue;
+    ::viam::service::mlmodel::v1::InferResponse* response) noexcept {
+    return make_service_helper<MLModelService>(
+        "MLModelServiceServer::Infer", this, request)([&](auto& helper, auto& mlms) {
+        if (!request->has_input_tensors()) {
+            return helper.fail(::grpc::INVALID_ARGUMENT, "Called with no input tensors");
```
acmorrow, line 42: Potentially, we could add new `fail` functions like `failInvalidArgument`, etc.

src/viam/sdk/services/mlmodel/server.cpp
```diff
@@ -34,184 +35,134 @@ void MLModelServiceServer::register_server(std::shared_ptr<Server> server) {
...
-        const auto tensor_type = MLModelService::tensor_info::tensor_views_to_data_type(tensor);
-        if (tensor_type != input.data_type) {
-            std::ostringstream message;
-            using ut = std::underlying_type<MLModelService::tensor_info::data_types>::type;
-            message << "Tensor input `" << input.name << "` was the wrong type; expected type "
-                    << static_cast<ut>(input.data_type) << " but got type "
-                    << static_cast<ut>(tensor_type);
-            return ::grpc::Status(grpc::INVALID_ARGUMENT, message.str());
+
+        const auto md = mlms->metadata({});
```
acmorrow, line 45: Previously this passed `extra` into this call to `metadata` but I think that was unnecessary. The `extra` parameters here are intended for the `Infer` call. The implementation of `Infer` needs access to the metadata, but that doesn't mean it should also be given the same `extra` params that were passed to `Infer`.

src/viam/sdk/CMakeLists.txt
```diff
@@ -46,6 +46,7 @@ target_sources(viamsdk
     common/linear_algebra.cpp
     common/pose.cpp
     common/proto_type.cpp
+    common/service_helper.cpp
```
acmorrow, line 49: Currently, I'm not publishing the `service_helper.hpp` header, so `ServiceHelper` is effectively a private utility of the SDK and only we can use it to implement our services. If we want others (module authors) to also have access to it, we should publish the header.

src/viam/sdk/common/service_helper.cpp
```diff
@@ -0,0 +1,42 @@
+#include <viam/sdk/common/service_helper.hpp>
+
+#include <sstream>
+
+namespace viam {
+namespace sdk {
+
+::grpc::Status ServiceHelperBase::fail(::grpc::StatusCode code, const char* message) const noexcept
+    try {
+    std::ostringstream stream;
```
acmorrow, line 10: It is a little wasteful that some paths will create multiple `ostringstream` objects along the way, but it is the error path so I'm not too concerned about it.

src/viam/sdk/CMakeLists.txt
benjirewis: I would be in favor of exposing it for module authors. When I was writing the complex module example I definitely looked at the existing server implementations for inspiration; it would be frustrating if those "examples" were leveraging utilities I didn't have access to.

src/viam/sdk/common/service_helper.cpp
```diff
@@ -0,0 +1,42 @@
...
+    stream << '[' << method_ << "]: " << message;
+    return {code, stream.str()};
+} catch (...) {
```
benjirewis, line 13: What exception are you `catch`ing here and in methods below? `str()` throwing?

src/viam/sdk/common/service_helper.hpp
benjirewis: Agree; the helper method is much easier to read given the existing meta programming.

src/viam/sdk/common/service_helper.hpp
benjirewis: What's the purpose of having both `ServiceHelper` and `ServiceHelperBase` if you're not hiding these methods and there are no other inheritors of `ServiceHelperBase`?

src/viam/sdk/common/service_helper.hpp
```diff
@@ -0,0 +1,92 @@
...
+   private:
+    template <typename Callable, typename... Args>
+    using is_void_result = std::is_void<std::result_of_t<Callable(Args...)>>;
+
+    // Implementation of `invoke_` for a Callable returning non-void,
+    // presumably an error return, which we return as a
+    // ::grpc::Status.
+    template <typename Callable,
+              typename ResourcePtrType,
+              std::enable_if_t<!is_void_result<Callable, ServiceHelper&, ResourcePtrType&&>::value,
```
benjirewis, line 64: This is so cool 🧑‍🔧 .

src/viam/sdk/components/encoder/server.cpp
```diff
@@ -14,132 +15,57 @@ EncoderServer::EncoderServer(std::shared_ptr<ResourceManager> manager) : Resourc
...
-
-    const Encoder::position result =
-        encoder->get_position(extra, Encoder::from_proto(request->position_type()));
-    response->set_value(result.value);
-    response->set_position_type(Encoder::to_proto(result.type));
-
-    return ::grpc::Status();
+    ::viam::component::encoder::v1::GetPositionResponse* response) noexcept {
+    return make_service_helper<Encoder>(
+        "EncoderServer::GetPosition", this, request)([&](auto& helper, auto& encoder) {
```
benjirewis, line 20: I wish there were a way to pull the method name from the request or something; I couldn't find anything clean.

src/viam/sdk/resource/resource_manager.hpp
```diff
@@ -27,6 +28,15 @@ class ResourceManager {
     /// @throws `std::runtime_error` if the desired resource does not exist.
     std::shared_ptr<Resource> resource(const std::string& name);
 
+    /// @brief Returns a resource after dynamically downcasting to `T`.
+    /// @param name of the desired resource.
+    /// @throws `std::runtime_error` if the desired resource does not exist.
+    template <typename T>
+    std::shared_ptr<T> resource(const std::string& name) {
```
benjirewis, line 35: Nice; I wonder if there could be any interaction between this and the templated [`resource_by_name`](https://github.com/viamrobotics/viam-cpp-sdk/blob/main/src/viam/sdk/robot/client.hpp#L91) method on the robot client.

src/viam/sdk/services/mlmodel/server.cpp
benjirewis: Yeah `failInvalidArgument` in particular seems helpful. One for every [gRPC status code](https://grpc.github.io/grpc/core/md_doc_statuscodes.html) could be cool but is probably overkill for now.

src/viam/sdk/services/mlmodel/server.cpp
benjirewis: The only actual implementations of that method I've seen are in the tflite module and the mock mlmodel service, neither of which use `extra`. I don't really see why it's a necessary argument to `metadata` either, so I'm fine removing it unless @stuqdog disagrees.

benjirewis (commented):
Have some questions but I'm definitely on board with this. 

Why do we feel a client side analog to this is necessary? From what I understand, we don't have to wrap thrown errors in gRPC statuses on the client side, and I'm not sure I see why a `ClientHelper` class would be more useful/readable than just copy pasting client code (I suppose there's less of a chance to miss a `nullptr` check).

src/viam/sdk/common/service_helper.cpp
acmorrow: More the `<<` operators. But the case must be handled or it isn't noexcept. We could also decide to `abort` in the `catch (...)` case on the argument that we were already in such trouble that we couldn't return a real error.

src/viam/sdk/common/service_helper.hpp
acmorrow: Because these functions don't depend on the template parameters like `ServiceType` and `RequestType`. By moving them into a base class, there is just one definition of these functions, rather than one definition per template instantiation. It also means the implementation can be out-of-line.

src/viam/sdk/common/service_helper.hpp
acmorrow: It is much easier with Concepts but this is as good as it gets in C++14. But yeah, SFINAE is sort of magical.

src/viam/sdk/components/encoder/server.cpp
acmorrow: Yes, that would be nice, but there really isn't a good way to get the string you want here.

src/viam/sdk/resource/resource_manager.hpp
acmorrow: Not sure quite what you mean by "interaction", can you explain a little further what you have in mind?

src/viam/sdk/common/service_helper.cpp
```diff
@@ -0,0 +1,42 @@
...
+}
+
+::grpc::Status ServiceHelperBase::failNoRequest() const noexcept {
+    return fail(::grpc::INVALID_ARGUMENT, "Called without a `request` object");
+}
+
+::grpc::Status ServiceHelperBase::failNoResource(const std::string& name) const noexcept try {
+    std::ostringstream stream;
+    stream << "Failed to find resource `" << name << "`";
+    return fail(::grpc::INVALID_ARGUMENT, stream.str().c_str());
```
acmorrow, line 24: I wonder a bit if this should actually be `::grpc::NOT_FOUND`.

benjirewis: More believable than `INVALID_ARGUMENT` imo. We certainly have not standardized what situations should result in what error codes across SDKs.

src/viam/sdk/common/service_helper.cpp
benjirewis: Got it. Current design seems fine to me, then.

src/viam/sdk/resource/resource_manager.hpp
benjirewis: I _think_ `resource_by_name` could just call this method; no need to do it here (if it's possible), just wanted to note that I've seen a similar method lying around.

src/viam/sdk/common/service_helper.hpp
benjirewis: Yeah semantic categories are hugely useful. I was initially surprised when `template` didn't really (seem to) offer that.

src/viam/sdk/CMakeLists.txt
benjirewis: > We could keep it private for now until we have more experience with it, and then release it.

SGTM especially if we're introducing this for only a single component and service for the time being.

src/viam/sdk/common/service_helper.hpp
benjirewis: > rather than one definition per template instantiation

As in the compiler would create a new definition per template instantiation? Is that bad?

> It also means the implementation can be out-of-line.

Gotcha.. didn't think about that constraint.

acmorrow: > > rather than one definition per template instantiation
> 
> As in the compiler would create a new definition per template instantiation? Is that bad?
> 

It'll cost you at least compile time. All object files will be a little bigger. Maybe bloat in the linked image unless the compiler figures out that it can common them all together. Harder to set breakpoints. More recompile if you ever need to change the definition. Etc.

benjirewis: Cool makes sense! Thanks.

benjirewis (approved):
Awesome! I assume it'd be ok to merge this if others LGTMed and I can propagate the pattern to at least all the sensor wrappers.

src/viam/sdk/services/mlmodel/server.cpp
stuqdog: Sounds reasonable to me! I added `extra` there out of an abundance of caution when adding `extra` overloads, not out of some principled stance on it being necessary to `metadata`.

src/viam/sdk/common/service_helper.cpp
stuqdog: agreed, `NOT_FOUND` seems more intuitive.

stuqdog (approved):
Awesome, looks great to me! Thanks for putting this together.

src/viam/sdk/common/service_helper.cpp
acmorrow: Done

acmorrow:
I'm glad everyone likes how this turned out. I'll get it merged. I have a similar idea about how to remove client side boilerplate but I'm not 100% happy with it yet. I'll push a new PR when I have something I think is worth reviewing.

PR #179, opened 2023-11-10
RSDK-5194 - use factory function for client ctx

stuqdog:
Uses a factory function to generate a client ctx with a set authority to minimize both runtime errors and boilerplate.

### DIFF
--- a/src/viam/examples/modules/complex/gizmo/api.cpp
+++ b/src/viam/examples/modules/complex/gizmo/api.cpp
@@ -211,10 +211,9 @@ bool GizmoClient::do_one(std::string arg1) {
 
 bool GizmoClient::do_one_client_stream(std::vector<std::string> arg1) {
     DoOneClientStreamResponse response;
-    grpc::ClientContext ctx;
-    set_client_ctx_authority(ctx);
+    ClientContext ctx;
 
-    auto writer(stub_->DoOneClientStream(&ctx, &response));
+    auto writer(stub_->DoOneClientStream(ctx, &response));
     for (std::string arg : arg1) {
         DoOneClientStreamRequest curr_req = {};
         *curr_req.mutable_name() = this->name();
@@ -235,13 +234,12 @@ bool GizmoClient::do_one_client_stream(std::vector<std::string> arg1) {
 
 std::vector<bool> GizmoClient::do_one_server_stream(std::string arg1) {
     DoOneServerStreamRequest request;
-    grpc::ClientContext ctx;
-    set_client_ctx_authority(ctx);
+    ClientContext ctx;
 
     *request.mutable_name() = this->name();
     request.set_arg1(arg1);
 
-    auto reader(stub_->DoOneServerStream(&ctx, request));
+    auto reader(stub_->DoOneServerStream(ctx, request));
     DoOneServerStreamResponse curr_resp = {};
     std::vector<bool> rets = {};
     while (reader->Read(&curr_resp)) {
@@ -256,10 +254,9 @@ std::vector<bool> GizmoClient::do_one_server_stream(std::string arg1) {
 }
 
 std::vector<bool> GizmoClient::do_one_bidi_stream(std::vector<std::string> arg1) {
-    grpc::ClientContext ctx;
-    set_client_ctx_authority(ctx);
+    ClientContext ctx;
 
-    auto stream(stub_->DoOneBiDiStream(&ctx));
+    auto stream(stub_->DoOneBiDiStream(ctx));
     for (std::string arg : arg1) {
         DoOneBiDiStreamRequest curr_req = {};
         *curr_req.mutable_name() = this->name();
@@ -289,13 +286,12 @@ std::string GizmoClient::do_two(bool arg1) {
     DoTwoRequest request;
     DoTwoResponse response;
 
-    grpc::ClientContext ctx;
-    set_client_ctx_authority(ctx);
+    ClientContext ctx;
 
     *request.mutable_name() = this->name();
     request.set_arg1(arg1);
 
-    const grpc::Status status = stub_->DoTwo(&ctx, request, &response);
+    const grpc::Status status = stub_->DoTwo(ctx, request, &response);
     if (!status.ok()) {
         throw std::runtime_error(status.error_message());
     }
--- a/src/viam/sdk/common/utils.cpp
+++ b/src/viam/sdk/common/utils.cpp
@@ -132,8 +132,20 @@ bool operator==(const response_metadata& lhs, const response_metadata& rhs) {
     return lhs.captured_at == rhs.captured_at;
 }
 
-void set_client_ctx_authority(grpc::ClientContext& ctx) {
-    ctx.set_authority("viam-placeholder");
+void ClientContext::set_client_ctx_authority_() {
+    wrapped_context_.set_authority("viam-placeholder");
+}
+
+ClientContext::ClientContext() {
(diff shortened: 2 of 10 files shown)


dgottlieb (approved):

src/viam/sdk/common/utils.cpp
```diff
@@ -135,6 +135,12 @@ void set_client_ctx_authority(grpc::ClientContext& ctx) {
     ctx.set_authority("viam-placeholder");
 }
 
+std::unique_ptr<grpc::ClientContext> make_cc() {
```
acmorrow, line 138: I'd prefer this not say `cc` and be called `make_client_context`, or `client_context_factory`.

src/viam/examples/modules/complex/gizmo/api.cpp
```diff
@@ -206,13 +206,12 @@ bool GizmoClient::do_one(std::string arg1) {
...
 
-    grpc::ClientContext ctx;
-    set_client_ctx_authority(ctx);
+    auto ctx = make_cc();
 
     *request.mutable_name() = this->name();
     request.set_arg1(arg1);
 
-    const grpc::Status status = stub_->DoOne(&ctx, request, &response);
+    const grpc::Status status = stub_->DoOne(ctx.get(), request, &response);
```
acmorrow, line 214: If you wanted to avoid needing to write `.get` everywhere, you could make a wrapper type instead of a factory function and give that wrapper a conversion to `ClientContext`:

https://godbolt.org/

Another potential advantage of the wrapper type approach is that if you ever wanted it to start carrying state you would have a place to store that.

acmorrow (approved):
LGTM with a naming nit, and an optional suggestion about another way to accomplish this, if you are interested.

Meanwhile, I'm planning on doing some additional boilerplate reduction in #178 for the client side. It is possible that within that structure the need to do this will entirely be eliminated.

Can we hold this review for a bit until we see what that ends up looking like?

src/viam/sdk/common/utils.cpp
stuqdog: Done!

stuqdog:
> LGTM with a naming nit, and an optional suggestion about another way to accomplish this, if you are interested.
> 
> Meanwhile, I'm planning on doing some additional boilerplate reduction in #178 for the client side. It is possible that within that structure the need to do this will entirely be eliminated.
> 
> Can we hold this review for a bit until we see what that ends up looking like?

Sounds good! Renamed `make_cc`, will glance at the godbolt link for ideas and otherwise hold off on merging this until we see how #178  pans out.

maximpertsov (approved):

acmorrow:
I've merged #180, so I think this is unblocked. However, it might make sense to first convert the rest of the codebase over to `ClientHelper` and `ServiceHelper` under https://viam.atlassian.net/browse/RSDK-3556, as I believe that will mean that almost all of these changes boil away.

acmorrow:
@stuqdog - I think now that the client wrappers are in place it is time to revisit this review.

stuqdog:
@acmorrow yes, completely agree! Been a bit held up and gonna probably start over given the changes but will update this PR soon, I expect it'll be a very minor change at this point.

acmorrow:
> @acmorrow yes, completely agree! Been a bit held up and gonna probably start over given the changes but will update this PR soon, I expect it'll be a very minor change at this point.

Great. Revisiting this, I noticed that the `MLModelServiceClient` hadn't been updated to use `ClientHelper`, and I realized it was because that client does something tricky with `Arena`s. Which got me to thinking about if the `ClientHelper` ought to always be allocating the Request and Response objects out of an Arena too. It wouldn't immediately help allow `MLModelServiceClient` to use ClientHelper though, because it needs to take ownership of the `Arena` to pass back the tensors. Something to think about.

src/viam/sdk/services/motion/client.cpp
```diff
@@ -101,44 +101,30 @@ pose_in_frame MotionClient::get_pose(
...
-    *request.mutable_name() = this->name();
-    *request.mutable_extra() = map_to_struct(extra);
-
-    const grpc::Status status = stub_->StopPlan(&ctx, request, &response);
-    if (!status.ok()) {
-        throw std::runtime_error(status.error_message());
-    }
+    return make_client_helper(this, *stub_, &StubType::StopPlan)
+        .with(extra, [&](auto& request) { *request.mutable_component_name() = name.to_proto(); })
+        .invoke();
```
stuqdog, line 106: (flyby) we missed wrapping this with `client_helper` due to rebase issues, and also failed to set component_name previously. Both are now fixed.

src/viam/sdk/services/motion/client.cpp
```diff
@@ -101,44 +101,30 @@ pose_in_frame MotionClient::get_pose(
...
+                  request.set_last_plan_only(last_plan_only);
+                  if (execution_id) {
+                      *request.mutable_execution_id() = *execution_id;
+                  }
+              })
+        .invoke([](auto& response) {
+            return std::pair<Motion::plan_with_status, std::vector<Motion::plan_with_status>>(
+                Motion::plan_with_status::from_proto(response.current_plan_with_status()),
+                Motion::plan_with_status::from_proto(response.replan_history()));
+        });
```
stuqdog, line 127: (flyby) we missed wrapping this with client_helper due to rebase issues, and also failed to set component_name previously. Both are now fixed.

src/viam/sdk/services/motion/client.cpp
```diff
@@ -166,26 +152,16 @@ MotionClient::get_latest_plan_with_replan_history(const Name& name, const Attrib
...
+    return make_client_helper(this, *stub_, &StubType::ListPlanStatuses)
+        .with(extra, [&](auto& request) { request.set_only_active_plans(only_active_plans); })
+        .invoke([](auto& response) {
+            std::vector<Motion::plan_status_with_id> statuses;
+            for (const auto& proto : response.plan_statuses_with_ids()) {
+                statuses.push_back(Motion::plan_status_with_id::from_proto(proto));
+            }
+
+            return statuses;
+        });
```
stuqdog, line 164: (flyby) we missed wrapping this with client_helper due to rebase issues.

src/viam/sdk/services/motion/motion.hpp
```diff
@@ -420,7 +420,7 @@ class Motion : public Service {
     /// @param component_name The name of the component which the MoveOnGlobe request asked to move.
     /// @param execution_id The execution id of the requested plan.
     /// @return the plan and status of the requested execution's move the requested component
-    inline plan_with_status get_plan(Name component_name, const std::string& execution_id) {
+    inline plan_with_status get_plan(const Name& component_name, const std::string& execution_id) {
```
stuqdog, line 423: (flyby) const&

stuqdog:
@acmorrow updated post-`client_helper`, made use of wrapper type. I didn't touch the `MLModelServiceClient` code yet.

acmorrow (approved):
Looks great! A nice side effect of all of this is that it is now pretty easy to find all the places that don't use the client helper because they need to explicitly declare a `ClientContext`.

acmorrow:
>  I didn't touch the `MLModelServiceClient` code yet.

Yeah, I don't think you can convert that one yet. It does something tricky with Arenas to extend the lifetime of the reply data out to the caller. It may be beyond what is reasonable to expect a helper to cover. As with the `Robot` client, if the helper is covering the vast majority of cases, it is doing its job.

PR #180, opened 2023-11-15
RSDK-794 Client side helper to replace `stub_wrapper`

acmorrow:
Sorry this took a little longer than expected, I ended up trying several different designs before settling on this one. This is the client side analogue of #178.

@stuqdog and @benjirewis - For main review
CC @zaporter-work and @abe-winter for interest.

Thanks to @abe-winter for starting with `stub_wrapper` which pointed to deriving most of the needed info from the pointer-to-member on the stub type.

### DIFF
--- a/src/viam/sdk/common/client_helper.hpp
+++ b/src/viam/sdk/common/client_helper.hpp
@@ -7,6 +7,10 @@
 namespace viam {
 namespace sdk {
 
+namespace client_helper_details {
+[[noreturn]] void errorHandlerReturnedUnexpectedly(const ::grpc::Status&) noexcept;
+}  // namespace client_helper_details
+
 template <typename ClientType, typename StubType, typename RequestType, typename ResponseType>
 class ClientHelper {
     static void default_rsc_(RequestType&) {}
@@ -56,7 +60,7 @@ class ClientHelper {
         }
 
         std::forward<ErrorHandlerCallable>(ehc)(result);
-        std::abort();
+        client_helper_details::errorHandlerReturnedUnexpectedly(result);
     }
 
    private:
--- a/src/viam/sdk/common/client_helper.hpp
+++ b/src/viam/sdk/common/client_helper.hpp
@@ -1,0 +1,82 @@
+#pragma once
+
+#include <grpcpp/client_context.h>
+#include <viam/sdk/common/proto_type.hpp>
+#include <viam/sdk/common/utils.hpp>
+
+namespace viam {
+namespace sdk {
+
+template <typename ClientType, typename StubType, typename RequestType, typename ResponseType>
+class ClientHelper {
+    static void default_rsc_(RequestType&) {}
+    static void default_rhc_(const ResponseType&) {}
+    static void default_ehc_(const ::grpc::Status& status) {
+        throw std::runtime_error(status.error_message());
+    }
+
+   public:
+    using PFn = ::grpc::Status (StubType::*)(::grpc::ClientContext*,
+                                             const RequestType&,
+                                             ResponseType*);
+    explicit ClientHelper(ClientType* client, StubType* stub, PFn pfn)
+        : client_(client), stub_(stub), pfn_(pfn) {}
+
+    ClientHelper& with(const AttributeMap& extra) {
+        return with(extra, default_rsc_);
+    }
+
+    template <typename RequestSetupCallable>
+    ClientHelper& with(RequestSetupCallable&& rsc) {
+        std::forward<RequestSetupCallable>(rsc)(request_);
+        return *this;
+    }
+
+    template <typename RequestSetupCallable>
+    ClientHelper& with(const AttributeMap& extra, RequestSetupCallable&& rsc) {
+        *request_.mutable_extra() = map_to_struct(extra);
+        return with(std::forward<RequestSetupCallable>(rsc));
+    }
+
+    template <typename ResponseHandlerCallable = decltype(default_rhc_)>
+    auto invoke(ResponseHandlerCallable&& rhc = default_rhc_) {
+        return invoke(std::forward<ResponseHandlerCallable>(rhc), default_ehc_);
+    }
+
+    template <typename ResponseHandlerCallable, typename ErrorHandlerCallable>
+    auto invoke(ResponseHandlerCallable&& rhc, ErrorHandlerCallable&& ehc) {
+        *request_.mutable_name() = client_->name();
+        ::grpc::ClientContext ctx;
+        set_client_ctx_authority(ctx);
+
+        const auto result = (stub_->*pfn_)(&ctx, request_, &response_);
+        if (result.ok()) {
+            return std::forward<ResponseHandlerCallable>(rhc)(
+                const_cast<const ResponseType&>(response_));
+        }
+
+        std::forward<ErrorHandlerCallable>(ehc)(result);
+        std::abort();
+    }
+
+   private:
+    ClientType* client_;
+    StubType* stub_;
+    PFn pfn_;
+    RequestType request_;
+    ResponseType response_;
+};
+
+// TODO: Stop returns the status???
+
+template <typename ClientType, typename StubType, typename RequestType, typename ResponseType>
+auto make_client_helper(ClientType* client,
+                        StubType& stub,
+                        ::grpc::Status (StubType::*method)(::grpc::ClientContext*,
+                                                           const RequestType&,
+                                                           ResponseType*)) {
+    return ClientHelper<ClientType, StubType, RequestType, ResponseType>(client, &stub, method);
+}
+
+}  // namespace sdk
+}  // namespace viam
--- a/src/viam/sdk/common/client_helper.hpp
+++ b/src/viam/sdk/common/client_helper.hpp
@@ -67,8 +67,6 @@ class ClientHelper {
     ResponseType response_;
 };
 
-// TODO: Stop returns the status???
-
 template <typename ClientType, typename StubType, typename RequestType, typename ResponseType>
 auto make_client_helper(ClientType* client,
                         StubType& stub,
--- a/src/viam/sdk/components/motor/client.cpp
+++ b/src/viam/sdk/components/motor/client.cpp
@@ -22,108 +22,95 @@ MotorClient::MotorClient(std::string name, std::shared_ptr<grpc::Channel> channe
       stub_(viam::component::motor::v1::MotorService::NewStub(channel)),
       channel_(std::move(channel)){};
 
-void MotorClient::set_power(double power_pct, const AttributeMap& extra) {
-    viam::component::motor::v1::SetPowerRequest request;
-    viam::component::motor::v1::SetPowerResponse response;
-
-    grpc::ClientContext ctx;
-    set_client_ctx_authority(ctx);
-
-    *request.mutable_name() = this->name();
-    *request.mutable_extra() = map_to_struct(extra);
-    request.set_power_pct(power_pct);
-
-    const grpc::Status status = stub_->SetPower(&ctx, request, &response);
-    if (!status.ok()) {
+template <typename ClientType, typename StubType, typename RequestType, typename ResponseType>
+class ClientHelper {
+    static void default_rsc(ClientHelper&, RequestType&) {}
+    static void default_rhc(ClientHelper&, const ResponseType&) {}
+    static void default_ehc(ClientHelper&, const ::grpc::Status& status) {
         throw std::runtime_error(status.error_message());
     }
+
+   public:
+    using PFn = ::grpc::Status (StubType::*)(::grpc::ClientContext*,
+                                             const RequestType&,
+                                             ResponseType*);
+    explicit ClientHelper(ClientType* client, StubType& stub, PFn pfn)
+        : client_(client), stub_(stub), pfn_(pfn) {}
+
+    template <typename RequestSetupCallable = decltype(default_rsc),
+              typename ResponseHandlerCallable = decltype(default_rhc),
+              typename ErrorHandlerCallable = decltype(default_ehc)>
+    auto operator()(RequestSetupCallable&& rsc = default_rsc,
+                    ResponseHandlerCallable&& rhc = default_rhc,
+                    ErrorHandlerCallable&& ehc = default_ehc) {
+        RequestType request;
+        *request.mutable_name() = client_->name();
+        std::forward<RequestSetupCallable>(rsc)(*this, request);
+
+        ::grpc::ClientContext ctx;
+        set_client_ctx_authority(ctx);
+
+        ResponseType response;
+        const auto result = (stub_.*pfn_)(&ctx, request, &response);
+        if (!result.ok()) {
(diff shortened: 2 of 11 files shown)


src/viam/sdk/common/client_helper.hpp
```diff
@@ -0,0 +1,82 @@
...
+        set_client_ctx_authority(ctx);
+
+        const auto result = (stub_->*pfn_)(&ctx, request_, &response_);
+        if (result.ok()) {
+            return std::forward<ResponseHandlerCallable>(rhc)(
+                const_cast<const ResponseType&>(response_));
+        }
+
+        std::forward<ErrorHandlerCallable>(ehc)(result);
+        std::abort();
```
acmorrow, line 59: The `ErrorHandlerCallable` is supposed to throw. But, it might not!

In that case we could either terminate, as I'm doing now, or we could call `default_ehc_` for you. I chose the former because the latter felt like papering over a programmer error.

Definitely interested in others thoughts here.

src/viam/sdk/common/client_helper.hpp
```diff
@@ -0,0 +1,82 @@
...
+    template <typename ResponseHandlerCallable = decltype(default_rhc_)>
+    auto invoke(ResponseHandlerCallable&& rhc = default_rhc_) {
+        return invoke(std::forward<ResponseHandlerCallable>(rhc), default_ehc_);
+    }
+
+    template <typename ResponseHandlerCallable, typename ErrorHandlerCallable>
+    auto invoke(ResponseHandlerCallable&& rhc, ErrorHandlerCallable&& ehc) {
+        *request_.mutable_name() = client_->name();
+        ::grpc::ClientContext ctx;
+        set_client_ctx_authority(ctx);
```
acmorrow, line 50: Note that I don't think this obviates the need for #179 of some form. This class shouldn't be responsible for properly configuring a client context. But it does mean that if we convert the codebase over to `ClientHelper`, there will be fewer places where the factory or wrapper class needs to be injected.

src/viam/sdk/common/client_helper.hpp
```diff
@@ -0,0 +1,82 @@
...
+    PFn pfn_;
+    RequestType request_;
+    ResponseType response_;
+};
+
+// TODO: Stop returns the status???
+
+template <typename ClientType, typename StubType, typename RequestType, typename ResponseType>
+auto make_client_helper(ClientType* client,
+                        StubType& stub,
```
acmorrow, line 74: The choice of reference over pointer here is because the Stub is usually held by `unique_ptr`, and it felt better to write `*stub_` than `stub.get()` everywhere. But it could be easily changed if that's too quirky.

src/viam/sdk/components/motor/client.cpp
```diff
@@ -141,66 +81,27 @@ grpc::StatusCode MotorClient::stop(const AttributeMap& extra) {
...
-
-    *request.mutable_name() = this->name();
-    *request.mutable_extra() = map_to_struct(extra);
-
-    const grpc::Status status = stub_->IsPowered(&ctx, request, &response);
-    if (!status.ok()) {
-        throw std::runtime_error(status.error_message());
-    }
-    return from_proto(response);
+    return make_client_helper(this, *stub_, &StubType::IsPowered)
```
acmorrow, line 84: ⬆️  - I did not convert `MotorClient::stop` because the behavior is non-standard. As written the `ClientHelper` cannot deal with it, because it splits the status into the good case where you handle a response and a bad case where you handle an error. It may actually be possible to SFINAE in the ability to have the invoke handler take a callback that takes a `grpc::Status` object, which would let us use the helper here.

src/viam/sdk/components/motor/client.cpp
```diff
@@ -141,66 +81,27 @@ grpc::StatusCode MotorClient::stop(const AttributeMap& extra) {
...
-    set_client_ctx_authority(ctx);
-
-    *request.mutable_name() = this->name();
-
-    const grpc::Status status = stub_->IsMoving(&ctx, request, &response);
-    if (!status.ok()) {
-        throw std::runtime_error(status.error_message());
-    }
-    return response.is_moving();
+    return make_client_helper(this, *stub_, &StubType::IsMoving).invoke([](auto& response) {
```
acmorrow, line 96: Note the lack of a `.with` call here: there are no params and no `extra` so we don't have anything to prepare.

src/viam/sdk/common/client_helper.hpp
```diff
@@ -0,0 +1,82 @@
...
+    }
+
+    template <typename ResponseHandlerCallable = decltype(default_rhc_)>
+    auto invoke(ResponseHandlerCallable&& rhc = default_rhc_) {
+        return invoke(std::forward<ResponseHandlerCallable>(rhc), default_ehc_);
+    }
+
+    template <typename ResponseHandlerCallable, typename ErrorHandlerCallable>
+    auto invoke(ResponseHandlerCallable&& rhc, ErrorHandlerCallable&& ehc) {
+        *request_.mutable_name() = client_->name();
```
acmorrow, line 48: Application of `name` is done here so that you are not required to call `with` but can instead directly call `invoke` if you have nothing to configure on the request.

src/viam/sdk/components/movement_sensor/client.cpp
```diff
@@ -26,90 +27,71 @@ MovementSensorClient::MovementSensorClient(std::string name, std::shared_ptr<grp
 using namespace viam::component::movementsensor::v1;
 
 Vector3 MovementSensorClient::get_linear_velocity(const AttributeMap& extra) {
-    GetLinearVelocityResponse response;
-    stub_wrapper<GetLinearVelocityRequest>(
```
acmorrow, line 30: One of the big advantages of `ClientHelper` over `stub_wrapper` is that you only need to know the function on the stub, and you don't need to name the request and response types explicitly (though you can, if you want, on the lambda).

src/viam/sdk/common/client_helper.hpp
```diff
@@ -0,0 +1,82 @@
...
+    static void default_rhc_(const ResponseType&) {}
+    static void default_ehc_(const ::grpc::Status& status) {
+        throw std::runtime_error(status.error_message());
+    }
+
+   public:
+    using PFn = ::grpc::Status (StubType::*)(::grpc::ClientContext*,
+                                             const RequestType&,
+                                             ResponseType*);
+    explicit ClientHelper(ClientType* client, StubType* stub, PFn pfn)
```
benjirewis, line 22: Why `explicit`?

src/viam/sdk/common/client_helper.hpp
benjirewis: > This class shouldn't be responsible for properly configuring a client context.

If we do convert the whole codebase to `ClientHelper` (including the complex module example), why would we need to configure a client context anywhere else?

src/viam/sdk/common/client_helper.hpp
benjirewis: What if an invocation of the `ClientHelper` had an `ehc` that swallowed an error and logged it or something? Should we just return some nullish value here instead of aborting? To be clear, I agree that we shouldn't call `default_ehc_` in the absence of a throw, but I'm not convinced we should `abort` either.

src/viam/sdk/common/client_helper.hpp
```diff
@@ -0,0 +1,82 @@
...
+template <typename ClientType, typename StubType, typename RequestType, typename ResponseType>
+class ClientHelper {
+    static void default_rsc_(RequestType&) {}
+    static void default_rhc_(const ResponseType&) {}
+    static void default_ehc_(const ::grpc::Status& status) {
+        throw std::runtime_error(status.error_message());
+    }
+
+   public:
+    using PFn = ::grpc::Status (StubType::*)(::grpc::ClientContext*,
```
benjirewis, line 19: [q] I got the other abbreviations, but what does `PFn` stand for?

src/viam/sdk/common/client_helper.hpp
benjirewis: No strong opinion; I don't view `*stub_` as any less strange than `stub.get()`, but references seem more ergonomic than pointers in my cursory C++ knowledge.

src/viam/sdk/common/client_helper.hpp
```diff
@@ -0,0 +1,82 @@
...
+
+   private:
+    ClientType* client_;
+    StubType* stub_;
+    PFn pfn_;
+    RequestType request_;
+    ResponseType response_;
+};
+
+// TODO: Stop returns the status???
```
benjirewis, line 70: Did we file a ticket for changing `stop` return types? Happy to do so if one hasn't been made.

src/viam/sdk/components/motor/client.cpp
benjirewis: Yeah I would rather change the `stop` return type than SFINAE our way around it if possible.

src/viam/sdk/components/movement_sensor/client.cpp
```diff
@@ -26,90 +27,71 @@ MovementSensorClient::MovementSensorClient(std::string name, std::shared_ptr<grp
...
-        this, response, extra, &MovementSensorClient::Stub::GetLinearVelocity);
-    return Vector3::from_proto(response.linear_velocity());
+    return make_client_helper(this, *stub_, &MovementSensorClient::Stub::GetLinearVelocity)
+        .with(extra)
+        .invoke([](auto& response) { return Vector3::from_proto(response.linear_velocity()); });
```
benjirewis, line 32:
```suggestion
        .invoke([&](auto& response) { return Vector3::from_proto(response.linear_velocity()); });
```
Why are these `[=]` lambdas when the ones on the motor are `[&]`?

src/viam/sdk/common/client_helper.hpp
```diff
@@ -0,0 +1,82 @@
...
+    }
+
+    template <typename RequestSetupCallable>
+    ClientHelper& with(const AttributeMap& extra, RequestSetupCallable&& rsc) {
+        *request_.mutable_extra() = map_to_struct(extra);
+        return with(std::forward<RequestSetupCallable>(rsc));
+    }
+
+    template <typename ResponseHandlerCallable = decltype(default_rhc_)>
+    auto invoke(ResponseHandlerCallable&& rhc = default_rhc_) {
```
benjirewis, line 42: Yeah FWIW I appreciate a named `invoke` method over overloading `operator()` because an `invoke` with an rhc and an ehc may be pretty tough to read. The `ServiceHelper` _does_ do the opposite, but it kind of makes sense that the client is the one "invoking" the gRPC endpoint.

benjirewis (commented):
Awesome; very cool code. Thanks for this, just have some qs!

src/viam/sdk/common/client_helper.hpp
acmorrow: @stuqdog ?

src/viam/sdk/components/movement_sensor/client.cpp
acmorrow: Because the response handler doesn't need access to any captured state, so we don't need `[&]`.  We need it on the `with` lambdas because we need access to the paramers to the function. I don't think motor does it any differently, and I'm pretty sure there aren't any `[=]` lambdas at all.

src/viam/sdk/common/client_helper.hpp
acmorrow: Unfortunately, I don't think that will work out. If we change this code to `return std::forward<ErrorHandlerCallable(ehc)>(result)`, then *every* ehc will need to have the same return type as `rhc` or this function won't compile. Which would mean we couldn't have a default for `ehc` since every `invoke` handler has its own return type.

So, we can't return, so we must throw, and the `ehc` didn't do that, so either we throw or we abort.

src/viam/sdk/common/client_helper.hpp
benjirewis: Ah gotcha; no that seems fine. Could do `Method`, but I don't have strong opinion.

src/viam/sdk/common/client_helper.hpp
benjirewis: Sounds good, and thanks for context; just didn't see one on the `ServiceHelper` constructor.

src/viam/sdk/common/client_helper.hpp
benjirewis: Cool that makes sense; thanks!

src/viam/sdk/common/client_helper.hpp
benjirewis: Ah gotcha. I was thinking about something like

```
std::forward<ErrorHandlerCallable>(ehc)(result);
return ["empty"-value-of-whatever-rhc-returns];
```

but yeah seems like that's not possible given there's no template type for the rhc return value. I can just recall MongoDB Go driver methods that would swallow certain errors from the server, and it seems like using `ClientHelper` means if a non-ok gRPC status is returned, some error _must_ be thrown or the program will abort. If we _really_ wanted to swallow an error though, we could just not use `ClientHelper`.

src/viam/sdk/components/movement_sensor/client.cpp
benjirewis: Ah sounds good; I was seeing `[&]` captures in motor in `with` functions.

> and I'm pretty sure there aren't any [=] lambdas at all.

Oh I thought `[]` implicitly meant `[=]`.

benjirewis (approved):
LGTM; thanks for handling these hard pieces!

src/viam/sdk/common/client_helper.hpp
stuqdog: yep! https://viam.atlassian.net/browse/RSDK-5760

src/viam/sdk/common/client_helper.hpp
stuqdog: Given that we can't reasonably return anything here, I agree that `abort` makes the most sense. I wonder if there's some value in logging the error message ourselves, just to provide more info to the user?

src/viam/sdk/common/client_helper.hpp
stuqdog: seems reasonable to me

stuqdog (approved):
looks great to me, thanks!

src/viam/sdk/common/client_helper.hpp
acmorrow: Thanks! I'll remove the TODO (which shouldn't have moved to here when I pulled this code out to a header anyway).

src/viam/sdk/common/client_helper.hpp
acmorrow: Sure, that's easy enough.

acmorrow:
@stuqdog - I removed the stale `TODO` comment, added logging for misbehaving `ehc`'s, and went ahead and published the headers for both `ClientHelper` and `ServiceHelper`. I didn't see any reason to hold them back, but I can easily pull that back out if you would like to wait.

stuqdog (approved):
looks great to me, thanks again!

PR #181, opened 2023-11-17
RSDK-5233 - implement missing methods

stuqdog:
#### Major Changes
Adds missing `board` and `motion` methods.

#### Minor changes
Type aliases our commonly used  `time_point`
Fixes an issues with proto conversion of `ResourceName`
Ensures return value of `MockMotion::do_command` is reliable
Ensures propagation of a couple default arguments missing in `client` code

### DIFF
--- a/src/viam/sdk/common/utils.cpp
+++ b/src/viam/sdk/common/utils.cpp
@@ -58,16 +58,14 @@ std::string bytes_to_string(const std::vector<unsigned char>& b) {
     return img_string;
 };
 
-std::chrono::time_point<long long, std::chrono::nanoseconds> timestamp_to_time_pt(
-    const google::protobuf::Timestamp& timestamp) {
+time_pt timestamp_to_time_pt(const google::protobuf::Timestamp& timestamp) {
     const std::chrono::seconds seconds(timestamp.seconds());
     const std::chrono::nanoseconds nanos(timestamp.nanos());
-    return std::chrono::time_point<long long, std::chrono::nanoseconds>(
-        std::chrono::duration_cast<std::chrono::system_clock::duration>(seconds) + nanos);
+    return time_pt(std::chrono::duration_cast<std::chrono::system_clock::duration>(seconds) +
+                   nanos);
 }
 
-google::protobuf::Timestamp time_pt_to_timestamp(
-    const std::chrono::time_point<long long, std::chrono::nanoseconds>& time_pt) {
+google::protobuf::Timestamp time_pt_to_timestamp(const time_pt& time_pt) {
     const std::chrono::seconds duration_s =
         std::chrono::duration_cast<std::chrono::seconds>(time_pt.time_since_epoch());
     const std::chrono::nanoseconds duration_ns = time_pt.time_since_epoch() - duration_s;
--- a/src/viam/sdk/common/utils.hpp
+++ b/src/viam/sdk/common/utils.hpp
@@ -23,6 +23,8 @@ const std::string kService = "service";
 const std::string kRDK = "rdk";
 const std::string kBuiltin = "builtin";
 
+using time_pt = std::chrono::time_point<long long, std::chrono::nanoseconds>;
+
 std::vector<viam::common::v1::ResourceName> resource_names_for_resource(
     const std::shared_ptr<Resource>& resource);
 
@@ -43,7 +45,7 @@ class ResourceNameEqual {
 };
 
 struct response_metadata {
-    std::chrono::time_point<long long, std::chrono::nanoseconds> captured_at;
+    time_pt captured_at;
 
     static response_metadata from_proto(const viam::common::v1::ResponseMetadata& proto);
     static viam::common::v1::ResponseMetadata to_proto(const response_metadata& metadata);
@@ -51,15 +53,11 @@ struct response_metadata {
 
 bool operator==(const response_metadata& lhs, const response_metadata& rhs);
 
-/// @brief convert a google::protobuf::Timestamp to std::chrono::time_point<long long,
-/// std::chrono::nanoseconds>.
-std::chrono::time_point<long long, std::chrono::nanoseconds> timestamp_to_time_pt(
-    const google::protobuf::Timestamp& timestamp);
+/// @brief convert a google::protobuf::Timestamp to time_pt
+time_pt timestamp_to_time_pt(const google::protobuf::Timestamp& timestamp);
 
-/// @brief convert a std::chrono::time_point<long long, std::chrono::nanoseconds> to a
-/// google::protobuf::Timestamp.
-google::protobuf::Timestamp time_pt_to_timestamp(
-    const std::chrono::time_point<long long, std::chrono::nanoseconds>& time_pt);
+/// @brief convert a time_pt to a google::protobuf::Timestamp.
+google::protobuf::Timestamp time_pt_to_timestamp(const time_pt& time_pt);
 
 std::vector<unsigned char> string_to_bytes(std::string const& s);
 std::string bytes_to_string(std::vector<unsigned char> const& b);
--- a/src/viam/sdk/components/board/board.hpp
+++ b/src/viam/sdk/components/board/board.hpp
@@ -213,8 +213,21 @@ class Board : public Component {
     virtual analog_value read_analog(const std::string& analog_reader_name,
                                      const AttributeMap& extra) = 0;
 
-    /// @brief Returns the current value of the interrupt which is based on the type of interrupt.
-    /// Consult Viam's `Board` docs for more information.
+    /// @brief Writes the value to the analog writer of the board.
+    /// @param pin the pin to write to
+    /// @param value the value to set the pin to
+    inline void write_analog(const std::string& pin, int value) {
+        return write_analog(pin, value, {});
+    }
+
+    /// @brief Writes the value to the analog writer of the board.
+    /// @param pin the pin to write to
+    /// @param value the value to set the pin to
+    /// @param extra any additional arguments to the method
+    virtual void write_analog(const std::string& pin, int value, const AttributeMap& extra) = 0;
+
+    /// @brief Returns the current value of the interrupt which is based on the type of
+    /// interrupt. Consult Viam's `Board` docs for more information.
     /// @param digital_interrupt_name digital interrupt to check
     inline digital_value read_digital_interrupt(const std::string& digital_interrupt_name) {
         return read_digital_interrupt(digital_interrupt_name, {});
--- a/src/viam/sdk/components/encoder/client.hpp
+++ b/src/viam/sdk/components/encoder/client.hpp
@@ -21,7 +21,8 @@ namespace sdk {
 class EncoderClient : public Encoder {
    public:
     EncoderClient(std::string name, std::shared_ptr<grpc::Channel> channel);
-    position get_position(const AttributeMap& extra, position_type position_type) override;
+    position get_position(const AttributeMap& extra,
+                          position_type position_type = position_type::unspecified) override;
     void reset_position(const AttributeMap& extra) override;
     properties get_properties(const AttributeMap& extra) override;
     std::vector<GeometryConfig> get_geometries(const AttributeMap& extra) override;
--- a/src/viam/sdk/resource/resource_api.cpp
+++ b/src/viam/sdk/resource/resource_api.cpp
@@ -97,7 +97,7 @@ const std::string& Name::remote_name() const {
 }
 
 std::string Name::to_string() const {
-    if (remote_name_.empty()) {
+    if (remote_name_.empty() || remote_name_ == ":") {
         return api_.to_string() + "/" + name_;
     }
     return api_.to_string() + "/" + remote_name_ + ":" + name_;
--- a/src/viam/sdk/services/motion/client.cpp
+++ b/src/viam/sdk/services/motion/client.cpp
@@ -135,6 +135,71 @@ pose_in_frame MotionClient::get_pose(
     return pose_in_frame::from_proto(response.pose());
 }
 
+void MotionClient::stop_plan(const Name& name, const AttributeMap& extra) {
+    service::motion::v1::StopPlanRequest request;
+    service::motion::v1::StopPlanResponse response;
+    grpc::ClientContext ctx;
+    set_client_ctx_authority(ctx);
+
+    *request.mutable_name() = this->name();
+    *request.mutable_extra() = map_to_struct(std::move(extra));
+
+    const grpc::Status status = stub_->StopPlan(&ctx, request, &response);
+    if (!status.ok()) {
+        throw std::runtime_error(status.error_message());
+    }
+}
+
+std::pair<Motion::plan_with_status, std::vector<Motion::plan_with_status>> MotionClient::get_plan(
+    const Name& name,
+    const AttributeMap& extra,
+    bool last_plan_only,
+    boost::optional<std::string> execution_id) {
+    service::motion::v1::GetPlanRequest request;
+    service::motion::v1::GetPlanResponse response;
+    grpc::ClientContext ctx;
+    set_client_ctx_authority(ctx);
+
+    *request.mutable_name() = this->name();
+    *request.mutable_extra() = map_to_struct(std::move(extra));
+    request.set_last_plan_only(last_plan_only);
+    if (execution_id) {
+        *request.mutable_execution_id() = *execution_id;
+    }
+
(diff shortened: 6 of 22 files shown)


src/viam/sdk/services/motion/motion.hpp
```diff
@@ -247,6 +366,63 @@ class Motion : public Service {
...
+    /// @param component_name The name of the component which the MoveOnGlobe request asked to move.
+    /// @param last_plan_only If true, causes only the last plan for the component/execution to be
+    /// returned.
+    /// @param execution_id Optional execution id of a previous plan, if you want to know about it.
+    /// @param extra Any additional arguments to the method.
+    virtual std::pair<plan_with_status, std::vector<plan_with_status>> get_plan(
+        const Name& component_name,
+        const AttributeMap& extra,
+        bool last_plan_only = false,
+        boost::optional<std::string> execution_id = {}) = 0;
```
stuqdog, line 406: Note to reviewers: this is a little awkward. `extra` is treated as optional in all methods (hence the `inline` versions) but here we have two other optional methods! I didn't want to require people to pass `execution_id` and/or `last_plan_only` at all times, but the ordering of multiple optional arguments can get a little tricky. Curious if anyone has thoughts on a better API here.

cc @acmorrow who I know has thoughts on defaults args in C++.

src/viam/sdk/resource/resource_api.cpp
```diff
@@ -97,7 +97,7 @@ const std::string& Name::remote_name() const {
 }
 
 std::string Name::to_string() const {
-    if (remote_name_.empty()) {
+    if (remote_name_.empty() || remote_name_ == ":") {
```
dgottlieb, line 100: Intentional?

src/viam/sdk/common/utils.cpp
```diff
@@ -58,16 +58,14 @@ std::string bytes_to_string(const std::vector<unsigned char>& b) {
...
 };
 
-std::chrono::time_point<long long, std::chrono::nanoseconds> timestamp_to_time_pt(
-    const google::protobuf::Timestamp& timestamp) {
+time_pt timestamp_to_time_pt(const google::protobuf::Timestamp& timestamp) {
     const std::chrono::seconds seconds(timestamp.seconds());
     const std::chrono::nanoseconds nanos(timestamp.nanos());
-    return std::chrono::time_point<long long, std::chrono::nanoseconds>(
-        std::chrono::duration_cast<std::chrono::system_clock::duration>(seconds) + nanos);
+    return time_pt(std::chrono::duration_cast<std::chrono::system_clock::duration>(seconds) +
```
dgottlieb, line 64: Not super familiar with the std::chrono stuff. I'm guessing a time_pt takes in a 64-bit number in the unit of nanoseconds?

src/viam/sdk/components/encoder/client.hpp
```diff
@@ -21,7 +21,8 @@ namespace sdk {
 class EncoderClient : public Encoder {
    public:
     EncoderClient(std::string name, std::shared_ptr<grpc::Channel> channel);
-    position get_position(const AttributeMap& extra, position_type position_type) override;
+    position get_position(const AttributeMap& extra,
+                          position_type position_type = position_type::unspecified) override;
```
dgottlieb, line 25: I assume this is intentional?

src/viam/sdk/services/motion/client.cpp
```diff
@@ -135,6 +135,71 @@ pose_in_frame MotionClient::get_pose(
     return pose_in_frame::from_proto(response.pose());
 }
 
+void MotionClient::stop_plan(const Name& name, const AttributeMap& extra) {
+    service::motion::v1::StopPlanRequest request;
+    service::motion::v1::StopPlanResponse response;
+    grpc::ClientContext ctx;
+    set_client_ctx_authority(ctx);
+
+    *request.mutable_name() = this->name();
```
dgottlieb, line 144: This weirds me out. It's clearly the existing pattern on how to set things on requests -- so no objections. Out of curiosity, was this a choice we made of is this just how proto creates C++ objects?

src/viam/sdk/services/motion/client.cpp
```diff
@@ -135,6 +135,71 @@ pose_in_frame MotionClient::get_pose(
...
+    *request.mutable_extra() = map_to_struct(std::move(extra));
```
dgottlieb, line 145: A [different line in this patch](https://github.com/viamrobotics/viam-cpp-sdk/pull/181/files#diff-99d8513b064c421efbffb53e9b3d68e08962e998f22ceb61037f9c8fd9a0d766R200) did not `std::move`.

I think* a `std::move(const reference)` just turns into a copy, so this seems safe to me. Just not useful.

src/viam/sdk/services/motion/client.cpp
```diff
@@ -135,6 +135,71 @@ pose_in_frame MotionClient::get_pose(
...
+    const AttributeMap& extra,
+    bool last_plan_only,
+    boost::optional<std::string> execution_id) {
+    service::motion::v1::GetPlanRequest request;
+    service::motion::v1::GetPlanResponse response;
+    grpc::ClientContext ctx;
+    set_client_ctx_authority(ctx);
+
+    *request.mutable_name() = this->name();
+    *request.mutable_extra() = map_to_struct(std::move(extra));
```
dgottlieb, line 164: Similar `std::move(const reference)` curiosity.

src/viam/sdk/services/motion/server.cpp
```diff
@@ -178,6 +178,96 @@ ::grpc::Status MotionServer::GetPose(::grpc::ServerContext* context,
     return ::grpc::Status();
 };
 
+::grpc::Status MotionServer::GetPlan(::grpc::ServerContext* context,
+                                     const ::viam::service::motion::v1::GetPlanRequest* request,
```
dgottlieb, line 182: I was expecting the request here to be a const reference like the grpc stub API. Once again, this is clearly the existing pattern, but curious if you're aware of any motivation here or if it's just incidental.

dgottlieb (commented):
Only specific request is clarity on the `std::move(const reference)` usages.

src/viam/sdk/common/utils.cpp
stuqdog: Yep! It's just a `typedef` to avoid having to write that out a bunch, details in `utils.hpp`.

src/viam/sdk/components/encoder/client.hpp
stuqdog: Yeah, we define the default behavior in the base `encoder` class but I found that the defaults didn't propagate unless specifically re-set.

src/viam/sdk/services/motion/client.cpp
stuqdog: Yeah it's kinda weird, some fields have both a `set_foo()` method and a `mutable_foo()` method, but some only had the `mutable_foo()` method. So early on we used the `mutable_` variant more often and it just kinda stuck. It's not a super principled stance and I would not oppose taking time at some point to go through and clean these up a bit.

src/viam/sdk/services/motion/client.cpp
stuqdog: Thanks for the catch! will remove the superfluous moves.

src/viam/sdk/services/motion/server.cpp
stuqdog: This is the API for the proto `MotionService::Service` that we're inheriting from so I think we're committed to it. I think this is a difference between the [server](https://github.com/viamrobotics/viam-cpp-sdk/blob/e1b1e4801e2d916918cf40e414ec32b6e743a450/src/viam/api/service/motion/v1/motion.grpc.pb.h#L333) and [client](https://github.com/viamrobotics/viam-cpp-sdk/blob/e1b1e4801e2d916918cf40e414ec32b6e743a450/src/viam/api/service/motion/v1/motion.grpc.pb.h#L105) code.

dgottlieb (approved):
I appreciate you taking the time to respond to my curiosities!

src/viam/sdk/services/motion/motion.hpp
acmorrow: Yeah, I'm finding the interaction of the virtual methods, the default parameters, and the mandatory handling for `extra` is starting to look pretty bad. I have a few suggestions.

I think the way forward is to admit that there needs to be more than one non-virtual entry point, and then have those delegate to the virtual method, potentially as private per the template method pattern I mentioned before. For instance, rather than having a defaulted bool parameter (more on that later) for whether you only want the last plan, make multiple methods:

`get_plans`
`get_last_plan`
`get_plans_by_id`
`get_last_plan_by_id`

Then you don't need defaults at all, since the signatures can differ (as can the return types!). And then it is easy to tack the optional `extra` on the end. All of these functions would just be wrappers around `get_plan` which would always require all of the parameters.

Meanwhile, regarding defaulted bool parameters, please try to avoid it, because at the call site it becomes very hard to know what it means when specified.

src/viam/sdk/services/motion/motion.hpp
```diff
@@ -87,6 +90,122 @@ struct motion_configuration {
 /// specific motion implementations. This class cannot be used on its own.
 class Motion : public Service {
    public:
+    /// @enum PlanState
+    /// @brief Describes the possible states a plan can be in.
+    /// @ingroup Motion
+    enum class PlanState : uint8_t {
+        in_progress = 0,
+        stopped = 1,
```
acmorrow, line 98: Are the numeric values important, or could you leave them off? Also, enum values are constants so consider adorning them as such.

src/viam/sdk/services/motion/motion.hpp
```diff
@@ -87,6 +90,122 @@ struct motion_configuration {
...
+        failed = 3,
+    };
+
+    static PlanState from_proto(const service::motion::v1::PlanState& proto);
+    static service::motion::v1::PlanState to_proto(const PlanState& state);
+
+    /// @struct plan_status
+    /// @brief Describes the state of a given plan at a point in time.
+    /// @ingroup Motion
+    struct plan_status {
```
acmorrow, line 109: Is there a consistent story for what types get PacalCase and which get snake_case? Because the enum above is PascalCase but this struct is snake.

src/viam/sdk/services/motion/motion.hpp
```diff
@@ -247,6 +366,63 @@ class Motion : public Service {
...
+        const AttributeMap& extra,
+        bool last_plan_only = false,
+        boost::optional<std::string> execution_id = {}) = 0;
+
+    /// @brief Returns the status of plans created by MoveOnGlobe requests.
+    /// Includes statuses of plans that are executing, or are part of an executing that changed
+    /// its state within the last 24 hours.
+    /// @param only_active_plans If true, filters returned plans to only those currently active.
+    /// Defaults to false.
+    inline std::vector<plan_status_with_id> list_plan_statuses(bool only_active_plans = false) {
```
acmorrow, line 413: Ditto here: I'd make this two methods: `list_plan_statusus` and `list_active_plan_statuses`.

src/viam/sdk/common/utils.hpp
```diff
@@ -23,6 +23,8 @@ const std::string kService = "service";
 const std::string kRDK = "rdk";
 const std::string kBuiltin = "builtin";
 
+using time_pt = std::chrono::time_point<long long, std::chrono::nanoseconds>;
```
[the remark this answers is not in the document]
stuqdog: Interesting. I'm trying to recall how we ended up with `long long` but I'm not finding it. I'll look into changing this to be a `clock` type and confirm that everything still runs as expected.

As for the typedef, can you elaborate a bit? I agree it doesn't save all that much (just some extra typing here or there) but are there costs or other reasons not to typedef that I'm not aware of?

src/viam/sdk/components/board/board.hpp
```diff
@@ -213,8 +213,21 @@ class Board : public Component {
...
+    /// @param value the value to set the pin to
+    inline void write_analog(const std::string& pin, int value) {
+        return write_analog(pin, value, {});
+    }
+
+    /// @brief Writes the value to the analog writer of the board.
+    /// @param pin the pin to write to
+    /// @param value the value to set the pin to
+    /// @param extra any additional arguments to the method
+    virtual void write_analog(const std::string& pin, int value, const AttributeMap& extra) = 0;
```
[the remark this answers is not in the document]
stuqdog: I'm not sure what you mean exactly. Do you mean an inline version with no `extra` parameter that passes an empty `AttributeMap`? That's defined directly above.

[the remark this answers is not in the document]
acmorrow: Oops I have no idea how I didn't see that! Resolving.

src/viam/sdk/common/utils.hpp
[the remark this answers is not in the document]
acmorrow: My thoughts on the typedef are individually minor but sort of add up to not wanting to do it:
- Every typedef has a cost when you need to figure out what it is.
- If we are standardizing on `std::chrono::time_point<X, Y>` as our standard time point type, I'd rather we just typedef it to`time_point` rather than `time_pt`.
- There is supposed to be a clock in there, and I think sorting that out should take precedence, since it may alter how the type gets used and undo some of the advantages of the typedef.

src/viam/sdk/services/motion/motion.hpp
stuqdog: definitely don't need the numbers! will drop.

src/viam/sdk/services/motion/motion.hpp
stuqdog: Just a brain fart, switching the enum to snake_case for consistency.

src/viam/sdk/common/utils.hpp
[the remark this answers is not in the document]
stuqdog: cool makes sense! I went with a typedef in `utils.cpp` just to avoid having to type out the full type every time, but in headers and call sites reverted to the actual full type.

Also, tests seem to be passing fine when replacing `long long` with `steady_clock` so I'm going ahead and making that change.

src/viam/sdk/components/encoder/client.hpp
stuqdog: I like your proposed change but I don't want to get this too bogged down right now. Reverting this `encoder` change, creating another ticket to switch defaults to using the `template method`, and have removed defaults from all the new methods in this PR.

stuqdog:
@acmorrow fyi a couple small changes are still TK (making clang tidy happy mostly) but didn't want to block too much on review.

src/viam/sdk/common/utils.hpp
```diff
@@ -43,23 +43,23 @@ class ResourceNameEqual {
 };
 
 struct response_metadata {
-    std::chrono::time_point<long long, std::chrono::nanoseconds> captured_at;
+    std::chrono::time_point<std::chrono::steady_clock, std::chrono::nanoseconds> captured_at;
```
acmorrow, line 46: Why `steady_clock`?

src/viam/sdk/common/utils.cpp
```diff
@@ -58,16 +59,14 @@ std::string bytes_to_string(const std::vector<unsigned char>& b) {
...
 };
 
-std::chrono::time_point<long long, std::chrono::nanoseconds> timestamp_to_time_pt(
-    const google::protobuf::Timestamp& timestamp) {
+time_point timestamp_to_time_pt(const google::protobuf::Timestamp& timestamp) {
     const std::chrono::seconds seconds(timestamp.seconds());
     const std::chrono::nanoseconds nanos(timestamp.nanos());
-    return std::chrono::time_point<long long, std::chrono::nanoseconds>(
-        std::chrono::duration_cast<std::chrono::system_clock::duration>(seconds) + nanos);
+    return time_point(std::chrono::duration_cast<std::chrono::system_clock::duration>(seconds) +
```
acmorrow, line 65: Elsewhere `steady_clock`, is `system_clock` here intentional?

acmorrow (approved):
The bulk of this LGTM but I still have some questions about the time stuff. Could it be separated out into another review to unblock the rest of the work in this review?

src/viam/sdk/common/utils.cpp
stuqdog: Sounds good! Will revert to `long long`, open a new ticket to revisit, and then merge.

cheukt:
just gonna remove myself since I don't have much context/knowledge here

PR #182, opened 2023-11-20
RSDK-5760 - change stop return to void

stuqdog:
#### Major changes

Change return type of `stop()` from `grpc::StatusCode` to `void`.

Most changes are pretty straightforward, there are two places where a bit more attention should be necessary. Will highlight those to reviewers.

### DIFF
--- a/src/viam/examples/modules/complex/base/impl.cpp
+++ b/src/viam/examples/modules/complex/base/impl.cpp
@@ -1,5 +1,6 @@
 #include "impl.hpp"
 
+#include <exception>
 #include <fstream>
 #include <iostream>
 #include <sstream>
@@ -63,18 +64,26 @@ bool MyBase::is_moving() {
     return left_->is_moving() || right_->is_moving();
 }
 
-grpc::StatusCode MyBase::stop(const AttributeMap& extra) {
-    auto left_stop = left_->stop(extra);
-    auto right_stop = right_->stop(extra);
+void MyBase::stop(const AttributeMap& extra) {
+    std::exception* e;
 
-    // Return first of any non-ok error code received from motors.
-    if (left_stop != grpc::StatusCode::OK) {
-        return left_stop;
+    // make sure we try to stop both motors, even if the first fails.
+    try {
+        left_->stop(extra);
+    } catch (const std::exception& err) {
+        *e = err;
     }
-    if (right_stop != grpc::StatusCode::OK) {
-        return right_stop;
+
+    try {
+        right_->stop(extra);
+    } catch (const std::exception& err) {
+        *e = err;
     }
-    return grpc::StatusCode::OK;
+
+    // if we received an err from either motor, throw it.
+    if (e) {
+        throw *e;
+    }
 }
 
 void MyBase::set_power(const Vector3& linear, const Vector3& angular, const AttributeMap& extra) {
--- a/src/viam/examples/modules/complex/base/impl.cpp
+++ b/src/viam/examples/modules/complex/base/impl.cpp
@@ -65,24 +65,27 @@ bool MyBase::is_moving() {
 }
 
 void MyBase::stop(const AttributeMap& extra) {
-    std::exception* e;
+    std::string err_message;
+    bool throw_err = false;
 
     // make sure we try to stop both motors, even if the first fails.
     try {
         left_->stop(extra);
     } catch (const std::exception& err) {
-        *e = err;
+        throw_err = true;
+        err_message = err.what();
     }
 
     try {
         right_->stop(extra);
     } catch (const std::exception& err) {
-        *e = err;
+        throw_err = true;
+        err_message = err.what();
     }
 
     // if we received an err from either motor, throw it.
-    if (e) {
-        throw *e;
+    if (throw_err) {
+        throw std::runtime_error(err_message);
     }
 }
 
--- a/src/viam/sdk/components/motor/client.cpp
+++ b/src/viam/sdk/components/motor/client.cpp
@@ -66,7 +66,7 @@ Motor::properties MotorClient::get_properties(const AttributeMap& extra) {
         .invoke([](auto& response) { return from_proto(response); });
 }
 
-grpc::StatusCode MotorClient::stop(const AttributeMap& extra) {
+void MotorClient::stop(const AttributeMap& extra) {
     viam::component::motor::v1::StopRequest request;
     viam::component::motor::v1::StopResponse response;
 
@@ -77,7 +77,9 @@ grpc::StatusCode MotorClient::stop(const AttributeMap& extra) {
     *request.mutable_extra() = map_to_struct(extra);
 
     const grpc::Status status = stub_->Stop(&ctx, request, &response);
-    return status.error_code();
+    if (!status.ok()) {
+        throw std::runtime_error(status.error_message());
+    }
 }
 
 Motor::power_status MotorClient::get_power_status(const AttributeMap& extra) {
--- a/src/viam/sdk/resource/resource.cpp
+++ b/src/viam/sdk/resource/resource.cpp
@@ -14,10 +14,6 @@ namespace sdk {
 
 Resource::~Resource() = default;
 
-void Resource::stop(const AttributeMap& extra) {
-    throw std::runtime_error("Unimplemented");
-}
-
 std::string Resource::name() const {
     return name_;
 }
--- a/src/viam/sdk/resource/resource.cpp
+++ b/src/viam/sdk/resource/resource.cpp
@@ -1,5 +1,6 @@
 #include <viam/sdk/resource/resource.hpp>
 
+#include <stdexcept>
 #include <unordered_map>
 
 #include <grpcpp/support/status.h>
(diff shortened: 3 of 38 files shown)


src/viam/examples/modules/complex/base/impl.cpp
```diff
@@ -63,18 +64,26 @@ bool MyBase::is_moving() {
...
+
+    try {
+        right_->stop(extra);
+    } catch (const std::exception& err) {
+        *e = err;
+    }
+
+    // if we received an err from either motor, throw it.
+    if (e) {
+        throw *e;
```
stuqdog, line 85: Note to reviewers: the logic here should be basically the same as before except earlier we would throw the first err we encounter, and here we throw the last. This was a bit cleaner to implement and I didn't think there was a particular reason to favor one over the other, but lemme know if you disagree!

src/viam/sdk/robot/service.cpp
```diff
@@ -160,14 +160,23 @@ ::grpc::Status RobotService_::StopAll(::grpc::ServerContext* context,
...
             }
         } else {
-            // NOLINTNEXTLINE(clang-analyzer-deadcode.DeadStores)
-            status = resource->stop();
+            try {
+                // NOLINTNEXTLINE(clang-analyzer-deadcode.DeadStores)
+                resource->stop();
+            } catch (...) {
+                status = grpc::UNKNOWN;
+            }
```
stuqdog, line 179: logic here should be pretty much the same, but have to do nested `try/catch` blocks in order to attempt `stop` both with and without an `extra`. Seems slightly clunky but simple enough that I was okay with it, happy to revisit if folks don't love this implementation!

njooma: I guess my only issue with this is that we'll never get the error from the `stop(extra)` call, since this code always assumes the issue was that we can't use `extra`, and not that there was some other gRPC/resource error.

src/viam/examples/modules/complex/base/impl.cpp
```diff
@@ -63,18 +64,26 @@ bool MyBase::is_moving() {
...
-grpc::StatusCode MyBase::stop(const AttributeMap& extra) {
-    auto left_stop = left_->stop(extra);
-    auto right_stop = right_->stop(extra);
+void MyBase::stop(const AttributeMap& extra) {
+    std::exception* e;
 
-    // Return first of any non-ok error code received from motors.
-    if (left_stop != grpc::StatusCode::OK) {
-        return left_stop;
+    // make sure we try to stop both motors, even if the first fails.
```
benjirewis, line 70: Ooh yeah; my bad we should definitely try stopping both.

src/viam/sdk/robot/service.cpp
benjirewis: Fair concern; I suppose that issue already existed. Looks like Python [only catches type errors](https://github.com/viamrobotics/viam-python-sdk/blob/main/src/viam/robot/service.py#L149); I suppose we could make the catch on L165 more granular than `...`.

stuqdog: Yeah definitely a valid concern! This was already a bit of a concern (we were before only capturing the status code of the error, not the error message itself). At this point, making sure we return the status code becomes pretty complicated but we _can_ update to return the text of the error when something fails. I made a quick change to that effect; it'll only keep track of the latest error message seen but seems better than nothing? Let me know what you think!

As far more granular matching on the kind of error, I'd propose holding off on that until we implement our custom `ViamError` type.

src/viam/examples/modules/complex/base/impl.cpp
```diff
@@ -63,18 +64,26 @@ bool MyBase::is_moving() {
...
+    try {
+        left_->stop(extra);
+    } catch (const std::exception& err) {
+        *e = err;
```
acmorrow, line 74: At first, I wondered why this compiled, since `err` is const and `e` non. Then I realized that this is actually invoking the copy assignment operator. But the object it is copying into doesn't exist! There is no object allocated at `*e`, since `e` is uninitialized. But, even if it was, this would end up slicing the object, since it would invoke the copy assignment operator for `std::exception`, but the object referred to by `err` probably is a derived class.

benjirewis (approved):
LGTM! Thanks 🧑‍🔧

src/viam/examples/modules/complex/base/impl.cpp
stuqdog: ooh thanks for catching this! Amended to a simpler, more straightforward approach to throwing here.

njooma (approved):

src/viam/sdk/components/motor/client.cpp
```diff
@@ -66,7 +66,7 @@ Motor::properties MotorClient::get_properties(const AttributeMap& extra) {
         .invoke([](auto& response) { return from_proto(response); });
 }
 
-grpc::StatusCode MotorClient::stop(const AttributeMap& extra) {
+void MotorClient::stop(const AttributeMap& extra) {
```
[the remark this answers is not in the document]
stuqdog: I agree it can! I kept it as-is just for a separation of the PR purpose, but I'm currently working on a PR to standardize use of `client_helper` and `service_helper` everywhere, and will be updating the `stop` methods in that PR.

src/viam/sdk/robot/service.cpp
```diff
@@ -147,31 +148,45 @@ ::grpc::Status RobotService_::StopAll(::grpc::ServerContext* context,
...
+                resource->stop(extra.at(rn_));
+            } catch (const std::runtime_error& err) {
+                try {
+                    status_message = err.what();
+                    // NOLINTNEXTLINE(clang-analyzer-deadcode.DeadStores)
+                    resource->stop();
+                } catch (std::runtime_error& err) {
+                    status_message = err.what();
+                    status = grpc::UNKNOWN;
+                }
```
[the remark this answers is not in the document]
stuqdog: The rationale as I recall was, "if it failed it could be because the `extra` was malformed in some way, so try calling the standard `stop` method as well just to be safe." This was from a while ago however; the "call `stop` twice" logic was in there before. My goal in this PR was to minimize changes to logic.

src/viam/sdk/robot/service.cpp
```diff
@@ -147,31 +148,45 @@ ::grpc::Status RobotService_::StopAll(::grpc::ServerContext* context,
...
+                resource->stop();
+            } catch (std::runtime_error& err) {
+                status_message = err.what();
+                status = grpc::UNKNOWN;
+            }
         }
     }
 
-    return grpc::Status(status, "");
+    return grpc::Status(status, status_message);
```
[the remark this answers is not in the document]
stuqdog: good call, I've create a ticket to address this (and potentially also the "call stop twice" addressed above) [here](https://viam.atlassian.net/browse/RSDK-5867).

src/viam/sdk/resource/resource.cpp
```diff
@@ -13,8 +14,8 @@ namespace sdk {
 
 Resource::~Resource() = default;
 
-grpc::StatusCode Resource::stop(const AttributeMap& extra) {
-    return grpc::StatusCode::UNIMPLEMENTED;
+void Resource::stop(const AttributeMap& extra) {
```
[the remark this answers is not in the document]
stuqdog: Ooh good catch. Yeah, I think in the case you describe `StopAll` will always return a non-OK grpc status. Agree that changing this to be pure virtual is better, will update and add a no-op `Stop` implementation to resources that don't currently have one.

src/viam/sdk/components/motor/client.cpp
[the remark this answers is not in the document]
acmorrow: Sounds good.

src/viam/sdk/robot/service.cpp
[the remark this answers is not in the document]
acmorrow: Thanks for filing the ticket.

src/viam/examples/modules/complex/base/impl.cpp
```diff
@@ -63,18 +64,29 @@ bool MyBase::is_moving() {
...
+    try {
+        right_->stop(extra);
+    } catch (const std::exception& err) {
+        throw_err = true;
+        err_message = err.what();
+    }
+
+    // if we received an err from either motor, throw it.
+    if (throw_err) {
+        throw std::runtime_error(err_message);
```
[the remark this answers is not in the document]
acmorrow: Ah, I had overlooked that this was just in example code. I think it is fine then. Resolving.

acmorrow (approved):
LGTM. I think having a `stop` method on `Resource` may not be the right thing to do, overall. See, for instance, `MLModelServiceClient::stop` which is a no-op, and `MLModelServiceTFLite::stop` which is obligated to exist but will never be invoked.

Perhaps `stop` should instead be pushed down into a `StoppableResource` base class, and then only those types that have real `StopRequest` messages would derive from that.

That'd add a little complexity to `StopAll` since you would need to try to downcast. But it might be worthwhile.

PR #184, opened 2023-11-28
RDSK-2417 Remove `ResourceType`

benjirewis:
cc @acmorrow


src/viam/examples/modules/complex/summation/impl.hpp
```diff
@@ -17,6 +17,7 @@ class MySummation : public Summation {
     };
     void reconfigure(Dependencies deps, ResourceConfig cfg) override;
     static std::vector<std::string> validate(ResourceConfig cfg);
+    void stop(const AttributeMap& extra) override{};
```
benjirewis, line 20: There were quite a few places I had to add `stop` due to it being a purely virtual method (I don't really understand how #182 was merged without these changes). I would say I'm against having every resource implement its own `stop` method, and would rather use a noop default method for the time being. IMO there should be an `ActuatingComponent` class or something that inherits from `Component` with virtual `is_moving` and `stop` methods, but that might be work for another ticket? Let me know if I'm missing some reason for the purely virtual `stop`.

stuqdog: There's a ticket to address this [here](
https://viam.atlassian.net/browse/RSDK-5881), I think the current proposal is to have a `StoppableResource` class that can be inherited from. 

I understand the inconvenience of having to implement a no-op `stop` for sure but I'm not convinced that having `stop` default is better. It's (trivially) less work and a bit less pointy, but it also makes it possible for people to potentially do Quite Bad things by failing to implement `stop` when it _is_ necessary. My preference would be to back out the `stop` changes here and just focus on the `ResourceType` removal, and then make the more principled change in a separate PR.

benjirewis: My bad; totally missed that ticket! Totally understand the concern; I'll remove all `stop` changes from this PR for now, thanks for the context.

acmorrow (approved):
Overall this looks good, with a question about the changes in `utils.cpp`.

Meanwhile, do we have any rough idea how many modules on the registry this breaking change will affect? I know that at least the triton module will need an update.

benjirewis:
> with a question about the changes in utils.cpp.

Sorry I don't see any GH comments in the `utils.cpp` file; which question are you referring to?

> do we have any rough idea how many modules on the registry this breaking change will affect

I _think_ it will break basically all C++ modules in the registry.

src/viam/sdk/common/utils.cpp
```diff
@@ -25,24 +25,26 @@ using viam::common::v1::ResourceName;
...
+            resource_subtype = resource->name();
         }
 
         ResourceName r;
         *r.mutable_namespace_() = kRDK;
-        *r.mutable_type() = resource->type().to_string();
         *r.mutable_name() = resource->name();
-        *r.mutable_subtype() = resource_type;
+        *r.mutable_type() = resource_type;
+        *r.mutable_subtype() = resource_subtype;
```
acmorrow, line 47: @benjirewis - These are the lines I was referring to in my comment.

acmorrow:
> > do we have any rough idea how many modules on the registry this breaking change will affect
> 
> I _think_ it will break basically all C++ modules in the registry.

It definitely will. I'm just wondering if we know somehow how many modules that might be, and, if so, how many are viam produced (100%?).

benjirewis:
> It definitely will. I'm just wondering if we know somehow how many modules that might be, and, if so, how many are viam produced (100%?)

Yeah good idea let me look into that; modules will break even more with #185, so it'll be good to get a sense of which modules we need to fix now.

src/viam/sdk/common/utils.cpp
acmorrow: Ah I see. You needed to change it because `resource->type()` is no longer a thing. At first I didn't see that and I couldn't see what it was that required changes here due to the removal of ResourceType.

I do agree that the code is pretty funky. Hopefully @stuqdog can clarify whether what you have here is OK or needs the handling you suggest, as I'm not certain either.

stuqdog: Sorry for the slow response. I think what we have here is correct. In the event that `(reg->api().to_string() != resource->dynamic_api().to_string())` then the registered model we're looking at is not of the provided resource type and so we `continue` on line 34. This isn't a case where we're creating a `ResourceName` that lacks a `resource_type` but rather a case where we aren't creating a `ResourceName` at all.

stuqdog (approved):
lgtm!

src/viam/sdk/common/utils.cpp
benjirewis: Ah cool that makes sense! Thanks for context.

benjirewis:
Gonna wait until we have a list of all C++ modules this change might break before merging.

njooma (approved):
Love seeing all the red

PR #185, opened 2023-11-28
RSDK-4844 Add better boilerplate for module development

benjirewis:
[RSDK-4844](https://viam.atlassian.net/browse/RSDK-4844)

- Renames `ModuleService_` to `ModuleService`
- Adds a `std::shared_ptr<Server> server_` field to `ModuleService`
   -  `ModuleService::server_` is created (not started) in the constructor
   -  `ModuleService::server_` is shutdown in the `ModuleService` destructor
 - Removes `std::shared_ptr<Server>` param from `add_model_from_registry` and `add_api_from_registry` 
 - Adds a `SignalManager signal_manager_` field to `ModuleService`
 - Changes the `ModuleService::start(std::shared_ptr<Server>)` method to `ModuleService::serve()` and alters its behavior
   - `serve` runs the `ModuleService::server_` until the `ModuleService::signal_manager_` finishes waiting
   - `serve` logs API/model pairs the module handles
 - Creates a new "boilerplate" overloaded `ModuleService` constructor `ModuleService(int argc, char** argv, std::vector<std::shared_ptr<ModelRegistration>>)`
   - Validates the socket path is present in `argv` and sets it on the class 
   - Sets global boost logger severity based on presence of the `--log-level` flag in `argv`
   - Registers the passed in `ModelRegistration` classes' models (not APIs)
   - Adds models and custom APIs from registry to `ModuleService`
 - Adds doc comments to `SignalManager` and `ModuleService` and makes more methods private

[RSDK-4844]: https://viam.atlassian.net/browse/RSDK-4844?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/viam/examples/modules/simple/main.cpp
+++ b/src/viam/examples/modules/simple/main.cpp
@@ -54,7 +54,7 @@ class Printer : public Generic {
     }
 
     static std::string find_to_print(ResourceConfig cfg) {
-        auto const& printer_name = cfg.name();
+        auto& printer_name = cfg.name();
         auto to_print = cfg.attributes()->find("to_print");
         if (to_print == cfg.attributes()->end()) {
             std::ostringstream buffer;
--- a/src/viam/examples/modules/simple/main.cpp
+++ b/src/viam/examples/modules/simple/main.cpp
@@ -21,7 +21,6 @@
 #include <viam/sdk/rpc/dial.hpp>
 #include <viam/sdk/rpc/server.hpp>
 
-using viam::component::generic::v1::GenericService;
 using namespace viam::sdk;
 
 // Printer is a modular resource that can print a to_print value to STDOUT when
@@ -55,7 +54,7 @@ class Printer : public Generic {
     }
 
     static std::string find_to_print(ResourceConfig cfg) {
-        auto printer_name = cfg.name();
+        auto const& printer_name = cfg.name();
         auto to_print = cfg.attributes()->find("to_print");
         if (to_print == cfg.attributes()->end()) {
             std::ostringstream buffer;
@@ -78,21 +77,6 @@ class Printer : public Generic {
 };
 
 int main(int argc, char** argv) {
-    if (argc < 2) {
-        std::cerr << "Need socket path as command line argument" << std::endl;
-        return EXIT_FAILURE;
-    }
-    std::string socket_addr = argv[1];
-
-    // Use set_logger_severity_from_args to set the boost trivial logger's
-    // severity depending on commandline arguments.
-    set_logger_severity_from_args(argc, argv);
-    BOOST_LOG_TRIVIAL(debug) << "Starting module with debug level logging";
-
-    // C++ modules must handle SIGINT and SIGTERM. You can use the SignalManager
-    // class and its wait method to handle the correct signals.
-    SignalManager signals;
-
     API generic = Generic::static_api();
     Model m("viam", "generic", "printer");
 
@@ -111,21 +95,9 @@ int main(int argc, char** argv) {
             return {};
         });
 
-    Registry::register_model(mr);
+    std::vector<std::shared_ptr<ModelRegistration>> mrs = {mr};
+    auto my_mod = std::make_shared<ModuleService>(argc, argv, mrs);
+    my_mod->serve();
 
-    // The `ModuleService_` must outlive the Server, so the declaration order
-    // here matters.
-    auto my_mod = std::make_shared<ModuleService_>(socket_addr);
-    auto server = std::make_shared<Server>();
-
-    my_mod->add_model_from_registry(server, generic, m);
-    my_mod->start(server);
-    BOOST_LOG_TRIVIAL(info) << "Module serving model " << m.to_string() << ", listening on "
-                            << socket_addr;
-
-    server->start();
-    int sig = 0;
-    auto result = signals.wait(&sig);
-    server->shutdown();
     return EXIT_SUCCESS;
 };
--- a/src/viam/examples/modules/tflite/main.cpp
+++ b/src/viam/examples/modules/tflite/main.cpp
@@ -723,13 +723,6 @@ class MLModelServiceTFLite : public vsdk::MLModelService, public vsdk::Stoppable
 };
 
 int serve(const std::string& socket_path) try {
-    // Block the signals we intend to wait for synchronously.
-    sigset_t sigset;
-    sigemptyset(&sigset);
-    sigaddset(&sigset, SIGINT);
-    sigaddset(&sigset, SIGTERM);
-    pthread_sigmask(SIG_BLOCK, &sigset, NULL);
-
     // Create a new model registration for the service.
     auto module_registration = std::make_shared<vsdk::ModelRegistration>(
         // Identify that this resource offers the MLModelService API
@@ -747,34 +740,15 @@ int serve(const std::string& socket_path) try {
     vsdk::Registry::register_model(module_registration);
 
     // Construct the module service and tell it where to place the socket path.
-    auto module_service = std::make_shared<vsdk::ModuleService_>(socket_path);
-
-    // Construct a new Server object.
-    auto server = std::make_shared<vsdk::Server>();
+    auto module_service = std::make_shared<vsdk::ModuleService>(socket_path);
 
     // Add the server as providing the API and model declared in the
     // registration.
-    module_service->add_model_from_registry(
-        server, module_registration->api(), module_registration->model());
+    module_service->add_model_from_registry(module_registration->api(),
+                                            module_registration->model());
 
     // Start the module service.
-    module_service->start(server);
-
-    // Create a thread which will start the server, await one of the
-    // blocked signals, and then gracefully shut down the server.
-    std::thread server_thread([&server, &sigset]() {
-        server->start();
-        int sig = 0;
-        auto result = sigwait(&sigset, &sig);
-        server->shutdown();
-    });
-
-    // The main thread waits for the server thread to indicate that
-    // the server shutdown has completed.
-    server->wait();
-
-    // Wait for the server thread to exit.
-    server_thread.join();
+    module_service->serve();
 
     return EXIT_SUCCESS;
 } catch (const std::exception& ex) {
--- a/src/viam/sdk/module/handler_map.cpp
+++ b/src/viam/sdk/module/handler_map.cpp
@@ -71,9 +71,9 @@ void HandlerMap_::add_model(Model model, RPCSubtype subtype) {
 
 std::ostream& operator<<(std::ostream& os, const HandlerMap_& hm) {
     for (const auto& kv : hm.handles_) {
-        os << "API: " << kv.first.api().to_string() << std::endl;
+        os << "API: " << kv.first.api().to_string() << '\n';
         for (const Model& model : kv.second) {
-            os << "\tModel: " << model.to_string() << std::endl;
+            os << "\tModel: " << model.to_string() << '\n';
         }
     }
     return os;
--- a/src/viam/sdk/module/handler_map.cpp
+++ b/src/viam/sdk/module/handler_map.cpp
@@ -69,5 +69,15 @@ void HandlerMap_::add_model(Model model, RPCSubtype subtype) {
     handles_[subtype].push_back(model);
 }
 
+std::ostream& operator<<(std::ostream& os, const HandlerMap_& hm) {
+    for (const auto& kv : hm.handles_) {
+        os << "API: " << kv.first.api().to_string() << std::endl;
+        for (const Model& model : kv.second) {
+            os << "\tModel: " << model.to_string() << std::endl;
+        }
+    }
+    return os;
+}
+
 }  // namespace sdk
 }  // namespace viam
--- a/src/viam/sdk/module/service.cpp
+++ b/src/viam/sdk/module/service.cpp
@@ -213,11 +213,10 @@ std::shared_ptr<Resource> ModuleService::get_parent_resource(Name name) {
     return grpc::Status();
 };
 
-ModuleService::ModuleService(std::string addr) {
-    module_ = std::make_shared<Module>(addr);
-    server_ = std::make_shared<Server>();
-    signal_manager_ = std::make_shared<SignalManager>();
-}
+ModuleService::ModuleService(std::string addr)
+    : module_(std::make_shared<Module>(std::move(addr))),
+      server_(std::make_shared<Server>()),
+      signal_manager_(std::make_unique<SignalManager>()) {}
 
 ModuleService::ModuleService(int argc,
                              char** argv,
@@ -227,7 +226,7 @@ ModuleService::ModuleService(int argc,
     }
     module_ = std::make_shared<Module>(argv[1]);
     server_ = std::make_shared<Server>();
-    signal_manager_ = std::make_shared<SignalManager>();
+    signal_manager_ = std::make_unique<SignalManager>();
     set_logger_severity_from_args(argc, argv);
 
     for (const std::shared_ptr<ModelRegistration>& mr : registrations) {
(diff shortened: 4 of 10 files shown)


src/viam/examples/modules/simple/main.cpp
```diff
@@ -21,7 +21,6 @@
 #include <viam/sdk/rpc/dial.hpp>
 #include <viam/sdk/rpc/server.hpp>
 
-using viam::component::generic::v1::GenericService;
```
benjirewis, line 24: Unnecessary.

src/viam/examples/modules/simple/main.cpp
```diff
@@ -112,21 +96,9 @@ int main(int argc, char** argv) {
...
-    // The `ModuleService_` must outlive the Server, so the declaration order
-    // here matters.
-    auto my_mod = std::make_shared<ModuleService_>(socket_addr);
-    auto server = std::make_shared<Server>();
-
-    my_mod->add_model_from_registry(server, generic, m);
-    my_mod->start(server);
-    BOOST_LOG_TRIVIAL(info) << "Module serving model " << m.to_string() << ", listening on "
-                            << socket_addr;
+    std::vector<std::shared_ptr<ModelRegistration>> mrs = {mr};
```
benjirewis, line 99: Why can I not just pass {mr} below or use `auto` here? I don't know 🤷🏻 ; something something implicit conversion.

src/viam/examples/modules/tflite/main.cpp
```diff
@@ -750,34 +743,15 @@ int serve(const std::string& socket_path) try {
     vsdk::Registry::register_model(module_registration);
 
     // Construct the module service and tell it where to place the socket path.
-    auto module_service = std::make_shared<vsdk::ModuleService_>(socket_path);
-
-    // Construct a new Server object.
-    auto server = std::make_shared<vsdk::Server>();
+    auto module_service = std::make_shared<vsdk::ModuleService>(socket_path);
```
benjirewis, line 746: I've left the tflite module as an example of using the "old" constructor (`ModuleService(std::string socket_path)`), and doing arg checking + model checking manually. Signal handling and manual calls to the underlying `Server` should no longer be necessary.

src/viam/sdk/module/service.cpp
```diff
@@ -44,14 +44,13 @@
 namespace viam {
 namespace sdk {
 
-namespace {
```
benjirewis, line 47: I'm not sure why this was in an unnamed namespace; happy to keep it that way if desired, but I moved it to be a private method on `ModuleService`.

src/viam/sdk/module/service.cpp
```diff
@@ -214,31 +212,53 @@ ::grpc::Status ModuleService_::Ready(::grpc::ServerContext* context,
...
+
+    int sig = 0;
+    signal_manager_->wait(&sig);
 }
 
-void ModuleService_::close() {
+ModuleService::~ModuleService() {
     BOOST_LOG_TRIVIAL(info) << "Shutting down gracefully.";
+    server_->shutdown();
+    // TODO(benji): custom cleanup functions
```
benjirewis, line 261: [RSDK-5509](https://viam.atlassian.net/browse/RSDK-5509) and an upcoming PR.

src/viam/sdk/module/service.hpp
```diff
@@ -41,36 +80,18 @@ class ModuleService_ : public viam::module::v1::ModuleService::Service {
...
-class SignalManager {
-   public:
-    SignalManager();
-    // wait on sigset
-    int wait(int* sig);
-
-   private:
-    sigset_t sigset;
+    std::shared_ptr<Server> server_;
+    std::shared_ptr<SignalManager> signal_manager_;
```
benjirewis, line 94: These could probably be `unique_ptr`s; I don't really know what I'm doing 😆 .

src/viam/sdk/module/service.cpp
```diff
@@ -214,31 +212,54 @@ ::grpc::Status ModuleService_::Ready(::grpc::ServerContext* context,
...
-    server->add_listening_port(address);
+    server_->add_listening_port(address);
 
     module_->set_ready();
-}
+    server_->start();
+
+    BOOST_LOG_TRIVIAL(info) << "Module listening on " << module_->addr();
+    BOOST_LOG_TRIVIAL(info) << "Module handles the following API/model pairs: " << std::endl
+                            << module_->handles();
```
benjirewis, line 253: These logs look like:
```
[2023-12-01 16:26:38.114981] [0x00007f78e65fdd00] [info]    Module listening on /tmp/viam-module-1887335542/MyModule.sock
[2023-12-01 16:26:38.114981] [0x00007f78e65fdd00] [info]    Module handles the following API/model pairs:
API: viam:service:summation
    Model: viam:summation:mysummation
API: viam:component:gizmo
    Model: viam:gizmo:mygizmo
API: rdk:component:base
    Model: viam:base:mybase
```
Let me know if you prefer a different format or don't like the idea of this log generally.

src/viam/examples/modules/simple/main.cpp
stuqdog: huh, that surprises me.

src/viam/sdk/module/handler_map.cpp
```diff
@@ -69,5 +69,15 @@ void HandlerMap_::add_model(Model model, RPCSubtype subtype) {
     handles_[subtype].push_back(model);
 }
 
+std::ostream& operator<<(std::ostream& os, const HandlerMap_& hm) {
+    for (const auto& kv : hm.handles_) {
+        os << "API: " << kv.first.api().to_string() << std::endl;
+        for (const Model& model : kv.second) {
+            os << "\tModel: " << model.to_string() << std::endl;
```
stuqdog, line 76: (minor) per clang-tidy PR, can we replace `std::endl` with `"\n"`?

src/viam/sdk/module/service.hpp
```diff
@@ -1,5 +1,6 @@
 #pragma once
 
+#include "viam/sdk/registry/registry.hpp"
```
stuqdog, line 3: Can we move this include to be more consistent with our include sorting elsewhere?

src/viam/sdk/module/service.cpp
stuqdog: I'm sure there was a reason but I can't recall what it is. Having it as a private method seems much better.

stuqdog (changes requested):
Nice! A couple small formatting things but this looks great.

stuqdog (approved):
lgtm!

src/viam/examples/modules/simple/main.cpp
```diff
@@ -55,7 +54,7 @@ class Printer : public Generic {
     }
 
     static std::string find_to_print(ResourceConfig cfg) {
-        auto printer_name = cfg.name();
+        auto const& printer_name = cfg.name();
```
acmorrow, line 57: I think the better change is to say `auto&`. The `const`-ness will be inferred. Plain `auto` doesn't capture references.

https://godbolt.org/z/4h9b5xq9a

src/viam/sdk/module/service.cpp
```diff
@@ -215,31 +213,53 @@ ::grpc::Status ModuleService_::Ready(::grpc::ServerContext* context,
...
+                             std::vector<std::shared_ptr<ModelRegistration>> registrations) {
+    if (argc < 2) {
+        throw std::runtime_error("Need socket path as command line argument");
+    }
+    module_ = std::make_shared<Module>(argv[1]);
+    server_ = std::make_shared<Server>();
+    signal_manager_ = std::make_unique<SignalManager>();
+    set_logger_severity_from_args(argc, argv);
+
+    for (const std::shared_ptr<ModelRegistration>& mr : registrations) {
```
acmorrow, line 232: `for (const auto& mr :` or `for (auto&& mr :`

src/viam/examples/modules/simple/main.cpp
acmorrow: I'd expect both to work?

https://godbolt.org/z/5sPWjjjzx

What error do you get when you try?

src/viam/sdk/module/service.cpp
acmorrow: Well, technically, it doesn't need any access to private state of the `ModuleService`, so it doesn't strictly need to be a member. In general, if something doesn't need to be a member of a class, it shouldn't be. I think this is basically an "implementation" helper needed only in this file to support defining `[Add|Reconfigure]Resource`. By making it a private member of `ModuleService`, it now has linkage/visibility even though it doesn't need it.

acmorrow (changes requested):
First pass, these are small things and comments on existing discussions. Once these are cleaned up I want to take a closer look at the design and API of ModuleService. I suspect it can be improved a bit, but I don't want to try to both address the small stuff and redesign the API in the same pass.

src/viam/examples/modules/simple/main.cpp
benjirewis: I get: `No matching function for call to 'make_shared' (clang  ovl_no_viable_function_in_call)`

src/viam/sdk/module/service.cpp
benjirewis: > Well, technically, it doesn't need any access to private state of the ModuleService, so it doesn't strictly need to be a member.

I think I was looking at `get_dependencies` and noticed it took a `ModuleService_* m`; that _looks_ a lot like a method on `ModuleService_` and not necessarily a helper function. I do see what you're saying about it not accessing private state, though. Happy to change back if you feel strongly.

src/viam/sdk/module/service.cpp
benjirewis: Sounds good; used the first. Just curious, it seems like you prefer `auto` when type annotations are too long? Or is it when they're obvious?

src/viam/sdk/module/service.cpp
benjirewis: They are indeed multiline and look a little gross when output by the RDK. NetCode is currently working on making module logging better, so I think I'd rather handle this at the RDK level instead of here (module authors should be able to log with `\n` and still get something vaguely greppable IMO).

src/viam/sdk/module/service.hpp
```diff
@@ -6,20 +6,59 @@
...
 namespace sdk {
 
-class ModuleService_ : public viam::module::v1::ModuleService::Service {
+/// @defgroup Module Classes related to C++ module development.
+
+/// @class SignalManager
+/// @brief Defines handling logic for SIGINT and SIGTERM required by all C++
+/// modules.
+/// @ingroup Module
+class SignalManager {
```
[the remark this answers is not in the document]
benjirewis: Ah good point; moved to its own file `module/signal_manager.*pp`. I think I'd like to still expose it as a public class.

src/viam/sdk/module/service.hpp
```diff
@@ -6,20 +6,59 @@
...
    public:
-    void start(std::shared_ptr<Server> server);
-    void close();
-    std::shared_ptr<Resource> get_parent_resource(Name name);
+    explicit SignalManager();
+
+    /// @brief Wait for SignalManager to receive SIGINT or SIGTERM.
+    /// @param sig Location reference where the signal number is stored.
+    /// @return The signal number if successful, -1 if not.
+    int wait(int* sig);
```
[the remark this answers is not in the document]
benjirewis: Cool; sounds good. I did replace `@param` with `@throws` to describe that it may throw. Not sure if that was necessary.

src/viam/sdk/module/service.hpp
```diff
@@ -6,20 +6,59 @@
...
+};
+
+/// @class ModuleService
+/// @brief Defines the gRPC receiving logic for a module. C++ module authors
+/// can construct a ModuleService and use its associated methods to write
+/// a working C++ module. See examples under `src/viam/examples/modules`.
+/// @ingroup Module
+class ModuleService : public viam::module::v1::ModuleService::Service {
+   public:
+    explicit ModuleService(std::string addr);
```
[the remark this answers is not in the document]
benjirewis: 👍🏻 added for all constructors.

src/viam/sdk/module/service.hpp
```diff
@@ -6,20 +6,59 @@
...
+    /// @brief Adds an API/model pair to the module; both the API and model should have
+    /// already been registered. If the ModuleService was constructed with a vector
+    /// of ModelRegistration, the passed in models will already be registered and added.
+    /// @param api The API to add.
+    /// @param model The model to add.
+    void add_model_from_registry(API api, Model model);
 
-    void add_api_from_registry(std::shared_ptr<Server> server, API api);
-    void add_model_from_registry(std::shared_ptr<Server> server, API api, Model model);
+   private:
```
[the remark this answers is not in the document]
benjirewis: Yeah I think we should derive privately now. Thanks for the cool example! I hadn't seen the godbolt playground before it's cool.

src/viam/sdk/module/service.hpp
```diff
@@ -41,36 +80,18 @@ class ModuleService_ : public viam::module::v1::ModuleService::Service {
...
-class SignalManager {
-   public:
-    SignalManager();
-    // wait on sigset
-    int wait(int* sig);
-
-   private:
-    sigset_t sigset;
+    std::shared_ptr<Server> server_;
+    std::unique_ptr<SignalManager> signal_manager_;
```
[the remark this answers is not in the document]
benjirewis: > Seems like signal_manager_ could just be held by value? Why does it need to be in a separate allocation?

It can be and done!

> I'm also not sure why Server can't be a unique_ptr, ditto Module.

Ah `module_` can be a unique_ptr (I thought it couldn't for some reason); changed. As for `server_`, from my other comment:

src/viam/examples/modules/simple/main.cpp
acmorrow: Hmm. Well, I don't think it is important enough to spend a lot of time on, the code you have is fine.

src/viam/sdk/module/service.cpp
acmorrow: No, I don't feel strongly, its fine as a private member.

src/viam/sdk/module/service.cpp
acmorrow: Alright, sounds good.

acmorrow (approved):
LGTM. I expect there will be further evolution of the ModuleService API, and the SignalManager too, but this is a meaningful step forward in SDK usability for module authors, so I think we should lock in the gains.

src/viam/sdk/module/service.cpp
acmorrow: I'm fairly pro-`auto` in general. In this case the type is obvious. Some interesting reading here: https://herbsutter.com/2013/08/12/gotw-94-solution-aaa-style-almost-always-auto/.

src/viam/examples/modules/simple/main.cpp
acmorrow: That's silly of the linter. It already is `const`. But I don't mind if you want to keep the `const` to placate the linter. Still, maybe we should consider turning that off: https://clang.llvm.org/extra/clang-tidy/checks/readability/qualified-auto.html#cmdoption-arg-AddConstToQualified.

benjirewis: > maybe we should consider turning that off

I'm not sure CI/our clang-tidy would complain. It might just be my local clang-tidy/IDE. I've left it without `const`, as it does seem redundant.

src/viam/sdk/module/service.cpp
benjirewis: Ah interesting stuff, thanks for the reading material. My intuition was that `auto` was like "cheating" in C++ and it was usually better to declare type. That article was insightful.

benjirewis:
This change breaks modules so will wait until Monday morning to merge and notify C++ module authors of breaking change.

PR #188, opened 2023-12-05
RSDK-5902 - misc. clang tidy fixes

stuqdog:
Small formatting fixes to cause `clang-tidy` to stop complaining.

Note: this doesn't compile yet until `Stoppable` PR is merged. Should be ready to review; will merge when able to.

### DIFF
--- a/.clang-tidy
+++ b/.clang-tidy
@@ -9,6 +9,7 @@
 # performance-unnecessary-value-param: TODO(RSDK-3007) remove this and fix all lints.
 # performance-move-const-arg: TODO(RSDK-3019)
 # misc-unused-parameters: TODO(RSDK-3008) remove this and fix all lints.
+# misc-include-cleaner: TODO(RSDK-5479) this is overly finnicky, add IWYU support and fix.
 # readability-function-cognitive-complexity: No, complexity is subjective and sometimes necessary.
 # readability-else-after-return: No, this causes code complexification
 Checks: >
@@ -30,6 +31,7 @@ Checks: >
   -performance-unnecessary-value-param,
   -performance-move-const-arg,
   -misc-unused-parameters,
+  -misc-include-cleaner,
   -readability-function-cognitive-complexity,
   -readability-else-after-return,
 WarningsAsErrors: '*'
--- a/src/viam/examples/modules/complex/proto/buf.lock
+++ b/src/viam/examples/modules/complex/proto/buf.lock
@@ -4,5 +4,5 @@ deps:
   - remote: buf.build
     owner: googleapis
     repository: googleapis
-    commit: 28151c0d0a1641bf938a7672c500e01d
-    digest: shake256:49215edf8ef57f7863004539deff8834cfb2195113f0b890dd1f67815d9353e28e668019165b9d872395871eeafcbab3ccfdb2b5f11734d3cca95be9e8d139de
+    commit: b30c5775bfb3485d9da2e87b26590ac9
+    digest: shake256:9d0caaf056949a0e1c883b9849d8a2fa66e22f18a2a48f867d1a8c700aa22abee50ad3ef0d8171637457cadc43c584998bdf3adac55da0f9e4614c72686b057d
--- a/src/viam/sdk/common/service_helper.cpp
+++ b/src/viam/sdk/common/service_helper.cpp
@@ -5,6 +5,7 @@
 namespace viam {
 namespace sdk {
 
+// NOLINTBEGIN(bugprone-exception-escape)
 ::grpc::Status ServiceHelperBase::fail(::grpc::StatusCode code, const char* message) const noexcept
     try {
     std::ostringstream stream;
@@ -37,6 +38,8 @@ namespace sdk {
 ::grpc::Status ServiceHelperBase::failUnknownException() const noexcept {
     return fail(::grpc::INTERNAL, "Failed with an unknown exception");
 }
+
+// NOLINTEND(bugprone-exception-escape)
 
 }  // namespace sdk
 }  // namespace viam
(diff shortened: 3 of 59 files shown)


.clang-tidy
```diff
@@ -9,6 +9,7 @@
 # performance-unnecessary-value-param: TODO(RSDK-3007) remove this and fix all lints.
 # performance-move-const-arg: TODO(RSDK-3019)
 # misc-unused-parameters: TODO(RSDK-3008) remove this and fix all lints.
+# misc-include-cleaner: TODO(RSDK-5479) this is overly finnicky, add IWYU support and fix.
```
stuqdog, line 12: A _lot_ of errors were generated without this change. Since there's already a ticket to add IWYU support, I figure we can clean this up at the same time.

src/viam/examples/modules/complex/proto/buf.lock
```diff
@@ -4,5 +4,5 @@ deps:
   - remote: buf.build
     owner: googleapis
     repository: googleapis
-    commit: 28151c0d0a1641bf938a7672c500e01d
-    digest: shake256:49215edf8ef57f7863004539deff8834cfb2195113f0b890dd1f67815d9353e28e668019165b9d872395871eeafcbab3ccfdb2b5f11734d3cca95be9e8d139de
+    commit: b30c5775bfb3485d9da2e87b26590ac9
+    digest: shake256:9d0caaf056949a0e1c883b9849d8a2fa66e22f18a2a48f867d1a8c700aa22abee50ad3ef0d8171637457cadc43c584998bdf3adac55da0f9e4614c72686b057d
```
stuqdog, line 8: This was changed automatically when compiling. Not positive this is right but didn't want to mess around with it.

src/viam/sdk/common/service_helper.cpp
```diff
@@ -5,6 +5,7 @@
 namespace viam {
 namespace sdk {
 
+// NOLINTBEGIN(bugprone-exception-escape)
```
stuqdog, line 8: For some reason we were getting complaints that these could throw errors, which caused `clang-tidy` to complain because of the `noexcept` marking. This is false though. I wasn't sure how to convince `clang-tidy` it was wrong so I just disabled the `bugprone-exception-escape` linting for this file.

src/viam/examples/modules/complex/proto/buf.lock
benjirewis: Looks right to me! Just an update to the googleapis buf repo; I haven't seen any errors updating locally, so I think this change is fine in this PR (or whichever PR gets in first, I think #186 has this change already).

.clang-tidy
benjirewis: Yeah SGTM; thanks for the comment on https://viam.atlassian.net/browse/RSDK-5479.

src/viam/examples/modules/complex/proto/buf.lock
benjirewis: Should be gone now after rebasing.

src/viam/sdk/common/service_helper.cpp
benjirewis: Weird... I think disabling is fine; I see other [SO posts](https://stackoverflow.com/questions/61014184/clang-tidys-bugprone-exception-escape-behaves-weirdly-with-msvc-stl) about that linter being a little buggy.

src/viam/sdk/services/motion/motion.cpp
```diff
@@ -135,8 +135,8 @@ bool operator==(const obstacle_detector& lhs, const obstacle_detector& rhs) {
 
 std::ostream& operator<<(std::ostream& os, const obstacle_detector& v) {
     os << "{ ";
-    os << "\tvision_service: " << v.vision_service << std::endl;
-    os << "\tcamera: " << v.camera << std::endl;
+    os << "\tvision_service: " << v.vision_service << "\n";
```
benjirewis, line 138: Why is `std::endl` disallowed?

benjirewis (commented):
Nice! Just a few qs. Also the `Stoppable` PR is in if you want to rebase.

src/viam/sdk/services/motion/motion.cpp
benjirewis: Ah so strange; I'm surprised the compiler can't just optimize there and _not_ flush with `std::endl` when using one of the standard library streams that automatically flushes on every write. Thanks for the context!

src/viam/sdk/common/service_helper.cpp
acmorrow: I feel like at a certain point you just want to let the compiler enforce `noexcept` by calling `terminate`. If you are at the point where you cant construct the status object safely, well, it's time to bail out.

src/viam/sdk/tests/mocks/generic_mocks.cpp
```diff
@@ -24,7 +24,7 @@ std::vector<GeometryConfig> MockGeneric::get_geometries() {
     return geometries_;
 }
 std::shared_ptr<MockGeneric> MockGeneric::get_mock_generic() {
-    auto generic = std::make_shared<MockGeneric>("mock_generic");
+    const auto& generic = std::make_shared<MockGeneric>("mock_generic");
```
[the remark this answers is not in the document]
acmorrow: Ah, I suspect it was telling you that `generic` is unmodified and so can be `const`. If that's the case, you don't need the `&`, just `const auto generic = ...` ought to do it.

acmorrow (approved):
This looks fine I just have a couple small comments and questions, nothing that needs to be changed unless you feel like it.

PR #190, opened 2023-12-15
RSDK-2655 Change RPC server to be owned by resource servers

benjirewis:
[RSDK-2655](https://viam.atlassian.net/browse/RSDK-2655)

Main changes:
- Removes all `register_server` methods
- Changes constructors of all children of `ResourceServer` to take in a `shared_ptr<Server>`
- Stores that server as `server_` in the `ResourceServer`
- Registers the relevant gRPC service for a server in `ResourceServer` construction
- Changes `create_resource_server` methods to take in a `shared_ptr<Server>`
- Calls the modified `create_resource_server` method in `ModuleService::add_api_from_registry_inlock_`

Secondary changes:
- Removes almost all `server_to_mock_pipeline` functions in favor of a further abstracted `client_to_mock_pipeline` function added to `test_utils`
- Adds a constructor to the `MockRobotService` to be used in testing
- Removes some (seemingly) duplicative tests and mock code

[RSDK-2655]: https://viam.atlassian.net/browse/RSDK-2655?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

cc @zaporter-work for interest


src/viam/examples/modules/complex/gizmo/api.cpp
```diff
@@ -22,8 +22,8 @@ GizmoRegistration::GizmoRegistration(const google::protobuf::ServiceDescriptor*
     : ResourceRegistration(service_descriptor){};
 
 std::shared_ptr<ResourceServer> GizmoRegistration::create_resource_server(
-    std::shared_ptr<ResourceManager> manager) {
-    return std::make_shared<GizmoServer>(manager);
+    std::shared_ptr<ResourceManager> manager, std::shared_ptr<Server> server) {
+    return std::make_shared<GizmoServer>(manager, server);
 };
```
benjirewis, line 27: As mentioned in the PR description, general pattern here for all `create_resource_server` methods is to accept a `std::shared_ptr<Server>` and pass it to construction of the `[TypeOfResource]Server`.

src/viam/examples/modules/complex/gizmo/api.cpp
```diff
@@ -55,8 +55,10 @@ Gizmo::Gizmo(std::string name) : Component(std::move(name)){};
 
 /* Gizmo server methods */
 
-GizmoServer::GizmoServer() : ResourceServer(std::make_shared<ResourceManager>()){};
-GizmoServer::GizmoServer(std::shared_ptr<ResourceManager> manager) : ResourceServer(manager){};
+GizmoServer::GizmoServer(std::shared_ptr<Server> server)
+    : ResourceServer(std::make_shared<ResourceManager>(), server, this){};
+GizmoServer::GizmoServer(std::shared_ptr<ResourceManager> manager, std::shared_ptr<Server> server)
+    : ResourceServer(manager, server, this){};
```
benjirewis, line 61: As mentioned in the PR description, general pattern here for all `[TypeOfResource]Server` constructors is to invoke the `ResourceServer` constructor, pass along the resource manager, and server, and also pass the `[TypeOfResource]Server` itself with `this`. Once we're in `resource_server_base`, `this` seems to no longer qualify as a gRPC service and cannot be registered via `register_service()`, so I chose to just pass `this` as a `grpc::Service*`.

src/viam/examples/modules/complex/gizmo/api.cpp
```diff
@@ -194,10 +196,6 @@ grpc::Status GizmoServer::DoTwo(::grpc::ServerContext* context,
     return grpc::Status();
 }
 
-void GizmoServer::register_server(std::shared_ptr<Server> server) {
-    server->register_service(this);
-}
```
benjirewis, line 199: `register_server` is now useless, as `register_service` is called in the `ResourceServer` constructor.

src/viam/examples/modules/complex/proto/buf.lock
```diff
@@ -4,5 +4,5 @@ deps:
   - remote: buf.build
     owner: googleapis
     repository: googleapis
-    commit: b30c5775bfb3485d9da2e87b26590ac9
-    digest: shake256:9d0caaf056949a0e1c883b9849d8a2fa66e22f18a2a48f867d1a8c700aa22abee50ad3ef0d8171637457cadc43c584998bdf3adac55da0f9e4614c72686b057d
+    commit: a86849a25cc04f4dbe9b15ddddfbc488
+    digest: shake256:e19143328f8cbfe13fc226aeee5e63773ca494693a72740a7560664270039a380d94a1344234b88c7691311460df9a9b1c2982190d0a2612eae80368718e1943
```
benjirewis, line 8: Another googleapis buf bump it seems 🤷🏻 .

src/viam/examples/modules/complex/test_complex_module.cpp
```diff
@@ -69,60 +70,25 @@ BOOST_AUTO_TEST_SUITE_END()
...
-    // Run the passed test on the created stack
-    std::forward<Lambda>(func)(client);
-    // shutdown afterwards
-    server->Shutdown();
-}
-
 BOOST_AUTO_TEST_CASE(test_do_one) {
-    gizmo_server_to_mock_pipeline([](Gizmo& client) -> void {
+    const auto mock = test_gizmo_impl::get_gizmo();
+    client_to_mock_pipeline<GizmoClient, GizmoServer>(mock, [](Gizmo& client) {
```
benjirewis, line 75: The new `client_to_mock_pipeline` requires specifying your `Client` and `Server` types and passing in your `mock` resource. The second argument lambda has access to the created client, and the mock can be accessed if needed by capturing by reference (`[&]`, I tried to use `[]` when not necessary)

src/viam/sdk/rpc/server.cpp
```diff
@@ -55,5 +55,9 @@ void Server::shutdown() {
     }
 }
 
+std::shared_ptr<grpc::Channel> Server::grpc_in_process_channel(const grpc::ChannelArguments& args) {
+    return server_->InProcessChannel(args);
+}
```
benjirewis, line 60: `client_to_mock_pipeline` needs access to an in process channel, so I exposed one on the `sdk::Server::server_` with this method.

src/viam/sdk/tests/mocks/camera_mocks.cpp
```diff
@@ -113,60 +113,6 @@ std::shared_ptr<MockCamera> MockCamera::get_mock_camera() {
     return camera;
 }
 
-MockCameraStub::MockCameraStub() : server_(std::make_shared<CameraServer>()) {
```
benjirewis, line 116: For both `camera_mocks` and `generic_mocks`, you'll see that I've removed mock client and server classes and their relevant code. From what I understand, a single test of a client method with `client_to_mock_pipeline` _should_ exercise client, server, and actual resource code, and I didn't think three separate tests (leveraging three classes like `MockCamera`, `MockCameraClient` and `MockCameraStub`) for each method was necessary. I could _totally_ be wrong in that assumption, and am happy to add this code back if desired.

src/viam/sdk/tests/mocks/generic_mocks.cpp
```diff
@@ -24,33 +24,13 @@ std::vector<GeometryConfig> MockGeneric::get_geometries(const AttributeMap& extr
     return geometries_;
 }
 std::shared_ptr<MockGeneric> MockGeneric::get_mock_generic() {
-    const auto generic = std::make_shared<MockGeneric>("mock_generic");
+    auto generic = std::make_shared<MockGeneric>("mock_generic");
```
benjirewis, line 27: Got a weird IDE error without this on the return of `generic` below: `Constness of 'generic' prevents automatic move (clang-tidy performance-no-automatic-move)` .

src/viam/sdk/tests/mocks/mock_robot.hpp
```diff
@@ -17,6 +17,9 @@ using namespace viam::sdk;
 // we should update the tests here accordingly.
 //
 class MockRobotService : public RobotService_ {
+   public:
+    MockRobotService(std::shared_ptr<Server> server) : RobotService_(server){};
```
benjirewis, line 21: Needed this constructor for `robot_client_to_mocks_pipeline` and it had to be public.

src/viam/sdk/tests/test_robot.cpp
```diff
@@ -31,30 +31,36 @@ using namespace viam::sdk;
...
-    ::grpc::ServerBuilder builder;
-    builder.RegisterService(&service);
-
-    std::unique_ptr<grpc::Server> server = builder.BuildAndStart();
-
+// Basically a copy of the client_to_mock_pipeline test utility. This version
+// has multiple mock resources and gives the passed in test case access to both
+// the robot client and the mock robot service.
+template <typename F>
+void robot_client_to_mocks_pipeline(F&& test_case) {
```
benjirewis, line 38: Couldn't quite replace this one with `client_to_mocks_pipeline`  as it was a subtly different setup; see my comment above.

src/viam/sdk/tests/test_utils.hpp
```diff
@@ -12,5 +16,48 @@ using namespace viam::sdk;
...
+//
+//        \/
+// -- MockResource
+//
+// This is as close to a real client->server->resource setup as we can get
+// without starting another process.
+//
+// The passed in test_case function will have access to the created ResourceClient.
+template <typename ClientType, typename ServerType, typename F>
+void client_to_mock_pipeline(std::shared_ptr<Resource> mock, F&& test_case) {
```
benjirewis, line 39: Stole this from @zaporter-work's pattern and made it more templated.

benjirewis (commented):
😮‍💨 this one had a lot of "decisions", let me know if you find anything strange.

src/viam/sdk/module/service.cpp
```diff
@@ -271,8 +271,8 @@ void ModuleService::add_api_from_registry_inlock_(API api, const std::lock_guard
     auto new_manager = std::make_shared<ResourceManager>();
 
     const std::shared_ptr<ResourceRegistration> rs = Registry::lookup_resource(api);
-    const std::shared_ptr<ResourceServer> resource_server = rs->create_resource_server(new_manager);
-    resource_server->register_server(server_);
+    const std::shared_ptr<ResourceServer> resource_server =
+        rs->create_resource_server(new_manager, server_);
```
benjirewis, line 275: I _thought_ I could make the `server_` field of `ModuleService` a `unique_ptr`, but I think we have to share it due to this call (a single `Server` needs to be shared between between potentially multiple `[TypeOfResource]Server` types).

src/viam/sdk/tests/test_robot.cpp
stuqdog: Is the issue just that we register multiple resources? We could make the `client_to_mocks_pipeline` take a vec of resources to register rather than a single resource and register everything in the vec... not sure this is actually better though, given the sort of unique space that the `robot` occupies and the slightly weird structure of having every other test file pass a vec with only a single element.

I think I'm convincing myself that it's fine as is, but including comment for posterity/to see if others have thoughts.

stuqdog (changes requested):
Nice, looks good! Would like to see us not expose the `grpc_in_process_channel` method from `Server` and instead use a `friend struct` in tests, otherwise this is great.

src/viam/examples/modules/complex/gizmo/api.cpp
benjirewis: Yeah that's a good idea. So I took a wack at this: it would certainly be a bit clearer, but it also does not work for some reason. gRPC service registration seems to break a bit and it's nontrivial for me to tell why. All I can see is that every test fails because any call to one our resource clients returns an exception (can't even introspect the error message for some reason). Everything _compiles_ fine but 🤷🏻 . If you feel strongly that we should try to adopt the pattern you're suggesting (which I do think is a good one), I could make a followup ticket for it?

src/viam/examples/dial/example_dial.cpp
```diff
@@ -68,7 +68,5 @@ int main() {
         std::cout << "got generic client named " << gc->name() << std::endl;
     }
 
-    robot->close();
```
benjirewis, line 71: Spoke with @stuqdog offline. Robot clients are closed in the types constructors, so this line leads to a double close and a "broken pipe" error; we should remove it.

src/viam/examples/modules/complex/gizmo/api.hpp
```diff
@@ -64,8 +64,8 @@ class GizmoClient : public Gizmo {
 // `GizmoServer` is the gRPC server implementation of a `Gizmo` component.
 class GizmoServer : public ResourceServer, public GizmoService::Service {
    public:
-    GizmoServer();
-    explicit GizmoServer(std::shared_ptr<ResourceManager> manager);
+    GizmoServer(std::shared_ptr<Server> server);
+    explicit GizmoServer(std::shared_ptr<ResourceManager> manager, std::shared_ptr<Server> server);
```
acmorrow, line 68: Various thoughts here:

- Feels boilerplate-ish if every `[TypeOfResource]Server` needs to declare two nearly identical constructors. What's the use-case for each? Is there a way we could combine them?

```
GizmoServer(std::shared_ptr<Server> server, std::shared_ptr<ResourceManager> manager = {}) : 
    ResourceServer(server, manager ? manager : std::make_shared<ResourceManager>()) {}
```

That still feels chatty. Could `ResourceServer` do the same trick instead?

- If there are going to be two signatures, I'd put the common parameter first (`std::shared_ptr<Server>`).

- I see a lot of places where `ResourceManager` and `Server` travel around together. Is there a missing abstraction? Why is there only sometimes a `ResourceManager`?

src/viam/examples/modules/complex/gizmo/api.cpp
```diff
@@ -22,8 +22,8 @@ GizmoRegistration::GizmoRegistration(const google::protobuf::ServiceDescriptor*
     : ResourceRegistration(service_descriptor){};
 
 std::shared_ptr<ResourceServer> GizmoRegistration::create_resource_server(
-    std::shared_ptr<ResourceManager> manager) {
-    return std::make_shared<GizmoServer>(manager);
+    std::shared_ptr<ResourceManager> manager, std::shared_ptr<Server> server) {
+    return std::make_shared<GizmoServer>(manager, server);
```
acmorrow, line 26: [nit]: `std::move(manager), std::move(server)`. You can argue that it makes no difference here and that'd be pretty true, but its a good habit to be consistent about using move. There's a lot of refcount jiggling happening here for basically no reason.

src/viam/examples/modules/complex/gizmo/api.cpp
```diff
@@ -55,8 +55,10 @@ Gizmo::Gizmo(std::string name) : Component(std::move(name)){};
 
 /* Gizmo server methods */
 
-GizmoServer::GizmoServer() : ResourceServer(std::make_shared<ResourceManager>()){};
-GizmoServer::GizmoServer(std::shared_ptr<ResourceManager> manager) : ResourceServer(manager){};
+GizmoServer::GizmoServer(std::shared_ptr<Server> server)
+    : ResourceServer(std::make_shared<ResourceManager>(), server, this){};
```
acmorrow, line 59: `std::move(server)`, here and below both that and `std::move(manager)`.

acmorrow, line 59: You could delegate to the two parameter `GizmoServer` constructor here.

src/viam/examples/modules/complex/gizmo/api.cpp
acmorrow: But the point of the virtual `register_server` was exactly to work around the issue you called out above, that up in `ResourceBase` the `this` pointer is not useable.

src/viam/sdk/resource/resource_server_base.hpp
```diff
@@ -8,14 +8,19 @@ namespace sdk {
...
    public:
-    virtual void register_server(std::shared_ptr<Server> server) = 0;
     const std::shared_ptr<ResourceManager>& resource_manager() const;
 
    protected:
-    ResourceServer(std::shared_ptr<ResourceManager> manager) : manager_(manager){};
+    ResourceServer(std::shared_ptr<ResourceManager> manager,
+                   std::shared_ptr<Server> server,
+                   grpc::Service* service)
+        : manager_(manager), server_(server) {
```
acmorrow, line 17: - Please move the implementation out of line.
- `std::move` on `manager` and `server`.
- Have some doubts about this approach of having the `grpc::Service*` become part of the constructor signature for an API type.

src/viam/examples/modules/complex/gizmo/api.cpp
acmorrow: I'm not sure it is a good idea to make our classes derive from gRPC classes that way. I think composition, as we are doing now, is going to get you further in the long run and better insulate you from changes in the gRPC world.

src/viam/examples/modules/complex/summation/api.cpp
```diff
@@ -23,8 +23,8 @@ SummationRegistration::SummationRegistration(
     : ResourceRegistration(service_descriptor){};
 
 std::shared_ptr<ResourceServer> SummationRegistration::create_resource_server(
-    std::shared_ptr<ResourceManager> manager) {
-    return std::make_shared<SummationServer>(manager);
+    std::shared_ptr<ResourceManager> manager, std::shared_ptr<Server> server) {
```
acmorrow, line 26: Same comments here re moves, ctor signature, etc as in Gizmo stuff.

src/viam/examples/modules/complex/gizmo/api.hpp
```diff
@@ -64,8 +64,8 @@ class GizmoClient : public Gizmo {
 // `GizmoServer` is the gRPC server implementation of a `Gizmo` component.
 class GizmoServer : public ResourceServer, public GizmoService::Service {
    public:
-    GizmoServer();
-    explicit GizmoServer(std::shared_ptr<ResourceManager> manager);
+    GizmoServer(std::shared_ptr<Server> server);
```
acmorrow, line 67: I think this ctor is the more important one to make `explicit`.

src/viam/examples/modules/complex/summation/api.hpp
```diff
@@ -60,12 +60,11 @@ class SummationClient : public Summation {
 // service.
 class SummationServer : public ResourceServer, public SummationService::Service {
    public:
-    SummationServer();
-    explicit SummationServer(std::shared_ptr<ResourceManager> manager);
+    SummationServer(std::shared_ptr<Server> server);
```
acmorrow, line 63: `explicit`

src/viam/examples/modules/complex/test_complex_module.cpp
```diff
@@ -69,60 +70,25 @@ BOOST_AUTO_TEST_SUITE_END()
...
-    GizmoClient client("testgizmo", grpc_channel);
-    // Run the passed test on the created stack
-    std::forward<Lambda>(func)(client);
-    // shutdown afterwards
-    server->Shutdown();
-}
-
 BOOST_AUTO_TEST_CASE(test_do_one) {
-    gizmo_server_to_mock_pipeline([](Gizmo& client) -> void {
+    const auto mock = test_gizmo_impl::get_gizmo();
```
acmorrow, line 74: I think there is a nice trick you can use to save some typing:

Up where you had `gizmo_server_to_mock_pipeline` before, do:

```
template<typename Lambda>
using gizmo_server_to_mock_pipeline = client_to_mock_pipeline<GizmoClient, GizmoServer, Lambda>;
```

Then you don't need to re-iterate the template parameters for the client and server everywhere. Maybe mod a rename if you don't like the name `gizmo_server_to_mock_pipeline` and want to call it `gizmo_client_to_mock_pipeline` or something.

src/viam/examples/dial/example_dial.cpp
acmorrow: Does that mean that the `close` method on `RobotClient` is dangerous and should be removed? Or, should it be ok to `close` a `RobotClient`, and the dtor for `RobotClient` should deal with it more gracefully?

src/viam/sdk/module/service.cpp
acmorrow: I think that depends on whether it is necessary for `[TypeOfResource]Server` objects to potentially extend the lifetime of the `Server` after the `ModuleService` has been destroyed. Does that actually make sense? Otherwise, the `ModuleService` could just own the `Server` object and you could pass it by bare pointer to `[TypeOfResource]Server` constructors. Their lifetime would need to be a subset of the `ModuleService`.

src/viam/sdk/rpc/server.hpp
```diff
@@ -3,11 +3,18 @@
 /// @brief Defines the `Server` class.
 #pragma once
 
+#include <grpcpp/channel.h>
```
acmorrow, line 6: Why?

src/viam/sdk/tests/mocks/generic_mocks.cpp
acmorrow: It isn't that weird. It is telling you that it will need to return by copy rather than move, since you can't move from `const`. That's a slight performance penalty, so it is warning you. I think it is a good warning and hopefully clang-tidy will tell us about it.

src/viam/sdk/tests/test_robot.cpp
acmorrow: I think it is fine how you have it.

src/viam/sdk/tests/test_utils.hpp
```diff
@@ -12,5 +16,61 @@ using namespace viam::sdk;
...
+    // Create a viam RPC server. Create a resource-specific server with that
+    // viam RPC server. Add the mock resource to the resource-specific server,
+    // and start the RPC server.
+    std::shared_ptr<sdk::Server> server = std::make_shared<sdk::Server>();
```
acmorrow, line 56: `auto`'d be fine here.

src/viam/sdk/tests/test_utils.hpp
```diff
@@ -12,5 +16,61 @@ using namespace viam::sdk;
...
+    ServerType resource_server(server);
+    resource_server.resource_manager()->add(mock->name(), mock);
+    server->start();
+
+    // Create a resource-specific client to the mock over an established
+    // in-process gRPC channel.
+    grpc::ChannelArguments args;
+    std::shared_ptr<grpc::Channel> grpc_channel = TestServer(server).grpc_in_process_channel(args);
```
acmorrow, line 64: `auto` would be fine here. Also, did you intend for the lifetime of `TestServer` to end on this line? I think it works because you extract a shared pointer. But if you ever wanted `TestServer` to do more (like hold some state) I think this would be fragile. Maybe just make a named `TestServer` object on the stack?

src/viam/examples/dial/example_dial.cpp
benjirewis: It's a good question; @stuqdog would know more about the semantics of `close()`. Judging from [the doc comments](https://github.com/viamrobotics/viam-cpp-sdk/blob/main/src/viam/sdk/robot/client.hpp#L42-L44), the necessity of calling `close` seems to depend on which way you've constructed the `RobotClient`.

src/viam/examples/modules/complex/gizmo/api.cpp
benjirewis: I'd agree that `register_server()` used to be the way that you could register a service for a `Server`; my "replacement" for this logic is to pass `this` up to `ResourceServerBase` (which is also not ideal). I don't mean to say this was a bad design or anything, just that `register_service()` gets called in `ResourceServerBase` construction now. Did you have a reason to keep these methods around?

src/viam/examples/modules/complex/gizmo/api.hpp
benjirewis: Yeah now that you mention it, I see no usages of the `[TypeOfResource]Server(std::shared_ptr<Server> server)` constructors. In other words, I think you are correct that we always create resource servers with both a `ResourceManager` and `Server`. I _could_ remove all the `[TypeOfResource]Server(std::shared_ptr<Server> server)` constructors, but that would be hypothetically backward breaking. In any case, I'll try an approach like what you've suggested with default values for `manager` and a ternary operator to make one if none is passed in (hopefully at the `ResourceServerBase` level to be as simple as possible).

src/viam/sdk/module/service.cpp
benjirewis: > Does that actually make sense? 

I don't think `[TypeOfResource]Server`s should ever outlive the `ModuleService` so that's a good point. I'll try passing by bare pointer.

src/viam/sdk/resource/resource_server_base.hpp
benjirewis: Can do the first two.

> Have some doubts about this approach of having the grpc::Service* become part of the constructor signature for an API type.

Did you have a different idea here?

src/viam/sdk/tests/test_utils.hpp
benjirewis: > Also, did you intend for the lifetime of TestServer to end on this line?

I did, but I can make an actual object on the stack instead.

benjirewis (commented):
Thanks for comments @acmorrow , still working through them but had a couple responses/qs.

src/viam/examples/modules/complex/gizmo/api.hpp
benjirewis: Actually I'll hold off on this until @stuqdog weighs in on whether we can remove `[TypeOfResource]Server` ctors that take no resource manager.

src/viam/examples/modules/complex/test_complex_module.cpp
benjirewis: Spoke a bit offline; this would be cool, but it doesn't fully work as written there. `client_to_mock_pipeline` is a function not a type, so it can't be on the right hand side of a `using` statement. I may just leave the unfortunately verbose `client_to_mock_pipeline<*Client, *Server>` pattern around.

src/viam/sdk/rpc/server.hpp
benjirewis: Not needed; good catch!

src/viam/sdk/tests/test_utils.hpp
benjirewis: Done.

src/viam/examples/dial/example_dial.cpp
stuqdog: That's correct. The thinking is/was, if people are asking us to build a channel for them (via the `at_address` call) then it makes sense for us to `close` that channel when the client is done with it. If a user is providing their own channel (via the `with_channel` call) then we don't want to make assumptions about that when to close that channel.

That said, I do think we may be painting with too broad a brush here. Just because the channel isn't being closed doesn't necessarily mean we shouldn't call the other aspects of `close` (namely, the `~thread()` calls). So I do think there's room for making `close` cleaner/making the `RobotClient` dtor to be more graceful in its handling of it.

src/viam/examples/modules/complex/gizmo/api.hpp
stuqdog: > whether we can remove [TypeOfResource]Server ctors that take no resource manager.
I think this is fine. I agree we don't seem to be using the `manager`-less version anywhere, it's easy enough to make a `ResourceManager` if we need to, and the current structure is a bit noisy.

stuqdog (approved):
Interested in seeing how ongoing conversations (`register_server()`, etc.) get resolved, but otherwise lgtm.

src/viam/examples/dial/example_dial.cpp
benjirewis: Gotcha; thanks for the context! Filed https://viam.atlassian.net/browse/RSDK-6212, as it seems slightly out of scope for this PR.

benjirewis:
Closing in favor of #199 .

PR #194, opened 2023-12-27
use EXCLUDE_FROM_ALL to shrink dependent builds

abe-winter:
## What changed
- Add `EXCLUDE_FROM_ALL` directive to sdk tests + examples
## Why
- Faster builds for dependent projects if they build the SDK rather than using an installed copy
- May unblock cpack
## Perf
**TLDR** 130 seconds -> 70 seconds

(Note: these are N=1, so may not be stable, but the similar times for this repo + the example module suggests some repeatability. I ran on my 32g ram / many core laptop)

The test command is:
```sh
mkdir build && cd build && time /snap/bin/cmake -G Ninja .. && time ninja
```

branch | project | # targets | time
---|---|---|---
main | cpp sdk | 177 | 2m12.940s
exclude-all | cpp sdk | 105 | 1m21.290s
exclude-all | example module | 108 | 1m22.449s
main | example module | 180 | 2m12.321s

### DIFF
--- a/etc/docker/tests/run_test.sh
+++ b/etc/docker/tests/run_test.sh
@@ -13,6 +13,7 @@ ninja install
 pushd src/viam/sdk/tests
 UBSAN_OPTIONS="print_stacktrace=1" ctest --output-on-failure
 popd
+ninja src/viam/examples/modules/complex/all
 pushd src/viam/examples/modules/complex
 UBSAN_OPTIONS="print_stacktrace=1" ctest --output-on-failure
 popd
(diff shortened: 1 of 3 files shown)


abe-winter:
I played around with this in a demo project. I think that warning is for when:
1. you have `install(TARGETS mylib myex)` in the root cmake config
2. `myex` is defined in examples/cmakelists, and examples dir is `exclude_from_all`

the warning *isn't* talking about an `install()` inside the examples/ directory

abe-winter:
(closing, this breaks some cmake commands)

acmorrow:
Unfortunately, I think this will result in a tree where `cmake --install . --component viam-cpp-sdk_examples` doesn't actually do anything.

PR #195, opened 2023-12-29
RSDK-5509 Allow custom handlers for module termination

benjirewis:
[RSDK-5509](https://viam.atlassian.net/browse/RSDK-5509)

Adds a public `register_cleanup_function` method to `ModuleService` that allows users to register custom logic to run before their module shuts down. Uses `register_cleanup_function` in the simple and complex module examples to log a small message before the module shuts down.


src/viam/sdk/module/service.hpp
```diff
@@ -53,6 +53,13 @@ class ModuleService : viam::module::v1::ModuleService::Service {
     /// @param model The model to add.
     void add_model_from_registry(API api, Model model);
 
+    /// @brief Registers a cleanup function to run when the module shuts down. There is
+    /// no default cleanup function and only a single cleanup function can be registered.
+    /// Cleanup functions must be registered before `serve` is called.
```
benjirewis, line 58: I chose to not allow registration of multiple cleanup functions. Hypothetically, we could allow it, maintain a "stack" of functions, and call them in a FILO order in `~ModuleService`, but I'd rather force module authors to consolidate their cleanup logic into a single function.

src/viam/sdk/module/service.cpp
```diff
@@ -250,7 +250,14 @@ void ModuleService::serve() {
...
 
 ModuleService::~ModuleService() {
-    // TODO(RSDK-5509): Run registered cleanup functions here.
+    try {
+        cleanup_function_();
+    } catch (const std::exception& xcp) {
+        BOOST_LOG_TRIVIAL(error) << "Module cleanup function failed with exception " << xcp.what();
+    } catch (...) {
+        BOOST_LOG_TRIVIAL(error) << "Module cleanup function failed with unknown exception";
+    }
```
benjirewis, line 259: This double catch is me trying to ensure that `server_` is always shut down and `parent_` is always closed (unless `cleanup_function_` segfaults or blocks forever or something).

stuqdog, line 259: I think this will cause error logging in cases where the cleanup function is never set (it certainly does in godbolt). We should maybe make this be a pointer to a func, and then only try to call when the funcptr is not null?

stuqdog (changes requested):

src/viam/sdk/module/service.cpp
benjirewis: Ooh you are very right; let me switch to a pointer and also do some more testing for these error logging paths.

src/viam/sdk/module/service.hpp
```diff
@@ -87,6 +94,7 @@ class ModuleService : viam::module::v1::ModuleService::Service {
     std::string parent_addr_;
     std::shared_ptr<Server> server_;
     SignalManager signal_manager_;
+    void (*cleanup_function_)();
```
benjirewis, line 97: Switched to what I think is a function pointer (just a pointer to a function?). The syntax here is pretty foreign to me so let me know if this looks off for a pointer to a function that takes no arguments and returns nothing. Tested with a module that has no cleanup function, one that has a cleanup function that logs something, and one that has a cleanup function that throws an error. Things seem to work as expected.

acmorrow: What motivated the switch from `std::function` to a function pointer? I'd much prefer the former, really.

acmorrow (commented):
I'm a little puzzled by this review. The ticket speaks to adding custom termination logic to the SignalManager, but that isn't exactly what is happening here. Instead, this amounts to adding user defined behavior to the ModuleService dtor by way of an injectable callback. But that could be achieved, when needed, by just deriving from `ModuleService` and adding your behavior to your derived class dtor. I'm not necessarily advocating for that design, but without some additional context/motivation or discussion around the use case, it is hard for me to know what direction to suggest to take here.

src/viam/sdk/module/service.hpp
stuqdog: The cleanup function is optional and the previous structure was causing error logging in cases where no function was set. The thinking was a ptr allows us to check if the function has been set at all, and only log if `cleanup_function_ != nullptr`. If there is a cleaner way to do this that doesn't involve using a function ptr I'm certainly happy to use that instead!

acmorrow: Ah, but a default constructed `std::function` is false-y: https://en.cppreference.com/w/cpp/utility/functional/function/operator_bool. You can't call it, but you can definitely do:

```
if (cleanup_function_)
    cleanup_function_();
```

stuqdog: Oh, great! No need to go pointer route then.

benjirewis: Ah nice; switched back to a `std::function<void()>`  but left the guard in around the actual call to `cleanup_function_`.

benjirewis:
> I'm a little puzzled by this review. The ticket speaks to adding custom termination logic to the SignalManager, but that isn't exactly what is happening here. Instead, this amounts to adding user defined behavior to the ModuleService dtor by way of an injectable callback. But that could be achieved, when needed, by just deriving from ModuleService and adding your behavior to your derived class dtor. I'm not necessarily advocating for that design, but without some additional context/motivation or discussion around the use case, it is hard for me to know what direction to suggest to take here.

It's true that I initially described the addition of this behavior as a method on `SignalManager`. Now that `SignalManager`s are constructed and managed by `ModuleService`s, I thought it made more sense to add the method to `ModuleService` instead. You're totally right that a module author could derive from `ModuleService` to add custom dtor behavior. To be clear, no module author has actually asked for this feature, it's just something I thought would be helpful. So, there's not much more context than what you're seeing here. I'm happy to forgo this change if it will only increase the `ModuleService` API surface area and confuse module authors (especially if it's common in C++ to derive from a class in order to add custom dtor behavior).

src/viam/sdk/module/service.hpp
acmorrow: Did you push that commit to the PR? I still looks like it is using a bare function pointer.

benjirewis: ...oops, sorry! Just force pushed.

src/viam/sdk/module/service.cpp
acmorrow: Even more than that, destructors are default `noexcept` so throwing from the cleanup function would lead to an immediate call to `std::terminate` if you didn't have this.

src/viam/sdk/module/service.cpp
```diff
@@ -306,5 +315,12 @@ void ModuleService::add_model_from_registry(API api, Model model) {
     return add_model_from_registry_inlock_(api, model, lock);
 }
 
+void ModuleService::register_cleanup_function(std::function<void()> func) {
+    if (cleanup_function_) {
+        throw std::runtime_error("cleanup function already registered");
+    }
+    cleanup_function_ = func;
```
acmorrow, line 322: `std::move(func)`

src/viam/examples/modules/complex/main.cpp
```diff
@@ -63,6 +63,10 @@ int main(int argc, char** argv) {
 
     std::vector<std::shared_ptr<ModelRegistration>> mrs = {mybase_mr, mygizmo_mr, mysummation_mr};
     auto my_mod = std::make_shared<ModuleService>(argc, argv, mrs);
+    // register_cleanup_function can be used to run custom logic at the time of
+    // shutdown. It must be called before `serve`.
+    my_mod->register_cleanup_function(
+        []() { BOOST_LOG_TRIVIAL(info) << "C++ complex module example cleaning up"; });
```
acmorrow, line 69: So, what can this cleanup function do that couldn't be done by re-writing the code like this:

```
{
    auto my_mod = std::make_shared<ModuleService>(argc, argv, mrs);
    my_mod->serve();
    // cleanup logic while my_mod still exists
}
```
Is there something that a registered cleanup function can do by virtue of executing "within" `my_mod`'s dtor that also can't just be written as ordinary code after `serve` returns and before `my_mod` is destroyed?

I'll note also that the cleanup function can really only have access to global state, other state in main, and the `ModuleService` itself. What functionality is enabled by having it that we can't write today by other means?

acmorrow (changes requested):

src/viam/examples/modules/complex/main.cpp
benjirewis: > Is there something that a registered cleanup function can do by virtue of executing "within" my_mod's dtor that also can't just be written as ordinary code after serve returns and before my_mod is destroyed?

I don't think it can do anything "special" for the exact reasons that you note (no access to private state of the `ModuleService` or anything that code at the end of a `main` function wouldn't have). Perhaps the only thing I could imagine is passing the signal received (SIGINT/SIGTERM/something else) to the cleanup function so the module author could do something special based on what signal was received? If that's not desirable/useful, though, I think I should probably close this PR.

acmorrow: Sure, let's close for now (we can always re-open it), and maybe move discussion over to [RSDK-5590](https://viam.atlassian.net/browse/RSDK-5509)? I don't actually want to drop the issue because I believe there is work we want to do in this space, and I think the ticket summary points in the direction of these sorts of activities:
- Registering a `terminate` handler that provides more diagnostics (exception introspection, backtracing) on unexpected shutdown via https://en.cppreference.com/w/cpp/error/set_terminate.
- Registering synchronous signal handlers for the common fatal signals (SIGSEGV, SIGILL, SIGBUS).
- Stack unwinding / backtracing support
- Allowing module authors to register their own custom async signal handlers with the signal manager (e.g. hooking SIGUSR1 to dump diagnostic data)

benjirewis: Cool; moved discussion back to the ticket.

PR #196, opened 2024-01-02
RSDK-1674: add servo wrappers

purplenicole730:


stuqdog (changes requested):
Awesome, thank you Nicole! I have some suggestions on formatting and improvements to testing, but broadly this looks great.

src/viam/sdk/components/servo/server.hpp
```diff
@@ -0,0 +1,54 @@
...
+#include <viam/api/component/servo/v1/servo.grpc.pb.h>
+
+#include <viam/sdk/resource/resource_manager.hpp>
+#include <viam/sdk/resource/resource_server_base.hpp>
+
+namespace viam {
+namespace sdk {
+
+/// @class ServoServer
+/// @brief gRPC server implementation of a `Servo` componnet.
```
benjirewis, line 16:
```suggestion
/// @brief gRPC server implementation of a `Servo` component.
```

src/viam/sdk/tests/test_servo.cpp
```diff
@@ -0,0 +1,153 @@
...
+namespace sdktests {
+namespace servo {
+
+using namespace viam::sdk;
+
+BOOST_AUTO_TEST_SUITE(test_mock)
+
+BOOST_AUTO_TEST_CASE(mock_get_position) {
+    std::shared_ptr<MockServo> servo = MockServo::get_mock_servo();
+    servo->move(static_cast<unsigned>(3));
```
benjirewis, line 33: What's up with these `static_cast`s? Tests won't compile without them? Unfortunate that users will have to do this, but maybe that's a normal C++ thing, @stuqdog?

benjirewis (approved):
Nice! Just a nit and a question.

src/viam/sdk/tests/test_servo.cpp
purplenicole730: Oh, interesting. Before it was complaining about types so I made these changes so that it would be of type `uint32_t`, but now it's working without doing so. Will be removing these now

benjirewis: Ah nice! 🎉 . I would be surprised if C++ cannot implicitly convert `3` to a `uint32_t`.

src/viam/sdk/tests/test_servo.cpp
```diff
@@ -95,6 +97,7 @@ BOOST_AUTO_TEST_SUITE(test_servo_client_server)
 template <typename Lambda>
 void server_to_mock_pipeline(Lambda&& func) {
     ServoServer servo_server;
+    std::shared_ptr<MockServo> mock = MockServo::get_mock_servo();
```
stuqdog, line 100: (nit) doesn't look like this `mock` is used anywhere, I think we can get rid of it.

stuqdog (approved):
Awesome, thanks Nicole! One nit, otherwise lgtm :)

PR #197, opened 2024-01-04
RSDK-5238 - Versioned releases

stuqdog:
Note to reviewers: the `release` job worked entirely in testing up until the actual release step (that is, everything built and uploaded, artifacts downloaded, etc., but github couldn't create the release). I did some digging and ensured that all the recommended fixes were in place, but nothing worked.

I am hopeful that the issue is simply that the job was running from a PR with a forked branch, where the workflow file didn't exist yet in `main`. I'd like to merge and test that theory; if it's wrong, I can go back to the drawing board. That said, if anyone has suggestions I'm certainly all ears.


.github/workflows/release.yml
```diff
@@ -0,0 +1,308 @@
...
+        id: commit
+        uses: EndBug/add-and-commit@v9.0.0
+        with:
+          default_author: github_actions
+          message: Bumping version to v${{ steps.which_version.outputs.version }} [skip ci]
+
+  build_macos:
+    if: github.repository_owner == 'viamrobotics'
+    needs: [prepare]
+    runs-on: [self-hosted, ARM64, macOS]
```
benjirewis, line 61: Should we still be using `self-hosted` at all?

.github/workflows/release.yml
```diff
@@ -0,0 +1,308 @@
...
+          wget -O - https://apt.kitware.com/keys/kitware-archive-latest.asc 2>/dev/null | gpg --dearmor - > /usr/share/keyrings/kitware-archive-keyring.gpg
+          echo 'deb [signed-by=/usr/share/keyrings/kitware-archive-keyring.gpg] https://apt.kitware.com/ubuntu/ focal main' > /etc/apt/sources.list.d/kitware.list
+
+          apt-get update
+          apt-get -y --no-install-recommends install -t llvm-toolchain-focal-15 clang-15 clang-tidy-15
+          apt-get -y install cmake
+
+          mkdir -p ${HOME}/opt/src
+          cd ${HOME}/opt/src && \
+            git clone --recurse-submodules -b v1.52.0 --depth 1 --shallow-submodules https://github.com/grpc/grpc && \
```
benjirewis, line 246: Why are we building gRPC from source?

.github/workflows/release.yml
```diff
@@ -0,0 +1,308 @@
...
+                -DgRPC_CARES_PROVIDER=package \
+                -DgRPC_RE2_PROVIDER=package \
+                -DgRPC_SSL_PROVIDER=package \
+                -DgRPC_INSTALL=ON \
+                -DgRPC_BUILD_TESTS=OFF \
+                -DBUILD_SHARED_LIBS=ON \
+                -DCMAKE_BUILD_TYPE=RelWithDebInfo \
+                -DCMAKE_INSTALL_PREFIX=/usr/local \
+                -DCMAKE_INSTALL_RPATH=/usr/local/lib && \
+            ninja install && \
```
benjirewis, line 261: No `ninja all` step here?

CMakeLists.txt
```diff
@@ -34,10 +34,11 @@
 # constrained by the version of CMake available on target systems.
 cmake_minimum_required(VERSION 3.25 FATAL_ERROR)
 
+set(CMAKE_PROJECT_VERSION 0.0.1)
```
benjirewis, line 37: Exciting! Do we want to start with `0.0.1` or `0.1.0`?

benjirewis (commented):
Awesome! Thanks for looking into this I'm sure it took a ton of testing. Also, if you're having problems getting the full release to run off a branch, I think it's fine to merge to main and then we can fix the workflow post-hoc if things don't seem to work.

.github/workflows/release.yml
njooma: For macOS we can continue to use self-hosted. Although GitHub now has [macOS runners}(https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners/about-github-hosted-runners#supported-runners-and-hardware-resources), but they are [more expensive](https://docs.github.com/en/billing/managing-billing-for-github-actions/about-billing-for-github-actions#minute-multipliers)

CMakeLists.txt
njooma: I think starting with `0.0.1` is fine until we remove the beta tag

.github/workflows/release.yml
```diff
@@ -0,0 +1,308 @@
...
+        with:
+          name: builds
+          path: builds
+
+  release:
+    needs: [prepare, build_macos, build_linux, build_aarch_linux]
+    if: github.repository_owner == 'viamrobotics'
+    runs-on: [self-hosted, x64]
+    container:
+      image: ghcr.io/viamrobotics/canon:amd64
```
njooma, line 294: We should run this on `ubuntu-latest`, and we don't need a container here

.github/workflows/release.yml
```diff
@@ -0,0 +1,308 @@
...
+      - name: Upload artifacts
+        uses: actions/upload-artifact@v3
+        with:
+          name: builds
+          path: builds
+
+  build_linux:
+    if: github.repository_owner == 'viamrobotics'
+    needs: [prepare]
+    runs-on: [self-hosted, x64]
```
njooma, line 187: This should run on a larger buildjet runner (probably 4 or 8 vcpu)

.github/workflows/release.yml
```diff
@@ -0,0 +1,308 @@
...
+      - name: Upload artifacts
+        uses: actions/upload-artifact@v3
+        with:
+          name: builds
+          path: builds
+
+  build_aarch_linux:
+    if: github.repository_owner == 'viamrobotics'
+    needs: [prepare]
+    runs-on: buildjet-8vcpu-ubuntu-2204-arm
```
njooma, line 112: We'll probably run into the same issue here that we did for rust-utils. We might want to use an `ubuntu:20.04` image

.github/workflows/release.yml
```diff
@@ -0,0 +1,308 @@
...
+# (erodkin) this is needed to create the release, per https://github.com/softprops/action-gh-release/issues/236
+permissions:
+  contents: write
+
+jobs:
+  prepare:
+    if: github.repository_owner == 'viamrobotics'
+    runs-on: buildjet-8vcpu-ubuntu-2204-arm
+    container:
+      image: ghcr.io/viamrobotics/canon:arm64
```
njooma, line 12: We don't need an additional container here

.github/workflows/release.yml
```diff
@@ -0,0 +1,308 @@
+name: Build + draft release
+
+# (erodkin) this is needed to create the release, per https://github.com/softprops/action-gh-release/issues/236
+permissions:
+  contents: write
+
+jobs:
+  prepare:
+    if: github.repository_owner == 'viamrobotics'
+    runs-on: buildjet-8vcpu-ubuntu-2204-arm
```
njooma, line 10: This can probably happen on just `ubuntu-latest`. Since there's no actual build step, we don't need a large buildjet instance

njooma (changes requested):
Looking good! My comments are more about which machine we should use for what step in the build process.

We shouldn't be using self-hosted at all anymore (except for macOS). We should use `ubuntu-latest` for smaller steps where we don't really need much compute power, and buildjet for when we need 4+ vcpus. 

Also, using container images to build only when we want to use tools installed in that image. So for example, we don't need an image in the `prepare` step because there's nothing in the `canon` image that is needed to prepare the release.

.github/workflows/release.yml
stuqdog: gRPC does not exist in apt for focal sadly, so we have to build from source.

.github/workflows/release.yml
stuqdog: Yeah, good question! This is cribbed a bit from some of the dockerfiles which didn't have a `ninja all` step, and it doesn't seem to be necessary to get builds to work correctly. I confess I'm not an expert on this.

.github/workflows/release.yml
stuqdog: 👍 I recall having some issues trying to get it to run at all on a `ubuntu:20.04` image but that might have just been an earlier mistake, will try to switch over!

.github/workflows/release.yml
stuqdog: Note: this change seems to cause non-aarch linux to fail quite quickly, and aarch-linux to hang indefinitely on install. Will dig into it on Tuesday.

.github/workflows/release.yml
```diff
@@ -0,0 +1,315 @@
...
+      - name: Setup build directory
+        run: mkdir builds
+
+      - name: Build
+        run: |
+          export BUF_CONFIG_DIR=$(mktemp -d)
+          mkdir build
+          pushd build/
+          cmake .. -G Ninja
+          ninja all && ninja install
```
acmorrow, line 91: I'm increasingly in favor of invoking the generated build system by way of `cmake --build`. So, here:
```
mkdir build
cmake -S . -B build -G Ninja
cmake --build build --target all install
```

I'd also consider building in verbose mode here, which you can do by passing additional arguments to `ninja` with `--`, so `cmake --build build --target all install -- -v`.

.github/workflows/release.yml
```diff
@@ -0,0 +1,315 @@
...
+
+      - name: Copy
+        run: |
+          cp build/install/lib/libviam_rust_utils.a builds/libviam_rust_utils-${{ matrix.platform }}.a
+          cp build/install/lib/libviamsdk.${{ needs.prepare.outputs.version }}.dylib builds/libviamsdk-${{ matrix.platform }}.dylib
+          cp build/install/lib/libviamapi.${{ needs.prepare.outputs.version }}.dylib builds/libviamapi-${{ matrix.platform }}.dylib
+      - name: Correct install path
+        run: |
+          install_name_tool -id "@rpath/libviamsdk.dylib" builds/libviamsdk-${{ matrix.platform }}.dylib
+          install_name_tool -id "@rpath/libviamapi.dylib" builds/libviamapi-${{ matrix.platform }}.dylib
```
acmorrow, line 102: I think you don't need to do this. You should just be able to use `cmake` in `install` mode and it'll do it for you:

```
cmake --install build --prefix builds/libviamsdk-${{ matrix.platform }}
```
It optionally takes a component, so if you really do want to restrict to just the runtime libs and not the dev components like headers and cmake support files, you could do:
```
cmake --install build --prefix builds/libviamsdk-${{ matrix.platform }} --component=viam-cpp-sdk_runtime
```

.github/workflows/release.yml
```diff
@@ -0,0 +1,315 @@
...
+
+      - name: Upload artifacts
```
acmorrow, line 104: What's the intended use case for these binary artifacts?

.github/workflows/release.yml
```diff
@@ -0,0 +1,315 @@
...
+          pushd build/
+          /usr/bin/cmake .. -G Ninja
+          ninja all -j 3 && ninja install
+          popd
+
+      - name: Copy
+        run: |
+          cp build/install/lib/libviam_rust_utils.a builds/libviam_rust_utils-${{ matrix.platform }}.a
+          cp build/install/lib/libviamapi.so.${{ needs.prepare.outputs.version }} builds/libviamapi-${{ matrix.platform }}.so
+          cp build/install/lib/libviamsdk.so.${{ needs.prepare.outputs.version }} builds/libviamsdk-${{ matrix.platform }}.so
```
acmorrow, line 290: Are these binaries functional without the gRPC binaries from `/usr/local`?

.github/workflows/release.yml
```diff
@@ -0,0 +1,315 @@
...
+          export BUF_CONFIG_DIR=$(mktemp -d)
+          mkdir build
+          pushd build/
+          cmake .. -G Ninja
+          ninja all && ninja install
+          popd
+
+      - name: Copy
+        run: |
+          cp build/install/lib/libviam_rust_utils.a builds/libviam_rust_utils-${{ matrix.platform }}.a
```
acmorrow, line 96: Why does this copy drop the `lib` part of the SDK installation tree from the name?

.github/workflows/release.yml
```diff
@@ -0,0 +1,315 @@
...
+        if: |
+          steps.is_organization_member.outputs.result == 'false'
+
+      - name: Checkout Code
+        uses: actions/checkout@v3
+
+      - name: Which Version
+        id: which_version
+        shell: bash
+        run: echo "version=$(grep 'set(CMAKE_PROJECT_VERSION' CMakeLists.txt | sed 's/)//' | awk '{print $2}')" >> $GITHUB_OUTPUT
```
acmorrow, line 37: Did you investigate my suggestion to derive the `CMAKE_PROJECT_VERSION` from tag metadata via `git describe`?

acmorrow (commented):
I'm a little puzzled by this review. Is it intending to produce binary releases for end-user consumption? Based on the title I was expecting work around automating versioning, not producing "release" builds.

.github/workflows/release.yml
stuqdog: `builds` is just a temporary path for us to store everything we want to upload. I don't think there's anything non-transient about the path.

.github/workflows/release.yml
stuqdog: The goal was to have these available for users to download/use, similar to the python SDK and rust-utils builds.

.github/workflows/release.yml
stuqdog: I did a bit but I ended up getting way too deep without feeling I was super confident in what I was doing. I'm sure that a `git describe` approach is possible but this one felt easier such that we could get _something_ out the door fast, and also I don't believe this does anything to preclude use of `git describe` in the future. If you have strong feelings that we should be using the `git describe` approach from jump then I'm certainly willing to reconsider!

stuqdog:
> I'm a little puzzled by this review. Is it intending to produce binary releases for end-user consumption? Based on the title I was expecting work around automating versioning, not producing "release" builds.

Yeah that's fair. My intent here is to have a process that follows that used by rust-utils and the Python and TS SDKs. Specifically, a job that users can run to create new downloadable versioned releases. This one is slightly more manual in that a human PR is required to bump version at this time but from conversation with @benjirewis and @njooma it seemed like this was acceptable at least in the short-term.

benjirewis:
> a human PR is required to bump version

The Mongo Go driver, Rust driver and C driver all required this at some point when I worked there FWIW.

.github/workflows/release.yml
stuqdog: I see, good to know! In that case perhaps we should scrap the whole "release binary" aspect of this PR entirely and just have the "release" job include source code. It still lets us have specific versioned releases to point folks to but will then require people to build from source.

.github/workflows/release.yml
stuqdog: We could certainly tag this commit, I was planning on making a versioned release a la the Python SDK and rust-utils. I think the plan would be to do biweekly "releases" in the same way that would bundle up the source code, so I expect we'd be leaving `0.0.1` behind pretty soon (though not hitting `0.1.0` for some time).

.github/workflows/release.yml
```diff
@@ -0,0 +1,238 @@
...
+            platform: macosx_arm64
+          - target: x86_64-apple-darwin
+            platform: macosx_x86_64
+    steps:
+      - name: Checkout Code
+        uses: actions/checkout@v3
+        with:
+          ref: ${{ needs.prepare.outputs.sha }}
+
+      - name: Install dependencies
```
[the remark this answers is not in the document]
stuqdog: There is a ticket for creating a `homebrew` formula as well!

.github/workflows/release.yml
acmorrow: It has been a while, but I believe that if you create a tag and then make a github "release" of that tag, it will automatically make a source tarball of it.

acmorrow: Also, to clarify: you still can make binary releases if you think it is useful, but you can't do it if you are building any of the dependencies (like grpc) yourself. So you probably could make releases for MacOS, Bookworm, Bullseye, and Jammy across various target CPUs. I just don't think you can do Focal.

.github/workflows/release.yml
acmorrow: I'd like to suggest that we call the tag either `releases/0.0.1` or `releases/v0.0.1`, in the interest of making it easier to ensure we pick up a `release` tag with a future `git describe`-based workflow, by using the `--match` flag.

acmorrow:
@stuqdog - Just checking in on this review: I think it is waiting on you, but I wanted to double check that it wasn't actually waiting on me for something.

.github/workflows/release.yml
stuqdog: Good to know, thanks!

stuqdog:
> @stuqdog - Just checking in on this review: I think it is waiting on you, but I wanted to double check that it wasn't actually waiting on me for something.

@acmorrow sorry I somehow missed your earlier comments! You are right, I will look into integrating those today.

.github/workflows/release.yml
```diff
@@ -0,0 +1,238 @@
...
+            image: ubuntu:20.04
+            runs_on: buildjet-8vcpu-ubuntu-2204-arm
+          - target: x86_64-unknown-linux-gnu
+            platform: linux_x86_64
+            image: ghcr.io/cross-rs/x86_64-unknown-linux-gnu:main
+            runs_on: buildjet-4vcpu-ubuntu-2204
+          - target: arm-unknown-linux-gnueabihf
+            platform: linux_armv6l
+            image: ghcr.io/cross-rs/arm-unknown-linux-gnueabihf:main
+            runs_on: buildjet-4vcpu-ubuntu-2204
```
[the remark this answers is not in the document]
stuqdog: @acmorrow To be honest these entries are cribbed from the equivalent `rust-utils` job. I don't have a strong principled reason for these ones as opposed to others. Certainly willing to to hear out reasons for removing/adding entries if you have them.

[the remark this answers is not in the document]
acmorrow: Ah, OK. That makes a little more sense. I believe those `cross-rs` images are for rust cross compilation of some sort. They are RHEL-type images, and I don't think we want or need releases for those. In particular, the resulting binaries are unlikely to work on ubuntu or debian.

I'd suggest instead building for `{x86_64, aarch64}x{debian:{bullseye,bookworm},ubuntu:22.04}`. Those are the platforms for which I think it is plausible that you can get re-usable binary installations.

I think you can skip 32-bit builds?

.github/workflows/release.yml
```diff
@@ -0,0 +1,309 @@
...
+          cmake --install build --prefix builds/libviamapi-${{ matrix.platform }}
+          cp build/install/lib/libviam_rust_utils.a builds/libviam_rust_utils-${{ matrix.platform }}.a
+
+      - name: Upload artifacts
+        uses: actions/upload-artifact@v3
+        with:
+          name: builds
+          path: builds
+
+  build_linux_ubuntu_jammy:
```
stuqdog, line 101: I wanted to have all the linux builds run in a single job but `ubuntu` and `debian` had too many nitpicky differences to make it worth trying to fit it all together.

benjirewis (approved):
LGTM but what's the deal with the uploaded artifacts? Do we think users will install the C++ SDK from the Github artifacts? Or will we point the eventual homebrew formula at those artifacts?

stuqdog:
> LGTM but what's the deal with the uploaded artifacts? Do we think users will install the C++ SDK from the Github artifacts? Or will we point the eventual homebrew formula at those artifacts?

Yeah I think for mac users we'll more likely point them to homebrew, but I don't think there's much harm in having these artifacts available as well. And, for jammy/bullseye/bookworm I think it'll be some time before we have anything in apt, so having the artifacts available seems nice.

.github/workflows/release.yml
```diff
@@ -0,0 +1,309 @@
...
+          releaseName: releases/v${{ steps.which_version.outputs.version }}
+          doNotFailIfNotFound: 'true'
+
+      - name: Cancelling - release already exists
+        uses: andymckay/cancel-action@0.2
+        if: steps.release_exists.outputs.id != ''
+
+      - name: Commit + Push
+        id: commit
+        uses: EndBug/add-and-commit@v9.0.0
```
acmorrow, line 51: Does this actually write a new tag?

.github/workflows/release.yml
```diff
@@ -0,0 +1,309 @@
...
+          export BUF_CONFIG_DIR=$(mktemp -d)
+          mkdir build
+          cmake -S . -B build -G Ninja
+          cmake --build build --target all install -- -v
+
+      - name: Copy
+        run: |
+          cmake --install build --prefix builds/libviamsdk-${{ matrix.platform }}
+          cmake --install build --prefix builds/libviamapi-${{ matrix.platform }}
+          cp build/install/lib/libviam_rust_utils.a builds/libviam_rust_utils-${{ matrix.platform }}.a
```
acmorrow, line 93: Why the manual copy of `rust_utils` still? I'd expect CMake to do that if necessary.

.github/workflows/release.yml
```diff
@@ -0,0 +1,309 @@
...
+              pkg-config \
+              protobuf-compiler \
+              protobuf-compiler-grpc \
+              software-properties-common \
+              sudo \
+              wget \
+
+          # Add the public key for the llvm repository to get the correct clang version
+          bash -c 'wget -O - https://apt.llvm.org/llvm-snapshot.gpg.key|apt-key add -'
+          apt-add-repository -y 'deb http://apt.llvm.org/jammy/ llvm-toolchain-jammy-15 main'
```
acmorrow, line 156: Do you actually need llvm? I sort of doubt it.

.github/workflows/release.yml
```diff
@@ -0,0 +1,309 @@
...
+        run: |
+          export BUF_CONFIG_DIR=$(mktemp -d)
+          mkdir build
+          cmake -S . -B build -G Ninja
+          cmake --build build --target all install -- -v
+
+      - name: Copy
+        run: |
+          cmake --install build --prefix builds/libviamsdk-${{ matrix.platform }}
+          cmake --install build --prefix builds/libviamapi-${{ matrix.platform }}
```
acmorrow, line 92: Why two `cmake --install`s here?

.github/workflows/release.yml
```diff
@@ -0,0 +1,309 @@
...
+        run: |
+          mkdir build
+          /usr/bin/cmake -S . -B build -G Ninja
+          /usr/bin/cmake --build build --target all install -- -v -j 3
+
+      - name: Copy
+        run: |
+          /usr/bin/cmake --install build --prefix builds/libviamsdk-${{ matrix.platform }}
+          /usr/bin/cmake --install build --prefix builds/libviamapi-${{ matrix.platform }}
+          cp build/install/lib/libviam_rust_utils.a builds/libviam_rust_utils-${{ matrix.platform }}.a
```
acmorrow, line 179: Same questions here: why two `cmake --install` calls, and why the manual `cp` of `viam_rust_utils`.

.github/workflows/release.yml
```diff
@@ -0,0 +1,309 @@
...
+            image: debian:bullseye
+            runs_on: buildjet-8vcpu-ubuntu-2204-arm
+            os: bullseye
+          - target: x86_64-debian-bullseye
+            platform: linux_x86_64-debian-bullseye
+            image: debian:bullseye
+            runs_on: buildjet-8vcpu-ubuntu-2204
+            os: bullseye
+          - target: aarch64-debian-bookworm
+            platform: linux_aarch64-ubuntu-jammy
```
acmorrow, line 209: `jammy` here is probably copy paste error.

.github/workflows/release.yml
```diff
@@ -0,0 +1,309 @@
...
+            image: debian:bookworm
+            runs_on: buildjet-8vcpu-ubuntu-2204-arm
+            os: bookworm
+          - target: x86_64-debian-bookworm
+            platform: linux_x86_64-ubuntu-jammy
```
acmorrow, line 214: Ditto jammy

.github/workflows/release.yml
```diff
@@ -0,0 +1,309 @@
...
+              software-properties-common \
+              sudo \
+              wget \
+              zlib1g-dev
+
+          # Add the public key for the llvm repository to get the correct clang version
+          bash -c 'wget -O - https://apt.llvm.org/llvm-snapshot.gpg.key|apt-key add -'
+          # For unclear reasons (possibly a debian bug), this install fails initially for
+          # bookworm but succeeds on second try
+          apt-add-repository -y 'deb http://apt.llvm.org/${{ matrix.os }}/ llvm-toolchain-${{ matrix.os }}-15 main'
```
acmorrow, line 260: I doubt you need llvm here too.

.github/workflows/release.yml
```diff
@@ -0,0 +1,309 @@
...
+        run: |
+          mkdir build
+          /usr/bin/cmake -S . -B build -G Ninja
+          /usr/bin/cmake --build build --target all install -- -v -j 3
+
+      - name: Copy
+        run: |
+          /usr/bin/cmake --install build --prefix builds/libviamsdk-${{ matrix.platform }}
+          /usr/bin/cmake --install build --prefix builds/libviamapi-${{ matrix.platform }}
+          cp build/install/lib/libviam_rust_utils.a builds/libviam_rust_utils-${{ matrix.platform }}.a
```
acmorrow, line 284: Ditto: double `--install` and `rust-utils`.

acmorrow (changes requested):

.github/workflows/release.yml
stuqdog: Ahh no, this is meant to commit changes made in the workflow run (the new tag is written in the `Release` step) but since we're not actually making changes (unlike in `rust-utils`) this is superfluous. Thanks for catching!

.github/workflows/release.yml
stuqdog: Ah I didn't realize `CMake` would do that for me! Will remove.

.github/workflows/release.yml
stuqdog: I had been assuming it was necessary for both `libviamsdk` and `libviamapi`. Upon testing I see this was mistaken! Will update.

.github/workflows/release.yml
```diff
@@ -151,15 +142,11 @@ jobs:
...
-          # Add the public key for the llvm repository to get the correct clang version
-          bash -c 'wget -O - https://apt.llvm.org/llvm-snapshot.gpg.key|apt-key add -'
-          apt-add-repository -y 'deb http://apt.llvm.org/jammy/ llvm-toolchain-jammy-15 main'
-
           sudo wget -O - https://apt.kitware.com/keys/kitware-archive-latest.asc 2>/dev/null | sudo gpg --dearmor - | sudo tee /usr/share/keyrings/kitware-archive-keyring.gpg >/dev/null
           sudo echo 'deb [signed-by=/usr/share/keyrings/kitware-archive-keyring.gpg] https://apt.kitware.com/ubuntu/ jammy main' | sudo tee /etc/apt/sources.list.d/kitware.list >/dev/null
 
           apt-get update
-          apt-get -y --no-install-recommends install -t llvm-toolchain-jammy-15 clang-15 clang-tidy-15
+          apt-get -y --no-install-recommends install -t clang-15 clang-tidy-15
```
acmorrow, line 149: I really think you can skip even this. The release build isn't using clang or clang-tidy as far as I know, and it will save you some CI time.

.github/workflows/release.yml
```diff
@@ -0,0 +1,281 @@
...
+      - name: Build
+        run: |
+          export BUF_CONFIG_DIR=$(mktemp -d)
+          mkdir build
+          cmake -S . -B build -G Ninja
+          cmake --build build --target all install -- -v
+
+      - name: Copy
+        run: |
+          cmake --install build --prefix builds/libviamsdk-${{ matrix.platform }}
```
acmorrow, line 84: [nitpick]: I think I'd call this `builds/viam-cpp-sdk-${{ matrix.platform }}`, as it contains more than just `libviamsdk`, here and everywhere.

.github/workflows/release.yml
```diff
@@ -0,0 +1,281 @@
...
+        run: |
+          brew install abseil cmake boost grpc protobuf xtensor
+          brew install pkg-config ninja buf
+
+      - name: Setup build directory
+        run: mkdir builds
+
+      - name: Build
+        run: |
+          export BUF_CONFIG_DIR=$(mktemp -d)
```
acmorrow, line 77: What does this `BUF_CONFIG_DIR` do?

.github/workflows/release.yml
```diff
@@ -0,0 +1,281 @@
...
+
+  release:
+    needs: [prepare, build_macos, build_linux_debian, build_linux_ubuntu_jammy]
+    if: github.repository_owner == 'viamrobotics'
+    runs-on: ubuntu-latest
+
+    steps:
+      - uses: actions/download-artifact@v3
+
+      - name: Release
```
acmorrow, line 272: Is there somewhere I can see the results of this step, or is that not really possible until this merges?

acmorrow (changes requested):
Looks really close, just small questions, a naming suggestion, and I think you can really drop clang stuff too.

.github/workflows/release.yml
stuqdog: Not possible at this time unfortunately, but the job is identical to that used by `rust-utils` so I think we should be able to analogize from the releases in that repo (though replace specific binaries with the `viam-cpp-sdk` directories).

.github/workflows/release.yml
stuqdog: `buf` tries to read a CLI config file located at `~/.config/buf/config.yaml` which we don't reliably have permission to see in CI. The config file isn't actually used however so we can just change where `buf` looks for it by overriding the `BUF_CONFIG_DIR` value. 

See https://github.com/bufbuild/buf/issues/2698 for more details.

acmorrow (approved):
LGTM!

njooma (approved):

PR #201, opened 2024-01-12
RSDK-3099: add resourceconfig tests

purplenicole730:
[Jira ticket](https://viam.atlassian.net/browse/RSDK-3099)

Notes:
* Added tests for `API`, `Name`, `Model`, `LinkConfig`, and `ResourceConfig`.
* Did not add tests for `OrientationConfig` or `GeometryConfig`. These tests assume that `OrientationConfig` and `GeometryConfig` methods work perfectly.
* **Found a bug where remotes were not stored properly** when a `Name` is converted `to_proto()` and then `from_proto()`. Fixed by:
- * `to_proto()`: Checking if `remote_name()` defined, and if so, add `remote_name()` to proto's `name()`.
- * `from_proto()`: Checking if `name_parts` is empty, and if so, `remote_name = ""`.
* Added getter functions to `LinkConfig`.


src/viam/sdk/tests/test_resource.cpp
```diff
@@ -0,0 +1,172 @@
...
+
+BOOST_AUTO_TEST_CASE(test_name) {
+    Name name1(API::from_string("ns:service:st"), "remote", "name");
+    BOOST_CHECK_EQUAL(name1.api().to_string(), "ns:service:st");
+    BOOST_CHECK_EQUAL(name1.remote_name(), "remote");
+    BOOST_CHECK_EQUAL(name1.name(), "name");
+    BOOST_CHECK_EQUAL(name1.to_string(), "ns:service:st/remote:name");
+    BOOST_CHECK_EQUAL(name1.short_name(), "remote:name");
+    // TODO: Currently, remotes not being returned from proto for whatever reason
+    // BOOST_CHECK_EQUAL(Name::from_proto(name1.to_proto()), name1);
```
purplenicole730, line 47: Bug where remotes are not being stored or retrieved properly when `to_proto()` and `from_proto()` are called

src/viam/sdk/config/resource.hpp
```diff
@@ -26,6 +26,11 @@ class ResourceConfig {
     viam::app::v1::ComponentConfig to_proto() const;
     ResourceConfig(std::string type);
     Name resource_name();
+    void set_api(API api);
+    void set_frame(LinkConfig frame);
+    void set_model(Model model);
+    void set_name(std::string name);
+    void set_namespace(std::string ns);
```
benjirewis, line 33: Hmm what's the point of adding these setters here and on `LinkConfig` (the getters on `LinkConfig` seem fine)? I see you're using them in tests, but it mostly seems like you're calling `set_*()` and then `*()` and asserting that the field was set as expected (which hypothetically does not test any existing functionality). I would be against providing these methods publicly to users if possible.

benjirewis (commented):
Nice! Just a q on the setters and tests don't seem to be passing (is that due to the remote bug?).

src/viam/sdk/config/resource.hpp
purplenicole730: The setters are just for testing, so you're right when you say that these methods are probably not the way to go. I guess what I can do is make a proto object and then call `from_proto()` on it. I will change that!

purplenicole730:
> Nice! Just a q on the setters and tests don't seem to be passing (is that due to the remote bug?).

I don't know why the test isn't failing, as it's passing locally.

src/viam/sdk/tests/test_resource.cpp
```diff
@@ -0,0 +1,172 @@
...
+    BOOST_CHECK_THROW(Model::from_str("@"), std::runtime_error);
+}
+
+BOOST_AUTO_TEST_CASE(test_linkconfig) {
+    translation t;
+    t.x = 0;
+    t.y = 1;
+    t.z = 2;
+    OrientationConfig ocfg = OrientationConfig();
+    GeometryConfig gcfg;
```
stuqdog, line 84: I suspect (but have not tested) that the errors causing test failure are centered around this value. Looking through the logs we see the following error message: `/tmp/src/viam/sdk/../../viam/sdk/spatialmath/geometry.hpp:42:7: runtime error: load of value 3983782352, which is not a valid value for type 'GeometryType'`.

What this reads as to me is the `GeometryType` value we're getting is random junk data rather than a default value. This makes some sense, because `GeometryType` is just an enum so an int under the hood. From some brief glancing around, `GeometryType` is really only used in `GeometryConfig`, and the other places where we're looking at `GeometryConfig` we're either getting it from a proto type (in which case we'd set the `GeometryType` in conversion) or we're explicitly setting it. 

I suspect if we set some sort of default value for `geometry_type_` when constructing a `GeometryConfig` then the errors (or at least the ones I saw) would go away. This is definitely something we'll want to see fixed before merging this PR. If you want to test locally and not go through the very slow turnaround process of committing and checking CI, you can use the `dockerfile`s in the SDK to imitate the CI process locally. Let me know if you want any help with that, happy to provide it!

stuqdog (changes requested):
Nice, love to see this! Some requested changes around the setters and testing but overall it looks great.

src/viam/sdk/config/resource.hpp
purplenicole730: Makes sense. I was using `GeometryConfig` as a template, not realizing that there must be a reason why we don't have these setter functions elsewhere.

src/viam/sdk/spatialmath/orientation.cpp
```diff
@@ -107,7 +107,7 @@ proto::Orientation OrientationConfig::to_proto() const {
             aa.set_y(a.y);
             aa.set_z(a.z);
             aa.set_theta(a.theta);
-            orientation.set_allocated_axis_angles(&aa);
+            *orientation.mutable_axis_angles() = aa;
```
acmorrow, line 110: You can likely ` = std::move(aa)` here, and similar below.

src/viam/sdk/tests/test_resource.cpp
acmorrow: `GeometryConfig` has no constructors at all, so all trivial fields like `geometry_type_` will be uninitialized when a `GeometryConfig` is declared on the stack. I suggest giving it a default constructor (or non-static data member initializers!) for such fields. Worth checking other related types like `pose` for similar issues.

Generally though, I dislike having a "default" for things like this. Notably `OrientationConfig` defaults to a quaternion with a particular quaternion-y value. But why that default and not some other?

My suggestion would be to 1) paper over the issue by adding some defaults to `GeometryConfig`, and then filing a follow-up ticket to take a closer look at the initialization/defaults/API of the spatial math and related types, ideally with a goal of avoiding the need for arbitrary defaults.

src/viam/sdk/tests/test_resource.cpp
```diff
@@ -0,0 +1,172 @@
...
+    api1.set_resource_type("component");
+    BOOST_CHECK_EQUAL(api1.resource_type(), "component");
+    api1.set_resource_subtype("st");
+    BOOST_CHECK_EQUAL(api1.resource_subtype(), "st");
+    BOOST_CHECK_EQUAL(api1.to_string(), "ns:component:st");
+    BOOST_CHECK(!api1.is_service_type());
+    BOOST_CHECK(api1.is_component_type());
+
+    API api2 = API::from_string("ns:service:st");
+    BOOST_CHECK_EQUAL(api2.to_string(), "ns:service:st");
```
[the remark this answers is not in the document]
purplenicole730: I chose not to because I extensively checked the functions with `api1`.
The `to_string()` function check also technically checks if the namespace and subtype is correct.
Should I still add it? There isn't a reason not to.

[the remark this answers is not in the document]
stuqdog: I think it's still useful. I don't think that we're showing here that `from_string` does the right thing w/r/t `namespace` and `subtype`, just `to_string`. It might be a bit overlapping but no harm in being explicit!

src/viam/sdk/tests/test_resource.cpp
```diff
@@ -0,0 +1,243 @@
...
+    BOOST_CHECK_EQUAL(api1.resource_type(), "component");
+    api1.set_resource_subtype("st");
+    BOOST_CHECK_EQUAL(api1.resource_subtype(), "st");
+    BOOST_CHECK_EQUAL(api1.to_string(), "ns:component:st");
+    BOOST_CHECK(!api1.is_service_type());
+    BOOST_CHECK(api1.is_component_type());
+
+    API api2 = API::from_string("ns:service:st");
+    BOOST_CHECK_EQUAL(api1.type_namespace(), "ns");
+    BOOST_CHECK_EQUAL(api1.resource_subtype(), "st");
```
stuqdog, line 35: (nit) can we move the `api1` tests here above with the `api1` block?

src/viam/sdk/tests/test_resource.cpp
purplenicole730: Done! Made a [ticket](https://viam.atlassian.net/browse/RSDK-6379).

src/viam/sdk/tests/test_resource.cpp
purplenicole730: Oh oops! These are supposed to be for `api2`! Thanks for catching that

src/viam/sdk/spatialmath/orientation.cpp
benjirewis: I _think_ @acmorrow is suggesting that you do `*orientation.mutable_axis_angles() = std::move(aa);`. I don't know if there's a big intrinsic difference between `set_allocated_axis_angles` vs `mutable_axis_angles() =`, but, from what I can tell, it's a good habit to move values around instead of accidentally copying them. What you have currently may copy the `proto::Orientation_AxisAngles` object, so I think Drew's suggesting you explicitly move the value. I believe what you had before was setting the `axis_angles` field to an rvalue, which will _also_ not copy the object if I'm not mistaken.

acmorrow: Yeah, I didn't even look so carefully. I suspect that using the `set_allocated` wasn't actually correct to begin with (and this file is the only file in the SDK that uses it), because I think it means that the proto takes ownership of the pointed to object, but here that's on the stack (!).

But all I was noting was that `*... = aa` is going to be a copy, but `*... = std::move(aa)` *might* be able to move, and will degrade to a copy if it can't, for some reason, so better to write it that way.

I'm happy with the code that is here now.

src/viam/sdk/resource/resource_api.cpp
```diff
@@ -113,7 +113,7 @@ std::string Name::short_name() const {
 viam::common::v1::ResourceName Name::to_proto() const {
     viam::common::v1::ResourceName rn;
     *rn.mutable_namespace_() = this->api().type_namespace();
-    *rn.mutable_name() = this->name();
+    *rn.mutable_name() = this->remote_name() + ":" + this->name();
```
benjirewis, line 116: What if there's no `remote_name()`? Won't this return `:foo` instead of `foo`?

src/viam/sdk/tests/test_resource.cpp
```diff
@@ -0,0 +1,250 @@
...
+    BOOST_CHECK_EQUAL(proto_gcfg.geometry_type_case(),
+                      viam::common::v1::Geometry::GeometryTypeCase::kBox);
+    BOOST_CHECK_EQUAL(proto_gcfg.box().dims_mm().x(), box.dims_mm().x());
+    BOOST_CHECK_EQUAL(proto_gcfg.box().dims_mm().y(), box.dims_mm().y());
+    BOOST_CHECK_EQUAL(proto_gcfg.box().dims_mm().z(), box.dims_mm().z());
+}
+
+BOOST_AUTO_TEST_CASE(test_resource) {
+    ResourceConfig resource1("type");
+    BOOST_CHECK_EQUAL(resource1.api().to_string(), "rdk:type:");
```
stuqdog, line 172: Out of scope for this PR, but looking at this it's pretty weird. The trailing colon, the fact that this is what we initialize for a resource (why is it the `type` that we're asking for? why not the more specific `subtype`?). From some digging I think there's definitely space to improve on the `ResourceConfig` type in general. Ticket here: https://viam.atlassian.net/browse/RSDK-6403

stuqdog (approved):
Nice, LGTM! 🚀

benjirewis (approved):
Awesome thanks!

PR #202, opened 2024-01-16
RSDK-2990 - Template static api

stuqdog:
#### Major Changes
Moves `static_api()` method to be a templated method on the `API` class, renames `dynamic_api()` methods to just `api()`.

#### Note to reviewers
There's the potential for a pretty significant change that comes with this. In short: getting the `API` on a `{ResourceType}Client` worked before because the `{ResourceType}` base class had a static `static_api` method that the `Client` type inherited. Now that the static method is on `API`, the `Client` types aren't inheriting it and so additional work is necessary. I can see two potential solutions here: 
1) start treating `{ResourceType}Client`s as `{ResourceType}`s in code. This works as expected, but is a bit of a gotcha. Additionally, if new methods existed solely on the `Client` class then we'd have to do some typecasting shenanigans to access then. I'm not sure how much of a concern this is in practice (I believe at this point all our `client`s just implement the base class methods without adding anything new) but in principle it could be concerning.
2) Additionally add the templated `struct` required by the `Api::for_t` method to all `Client` wrappers. This is verbose and adds more boilerplate, which makes the gains from this PR feel a bit smaller. It still creats a more intuitive user experience however, and is the more defensive approach imo.
3) We could create a dummy `T` in the `resource_by_name<T>(...)` method and then grab the `API` dynamically from the `T::api()` method. This seems bad because 1) it is hacky, and 2) it solves the problem in its current expression (`resource_by_name<T>` isn't working when the `T` is a `client` wrapper) only and fails to solve it if `Api::for_t<Client>` is called directly.

I ended up going with option (1) in this PR not because I'm convinced it's the better choice, but simply to provide more direct visibility into what this change necessitates. Personally I'm a bit inclined to go with option (2) but I certainly welcome feedback from reviewers on their preferences, or suggestions for a better approach.

### DIFF
--- a/src/viam/examples/camera/example_camera.cpp
+++ b/src/viam/examples/camera/example_camera.cpp
@@ -54,9 +54,9 @@ int main() {
         std::string camera_name("camera1");
 
         cout << "Getting camera: " << camera_name << endl;
-        std::shared_ptr<vs::CameraClient> camera;
+        std::shared_ptr<vs::Camera> camera;
         try {
-            camera = robot->resource_by_name<vs::CameraClient>(camera_name);
+            camera = robot->resource_by_name<vs::Camera>(camera_name);
         } catch (const std::exception& e) {
             cerr << "Failed to find " << camera_name << ". Exiting." << endl;
             throw;
--- a/src/viam/examples/modules/complex/gizmo/api.hpp
+++ b/src/viam/examples/modules/complex/gizmo/api.hpp
@@ -31,8 +31,7 @@ class Gizmo : public Component {
    public:
     // methods shared across all components
     static std::shared_ptr<ResourceRegistration> resource_registration();
-    static API static_api();
-    API dynamic_api() const override;
+    API api() const override;
 
     virtual bool do_one(std::string arg1) = 0;
     virtual bool do_one_client_stream(std::vector<std::string> arg1) = 0;
@@ -42,6 +41,13 @@ class Gizmo : public Component {
 
    protected:
     explicit Gizmo(std::string name);
+};
+
+template <>
+struct API::api_map<Gizmo> {
+    static API api() {
+        return {"viam", "component", "gizmo"};
+    }
 };
 
 // `GizmoClient` is the gRPC client implementation of a `Gizmo` component.
--- a/src/viam/examples/modules/complex/gizmo/api.hpp
+++ b/src/viam/examples/modules/complex/gizmo/api.hpp
@@ -43,12 +43,12 @@ class Gizmo : public Component {
     explicit Gizmo(std::string name);
 };
 
+namespace viam::sdk {
 template <>
-struct API::api_map<Gizmo> {
-    static API api() {
-        return {"viam", "component", "gizmo"};
-    }
+struct API::traits<Gizmo> {
+    static ::viam::sdk::API api();
 };
+}  // namespace viam::sdk
 
 // `GizmoClient` is the gRPC client implementation of a `Gizmo` component.
 class GizmoClient : public Gizmo {
--- a/src/viam/sdk/resource/resource.hpp
+++ b/src/viam/sdk/resource/resource.hpp
@@ -18,10 +18,9 @@ class Resource {
    public:
     virtual ~Resource();
     explicit Resource(std::string name);
-    static API static_api();
 
     /// @brief Returns the `API` associated with a particular resource.
-    virtual API dynamic_api() const = 0;
+    virtual API api() const = 0;
 
     /// @brief Returns a `ResourceName` for a particular resource name.
     virtual viam::common::v1::ResourceName get_resource_name(std::string name);
@@ -36,6 +35,13 @@ class Resource {
 
    private:
     std::string name_;
+};
+
+template <>
+struct API::api_map<Resource> {
+    static API api() {
+        return {"rdk", "resource", "Resource"};
+    }
 };
 
 }  // namespace sdk
--- a/src/viam/sdk/resource/resource.hpp
+++ b/src/viam/sdk/resource/resource.hpp
@@ -38,7 +38,7 @@ class Resource {
 };
 
 template <>
-struct API::api_map<Resource> {
+struct API::traits<Resource> {
     static API api() {
         return {"rdk", "resource", "Resource"};
     }
--- a/src/viam/sdk/resource/resource_api.cpp
+++ b/src/viam/sdk/resource/resource_api.cpp
@@ -162,6 +162,11 @@ bool operator<(const API& lhs, const API& rhs) {
     return lhs.to_string() < rhs.to_string();
 }
 
+std::ostream& operator<<(std::ostream& os, const API& v) {
+    os << v.to_string();
+    return os;
+}
+
 bool operator==(const Name& lhs, const Name& rhs) {
     return lhs.to_string() == rhs.to_string();
 }
--- a/src/viam/sdk/resource/resource_api.hpp
+++ b/src/viam/sdk/resource/resource_api.hpp
@@ -47,11 +47,11 @@ class API : public APIType {
     friend std::ostream& operator<<(std::ostream& os, const API& v);
 
     template <typename T>
-    struct api_map;
+    struct traits;
 
     template <typename T>
     static API get() {
-        return api_map<T>::api();
+        return traits<T>::api();
     }
 
    private:
--- a/src/viam/sdk/resource/resource_api.hpp
+++ b/src/viam/sdk/resource/resource_api.hpp
@@ -50,7 +50,7 @@ class API : public APIType {
     struct api_map;
 
     template <typename T>
-    static API for_t() {
+    static API get() {
         return api_map<T>::api();
     }
 
--- a/src/viam/sdk/resource/resource_api.hpp
+++ b/src/viam/sdk/resource/resource_api.hpp
@@ -44,6 +44,15 @@ class API : public APIType {
     bool is_service_type();
     friend bool operator==(API const& lhs, API const& rhs);
     friend bool operator<(const API& lhs, const API& rhs);
+    friend std::ostream& operator<<(std::ostream& os, const API& v);
+
+    template <typename T>
+    struct api_map;
+
+    template <typename T>
+    static API for_t() {
+        return api_map<T>::api();
+    }
 
    private:
     std::string resource_subtype_;
--- a/src/viam/examples/dial/example_dial.cpp
+++ b/src/viam/examples/dial/example_dial.cpp
@@ -63,7 +63,7 @@ int main() {
     }
 
     // ensure we can create clients to the robot
-    auto gc = robot->resource_by_name<GenericClient>("generic1");
+    auto gc = robot->resource_by_name<Generic>("generic1");
     if (gc) {
         std::cout << "got generic client named " << gc->name() << std::endl;
     }
--- a/src/viam/examples/modules/complex/client.cpp
+++ b/src/viam/examples/modules/complex/client.cpp
@@ -39,8 +39,8 @@ int main() {
 
     // Register custom gizmo and summation API so robot client can access resources
     // of that type from the server.
-    Registry::register_resource(API::for_t<Gizmo>(), Gizmo::resource_registration());
-    Registry::register_resource(API::for_t<Summation>(), Summation::resource_registration());
+    Registry::register_resource(API::get<Gizmo>(), Gizmo::resource_registration());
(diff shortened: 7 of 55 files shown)


src/viam/examples/camera/example_camera.cpp
```diff
@@ -54,9 +54,9 @@ int main() {
         std::string camera_name("camera1");
 
         cout << "Getting camera: " << camera_name << endl;
-        std::shared_ptr<vs::CameraClient> camera;
+        std::shared_ptr<vs::Camera> camera;
         try {
-            camera = robot->resource_by_name<vs::CameraClient>(camera_name);
+            camera = robot->resource_by_name<vs::Camera>(camera_name);
```
stuqdog, line 59: This is an example of the change necessitated by option (1) in the note to reviewers. Unless we add an additional templated `api_map` struct to the `CameraClient` class, `resource_by_name<CameraClient>(...)` will fail to work because `resource_by_name` needs to get the `CameraClient` API.

stuqdog:
Note to reviewers: this PR has exposed what appears to be a bug in the `resource_by_name<T>(...)` logic. The `test_get_resource` test in `test_robot.cpp` is now failing, but even on `main` this test fails as soon as we try to do anything at all with the `mock_motor` we get. 

I'm investigating and will update the PR once this is resolved but I don't think this should be blocking on review in the meantime.

src/viam/examples/camera/example_camera.cpp
benjirewis: FWIW I don't think stating the templated type as `Camera` here is too crazy. In the Go SDK, `RobotClient.ResourceByName` returns a `resource.Resource`, even though the value underlying that interface that's actually returned is a resource _client_.

src/viam/sdk/resource/resource_api.cpp
```diff
@@ -162,6 +162,11 @@ bool operator<(const API& lhs, const API& rhs) {
     return lhs.to_string() < rhs.to_string();
 }
 
+std::ostream& operator<<(std::ostream& os, const API& v) {
+    os << v.to_string();
+    return os;
+}
```
benjirewis, line 168: Good to add a `<<` implementation; thanks.

src/viam/sdk/resource/resource.hpp
```diff
@@ -18,10 +18,9 @@ class Resource {
    public:
     virtual ~Resource();
     explicit Resource(std::string name);
-    static API static_api();
 
     /// @brief Returns the `API` associated with a particular resource.
-    virtual API dynamic_api() const = 0;
+    virtual API api() const = 0;
```
benjirewis, line 23: [q] I know this isn't `dynamic_api` anymore, but where do we need access to this second "dynamic" `api` method? When do we not know the name of the class whose static `api` method to call?

src/viam/examples/modules/complex/gizmo/api.hpp
```diff
@@ -44,6 +43,13 @@ class Gizmo : public Component {
     explicit Gizmo(std::string name);
 };
 
+template <>
+struct API::api_map<Gizmo> {
```
benjirewis, line 47: Cool idea! Some qs: what's up with the `template <>`? Does this pattern ensure at compilation time that all derived classes of `Resource` implement a static `api` method?

benjirewis (commented):
🧑‍🔧 nifty! Have some qs.

src/viam/sdk/resource/resource.hpp
stuqdog: Yeah, good question! It's needed when doing methods on all resources on a `robot` where we don't know what type of `resource` we have in the abstract. So for `generate_status`, `stop_all`, etc. we need a proto `ResourceName` and we build that from the `API`.

benjirewis: > we need a proto ResourceName and we build that from the API.

Hmm I see; I suppose I question whether you need a fully qualified `ResourceName` (or just the short version) to run those methods. Seems fine to have both for now, though!

src/viam/examples/modules/complex/gizmo/api.hpp
stuqdog: Ahh yeah, good question. I think there is nothing that necessarily enforces that we get a compilation error. We'll get a runtime error if we try to call `API::for_t<lightsaber>()` but it will compile. I added the additional `test_get_api` tests for each resource to create in part to create another point of contact where someone implementing a new resource would realize that adding this template is necessary.

Another option would be to do this template creation within the `ResourceRegistration` class, which would slot nicely in with some refactors of the (already templated) `ResourceRegistration2` class. But this seems a bit more complicated and worthy of pulling out into a separate PR imo.

benjirewis: > Another option would be to do this template creation within the ResourceRegistration class, which would slot nicely in with some refactors of the (already templated) ResourceRegistration2 class.

Yeah this sounds like a valuable add for [RSDK-3030](https://viam.atlassian.net/browse/RSDK-3030) once we revisit that class. Would you mind leaving a `TODO` in `ResourceRegistration2` to think about an even further templated static `API` method? Otherwise, this PR seems like a solid step in the right direction, thank you! Curious what @acmorrow thinks.

benjirewis (approved):

src/viam/examples/camera/example_camera.cpp
acmorrow: I think there is an interesting question here. Because the effect of this change is wider than just whether `resource_by_name<FooClient>` works vs `resource_by_name<Foo>`. The former is going to return you a downcasted pointer with static type `FooClient`, and the latter will have static type `Foo`. It means that you can no longer really access the `*Client` types directly. Are they even still part of the public API of the SDK? Or are they an implementation detail now and all interaction is mediated via the `Foo` type, even though the dynamic type of the pointed to object is actually a `FooClient` object?

If the answer is that `FooClient` types are now an implementation detail, then I sort of like that the result here is that it forces us to talk about things via the resource type, not the concrete type. If `FooClient` types are *not* an implementation detail, what might they be needed for by an end-user of the SDK that they can't get by using `Foo` instead?

If we do think it is important that it be possible to get a resource with the `FooClient` static type instead of only `Foo`, I think there is a way to make both work, without too much boilerplate, by adding a `resource_type` typedef to `FooClient` and updating `resource_by_name` with some SINFAE magic to make it work for both `FooClient` and `Foo` depending on whether a `resource_type` is present. I'd need to check to make sure that can actually be done but I'm fairly confident it could.

Either way, I'm happy that this review is forcing us to decide the issue of whether the `FooClient` types are actually user-facing or an implementation detail.

src/viam/examples/modules/complex/gizmo/api.hpp
acmorrow: I think you would get a compile error on `API::for_t<Lightsaber>` if there was no specialization, not a runtime error, as long as there isn't non-specialized version that would fall back to some runtime behavior. I guess I'll see when I get further into the review.

src/viam/examples/modules/complex/gizmo/api.hpp
```diff
@@ -44,6 +43,13 @@ class Gizmo : public Component {
...
+    static API api() {
+        return {"viam", "component", "gizmo"};
```
acmorrow, line 49: Just FYI, you don't need to implement these in the header. You can declare it here but define it out of line, and I recommend doing so to keep these strings out of the headers (they were in the `.cpp` before).

src/viam/sdk/resource/resource_api.hpp
```diff
@@ -44,6 +44,15 @@ class API : public APIType {
     bool is_service_type();
     friend bool operator==(API const& lhs, API const& rhs);
     friend bool operator<(const API& lhs, const API& rhs);
+    friend std::ostream& operator<<(std::ostream& os, const API& v);
+
+    template <typename T>
+    struct api_map;
+
+    template <typename T>
+    static API for_t() {
```
acmorrow, line 53: I think you could plausibly drop the `_t` here.

src/viam/examples/camera/example_camera.cpp
benjirewis: > If the answer is that `FooClient` types are now an implementation detail

I like `FooClient` becoming hidden from SDK users; that's effectively what happens in the Go SDK. 

For more detail: There is a `resource.Resource` interface embedded in another interface like `motor.Motor`. `motor.Motor` is implemented by actual motors _and_ by `motor.MotorClient`. When an SDK user calls `motor.FromRobot(r, "mymotor")`, an object implementing `motor.Motor` is returned, despite the actual type of that object being `motor.MotorClient`. In that way, `motor.MotorClient` is hidden from SDK users.

acmorrow: Has me wondering about the `FooServer` types. Maybe they too don't actually form part of the public SDK API.

acmorrow: Yeah, interesting. The `MLModelServiceServer` type does not appear in the triton server module at all.

src/viam/sdk/resource/resource_api.hpp
stuqdog: Possibly just a linter thing, but when I tried to name it `for()` I got an error message (`Expected member name or ';' after declaration specifiers` which I took to mean that `for` is a reserved keyword and so couldn't be used here. Even if it is allowed, the fact that my editor complained about and reformatted it makes me think that having a more qualified name is going to be preferable for avoiding confusion.

src/viam/examples/modules/complex/gizmo/api.hpp
stuqdog: Ahh yes, you're absolutely right. Sorry, I had two ideas mixed up in my head! The concern I have is, if we don't implement the template specialization for `Foo` and _also_ don't call `API::for_t<Foo>` anywhere in the SDK, then `API::for_t<Foo>` just won't work but the I believe the code will compile, which I could see being a problem for users/module developers.

src/viam/examples/modules/complex/gizmo/api.hpp
stuqdog: Ahh okay great. For some reason I thought template specializations had to be in the `.hpp` file, but will move these implementations into the `.cpp` files.

src/viam/examples/modules/complex/gizmo/api.hpp
acmorrow: I think that was true before though: if you failed to implement `static_api` for some type, but also never called it (or 
called anything that called it), everything would still compile but it wouldn't be useable for clients.

src/viam/examples/modules/complex/gizmo/api.hpp
stuqdog: Oh! So not only is this cleaner and keeping strings outside of headers, but somehow it seems to have solved that bizarre bug I was encountering. I'd be lying if I said I understood _why_ that was the case but I'm happy to see resolution at least.

src/viam/sdk/resource/resource_api.hpp
acmorrow: Oh, of course. It's a keyword then. Sorry. I thought you were trying to echo the `_t` from type trait helper functions (e.g. `std::remove_reference_t`), which I don't think you need, rather than working around it being a keyword. Maybe `API::get<Motor>` or `API::of<Motor>`?

stuqdog: Sure, I like `API::get<T>`. Just changed!

src/viam/examples/camera/example_camera.cpp
stuqdog: Yeah the more I think about it the more it seems that both the `Server` and `Client` types are implementation details that users don't really need to know about/interact with. So I think I'm pretty happy with changing the `resource_by_name` call pattern from being `<FooClient>` to `<Foo>`.

src/viam/examples/modules/complex/gizmo/api.hpp
```diff
@@ -44,6 +43,13 @@ class Gizmo : public Component {
     explicit Gizmo(std::string name);
 };
 
+template <>
+struct API::api_map<Gizmo> {
+    static API api() {
```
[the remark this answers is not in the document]
stuqdog: Ahh yeah sorry, somehow missed the examples! Will update.

acmorrow (approved):
Ship it!

PR #203, opened 2024-01-24
RSDK-6374 Add wrappers for the generic service

benjirewis:
[RSDK-6374](https://viam.atlassian.net/browse/RSDK-6374)

Adds wrappers for the new generic service (almost identical to generic component but makes more sense semantically for module authors that want to implement a generic piece of logic without the concept of a piece of hardware). Renames preexisting `*Generic*` types to `*GenericComponent*`. While this is backward-breaking, I think it's preferable to leaving those as `*Generic*` and just having new `GenericService` classes.

[RSDK-6374]: https://viam.atlassian.net/browse/RSDK-6374?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/viam/examples/modules/simple/main.cpp
```diff
@@ -8,25 +8,25 @@
...
+#include <viam/sdk/services/generic/generic.hpp>
+#include <viam/sdk/services/service.hpp>
 
 using namespace viam::sdk;
 
 // Printer is a modular resource that can print a to_print value to STDOUT when
 // a DoCommand request is received or when reconfiguring. The to_print value
 // must be provided as an attribute in the config.
-class Printer : public Generic {
+class Printer : public GenericService {
```
benjirewis, line 29: This is a great example of a modular resource that is not actually a generic component, but a generic service (no need for `get_geometries` and it makes more sense semantically).

src/viam/sdk/components/generic/generic.hpp
```diff
@@ -14,34 +15,33 @@
...
 /// This acts as an abstract base class to be inherited from by any drivers representing
 /// specific generic implementations. This class cannot be used on its own.
-class Generic : public Component {
+class GenericComponent : public Component {
    public:
-    /// @brief Creates a `ResourceRegistration` for the `Generic` component.
+    /// @brief Creates a `ResourceRegistration` for `GenericComponent`.
     static std::shared_ptr<ResourceRegistration> resource_registration();
 
-    /// @brief Creates a `Generic` `API`.
```
benjirewis, line 43: This doc comment was misplaced.

acmorrow (approved):
LGTM. I went through pretty quickly because it seems largely mechanical.

stuqdog (approved):
lgtm!

PR #205, opened 2024-01-25
RSDK-5238 - fix cmake for bullseye

stuqdog:
#### Major changes

- Move bullseye fixes into separate bespoke step. Not sure why this was working before but not anymore, but either way it's cleaner now.
- Upload tarball of directories for release rather than trying to add whole directory.


njooma (approved):

purplenicole730 (commented):
I'm not familiar with cmake and what a workflow entails, so I don't have the knowledge necessary to review this PR.

PR #207, opened 2024-01-28
RSDK-4573: remove proto types from robot client apis

purplenicole730:
[Jira Ticket](https://viam.atlassian.net/browse/RSDK-4573)

Removed the proto types that were in the robot client API's:
- defined new `discovery_query`, `discovery`, `frame_system_config`, `status`, and `operation`
- changed robot client API's to use already defined `WorldState::transform` and `Name`
- defined proto conversion methods in a private namespace in `robot/client.cpp`
- created functions to return custom defined types while keeping old proto functions for `MockRobotService`
- added new tests to compare mock proto and custom methods to ensure same fields
- changed `test_robot.cpp` to test custom types
- added a new Impl object to handle `robot::v1::RobotService`


src/viam/sdk/robot/client.hpp
```diff
@@ -23,13 +25,50 @@ namespace viam {
...
+};
+
+struct status {
+    status(){};
+    viam::robot::v1::Status to_proto() const;
+    static status from_proto(const viam::robot::v1::Status& proto);
+
+    std::optional<Name> name;
+    std::optional<AttributeMap> status_map;
+    std::optional<time_point> last_reconfigured;
```
purplenicole730, line 68: This is giving errors, but when I do try to use `boost::optional`, I automatically switches to `std::optional` when I run `ninja all && ninja install`. Not entirely sure what to do here.

acmorrow: `std::optional` is only part of C++17, and the driver targets C++14. Somewhat unfortunately, our abseil dependency can automatically induce the driver to build in C++17 mode, since certain versions of abseil require it. Notably, the version of abseil in homebrew requires C++17, so builds on macOS are non-canonical and it is easy to accidentally C++17.

So, you need to use `boost::optional` here. I don't totally understand what you mean by " it automatically switches to std::optional". Can you elaborate, or, better, link to a log showing what you are seeing?

acmorrow: Meanwhile, this reminded me to file https://viam.atlassian.net/browse/RSDK-6561

purplenicole730: I don't know how to elaborate and I'm not sure where the logs are for this. But I do quite literally mean, running `ninja all && ninja install` changes `boost::optional` to `std::optional`.

```
    boost::optional<Name> name;
    boost::optional<AttributeMap> status_map;
    boost::optional<time_point> last_reconfigured;
```
then I run: `ninja all && ninja install`, and I look at the same code again and it is now:

...Never mind. Now that I tried it again to show you, it is no longer automatically switching. I'm not sure what was happening there.

src/viam/sdk/robot/client.hpp
```diff
@@ -23,13 +25,62 @@ namespace viam {
 namespace sdk {
 
 using grpc::Channel;
-using viam::common::v1::ResourceName;
-using viam::common::v1::Transform;
-using viam::robot::v1::FrameSystemConfig;
 using viam::robot::v1::RobotService;
```
purplenicole730, line 28: I wasn't really sure what to do with this.

acmorrow: Because of `std::unique_ptr<RobotService::Stub> stub_` in `RobotClient`? If so, I have a few suggestions:
 
- You could introduce a `Robot` base class. Then `RobotClient` would implement it and the factory functions would return a `shared_ptr<Robot>` but it would point to an instance of `RobotClient`.
- You could use the "pimpl" idiom to make a little opaque holder for the `Stub`:

In `client.hpp`:
```
class RobotClient {
  ...
private:
    struct Impl;
    unique_ptr<Impl> impl_;
};
```

Then in `client.cpp`:
```
struct RobotClient::Impl {
    Impl(RobotService::Stub stub) : stub_(std::move(stub)) {}
    RobotService::Stub stub_;
};

...

class RobotClient::RobotClient(...) : ..., impl_(std::make_unique<Impl>(RobotService::NewStub(channel_)), ... {
  ...
}

```

Unless @stuqdog wants to introduce a `Robot` base class, I think I'd follow the second suggestion for now. It can always be revisited.

src/viam/sdk/robot/client.hpp
```diff
@@ -23,13 +25,62 @@ namespace viam {
...
+    friend bool operator==(const discovery_query& lhs, const discovery_query& rhs);
+};
+
+struct status {
+    status(){};
+    viam::robot::v1::Status to_proto() const;
+    static status from_proto(const viam::robot::v1::Status& proto);
+
+    boost::optional<Name> name;
+    boost::optional<AttributeMap> status_map;
```
acmorrow, line 67: If I recall correctly, `AttributeMap` is already a `shared_ptr`? I don't think you need the optional.

src/viam/sdk/robot/client.hpp
```diff
@@ -23,13 +25,62 @@ namespace viam {
...
+
+struct operation {
+    operation(std::string id, std::string method) : id(std::move(id)), method(std::move(method)){};
+    viam::robot::v1::Operation to_proto() const;
+    static operation from_proto(const viam::robot::v1::Operation& proto);
+
+    std::string id;
+    std::string method;
+    boost::optional<std::string> session_id;
+    boost::optional<AttributeMap> arguments;
```
acmorrow, line 80: Ditto

src/viam/sdk/robot/client.hpp
```diff
@@ -23,13 +25,62 @@ namespace viam {
...
+    viam::robot::v1::DiscoveryQuery to_proto() const;
+    static discovery_query from_proto(const viam::robot::v1::DiscoveryQuery& proto);
+
+    std::string subtype;
+    std::string model;
+    friend bool operator==(const discovery_query& lhs, const discovery_query& rhs);
+};
+
+struct discovery {
+    discovery(){};
```
acmorrow, line 42: `= default`, etc.

src/viam/sdk/robot/client.hpp
```diff
@@ -23,13 +25,62 @@ namespace viam {
...
+    static frameSystemConfig from_proto(const viam::robot::v1::FrameSystemConfig& proto);
+    frameSystemConfig(WorldState::transform frame) : frame(std::move(frame)){};
+
+    WorldState::transform frame;
+    AttributeMap kinematics;
+    friend bool operator==(const discovery_query& lhs, const discovery_query& rhs);
+};
+
+struct status {
+    status(){};
```
acmorrow, line 62: `=default`, etc.

src/viam/sdk/robot/client.hpp
```diff
@@ -68,10 +119,10 @@ class RobotClient {
     static std::shared_ptr<RobotClient> with_channel(std::shared_ptr<ViamChannel> channel,
                                                      Options options);
     RobotClient(std::shared_ptr<ViamChannel> channel);
-    std::vector<ResourceName>* resource_names();
+    std::vector<Name>* resource_names();
```
acmorrow, line 122: Whoa. This returns a pointer to an internal field? That seems really dangerous. How bad would it be to convert this to `const std::vector<Name>& resource_names() const`? Or even return a copy?

src/viam/sdk/robot/client.hpp
```diff
@@ -81,54 +132,50 @@ class RobotClient {
...
     /// @throws `std::runtime_error` if the requested resource doesn't exist or is the wrong type.
     /// @return a `shared_ptr` to the requested resource.
     std::shared_ptr<T> resource_by_name(std::string name) {
-        ResourceName r;
         API api = API::get<T>();
-        *r.mutable_namespace_() = api.type_namespace();
-        *r.mutable_type() = api.resource_type();
-        *r.mutable_subtype() = api.resource_subtype();
-        *r.mutable_name() = std::move(name);
+        Name r = Name::from_string(api.to_string() + "/" + name);
```
acmorrow, line 144: This is more "assembling a string to have `Name::from_string` decompose it". Is there a better `Name` API to use here?

src/viam/sdk/robot/client.hpp
```diff
@@ -138,8 +185,8 @@ class RobotClient {
     void stop_all();
 
     /// @brief Cancel all operations for the robot and stop all actuators and movement.
-    /// @param extra Any extra params to pass to resources' `stop` methods, keyed by `ResourceName`.
-    void stop_all(std::unordered_map<ResourceName,
+    /// @param extra Any extra params to pass to resources' `stop` methods, keyed by `Name`.
+    void stop_all(std::unordered_map<Name,
```
acmorrow, line 189: Huh. So the second parameter here is like an unwrapped `AttributeMap` that's not held by `shared_ptr`. Not asking you to change it, but I think the `AttributeMap` infra is something we should look at cleaning up.

src/viam/examples/camera/example_camera.cpp
```diff
@@ -44,11 +44,12 @@ int main() {
             throw;
         }
 
-        std::vector<vs::ResourceName>* resource_names = robot->resource_names();
+        std::vector<vs::Name>* resource_names = robot->resource_names();
 
         cout << "Resources of the robot:" << endl;
-        for (vs::ResourceName resource : *resource_names) {
-            cout << " - " << resource.name() << " (" << resource.subtype() << ")" << endl;
+        for (vs::Name resource : *resource_names) {
```
acmorrow, line 50: `for (const auto& resource : *resource_names)` while you are here.

src/viam/sdk/resource/resource.cpp
```diff
@@ -21,8 +21,8 @@ std::string Resource::name() const {
 
 void Resource::reconfigure(Dependencies deps, ResourceConfig cfg){};
 
-ResourceName Resource::get_resource_name(std::string name) {
-    ResourceName r;
+common::v1::ResourceName Resource::get_resource_name(std::string name) {
```
acmorrow, line 24: You can bring this in with `using` since you are in .cpp scope, if you like.

src/viam/sdk/robot/client.cpp
```diff
@@ -57,6 +47,110 @@ using viam::robot::v1::Status;
 // NOLINTNEXTLINE
 const std::string kStreamRemoved("Stream removed");
 
+viam::robot::v1::DiscoveryQuery discovery_query::to_proto() const {
```
acmorrow, line 50: @stuqdog - I'm considering whether we want some sort of traits class for proto to type interconversion. Think From/Into sort of thing. It'd let us use a uniform syntax for moving between our vocabulary types and the backing protos. I don't have syntax in mind yet, but any thoughts?

src/viam/sdk/robot/client.cpp
```diff
@@ -228,9 +322,9 @@ RobotClient::RobotClient(std::shared_ptr<ViamChannel> channel)
       should_close_channel_(false),
       stub_(RobotService::NewStub(channel_)) {}
 
-std::vector<ResourceName>* RobotClient::resource_names() {
+std::vector<Name>* RobotClient::resource_names() {
     const std::lock_guard<std::mutex> lock(lock_);
-    std::vector<ResourceName>* resources = &resource_names_;
+    std::vector<Name>* resources = &resource_names_;
```
acmorrow, line 327: Oh, this is worse than I thought. The lock is completely defeated here because the returned pointer is to internal state. This needs to return by value.

acmorrow (changes requested):
Lots of comments and ideas, but I think this is headed in a good direction. @stuqdog I have some questions here and there for you.

In some places I've left notes on existing code that looks like it needs help or can be improved. Those things don't necessarily need to be acted on to land this review but if they aren't they should probably be ticketed and `TODO`s added to the review.

src/viam/sdk/tests/test_robot.cpp
```diff
@@ -155,14 +177,20 @@ BOOST_AUTO_TEST_CASE(test_get_status) {
...
+
+        std::vector<viam::robot::v1::FrameSystemConfig> fs_config_proto;
+        for (const frameSystemConfig& f : fs_config) {
+            fs_config_proto.push_back(f.to_proto());
+        }
+
+        BOOST_TEST(
+            proto_vec_to_string_util(mock_fs_config) == proto_vec_to_string_util(fs_config_proto),
+            boost::test_tools::per_element());
+    });
```
stuqdog, line 193: So there's a kind of interesting change here (and in the other tests in this file). Previously we were just testing that, e.g., `get_frame_system_config()` worked properly. But now we're also testing the proto conversion! This is definitely good, but it seems a bit half-complete in that, while we're testing the `to_proto` method, we don't seem to be testing the `of_proto` method (at least not directly).

What I'd suggest is a separation of testing concerns here. As per an above comment, we can update the return type of the `mock_{type}_response()` methods so that no conversion is necessary here in `test_robot`. Then, we can add tests for the `{to|from}_proto` methods to the `test_types.cpp` file.

src/viam/sdk/robot/client.hpp
stuqdog: I think getting rid of this `RobotService` include involves moving much of the logic here and is probably outside the scope of this PR.

src/viam/sdk/robot/client.hpp
```diff
@@ -23,13 +25,62 @@ namespace viam {
 namespace sdk {
 
 using grpc::Channel;
-using viam::common::v1::ResourceName;
-using viam::common::v1::Transform;
-using viam::robot::v1::FrameSystemConfig;
 using viam::robot::v1::RobotService;
-using viam::robot::v1::Status;
+using time_point = std::chrono::time_point<long long, std::chrono::nanoseconds>;
```
stuqdog, line 29: This is a duplicate from `utils.hpp`, which is already being included in this file. I think we can probably get rid of it.

src/viam/sdk/robot/client.hpp
```diff
@@ -23,13 +25,62 @@ namespace viam {
...
+
+    discovery_query query;
+    AttributeMap results;
+    friend bool operator==(const discovery& lhs, const discovery& rhs);
+};
+
+struct frameSystemConfig {
+    viam::robot::v1::FrameSystemConfig to_proto() const;
+    static frameSystemConfig from_proto(const viam::robot::v1::FrameSystemConfig& proto);
+    frameSystemConfig(WorldState::transform frame) : frame(std::move(frame)){};
```
stuqdog, line 54: Should we have a constructor here that includes both `frame` and `kinematics`?

src/viam/sdk/robot/client.hpp
```diff
@@ -23,13 +25,62 @@ namespace viam {
...
 using grpc::Channel;
-using viam::common::v1::ResourceName;
-using viam::common::v1::Transform;
-using viam::robot::v1::FrameSystemConfig;
 using viam::robot::v1::RobotService;
-using viam::robot::v1::Status;
+using time_point = std::chrono::time_point<long long, std::chrono::nanoseconds>;
+
+struct discovery_query {
+    discovery_query(){};
```
stuqdog, line 32: Some of these new structs have constructors that set fields, others do not. is there a reason to not include the more comprehensive constructors here and below?

src/viam/sdk/robot/client.hpp
stuqdog, line 67: `AttributeMap` is a pointer under the hood already, so making this value `optional` on top of that seems a bit redundant and confusing. (What if it exists but is a `nullptr`?) Would suggest making this non-optional (and other optional`AttributeMap` fields below).

src/viam/sdk/robot/client.hpp
stuqdog, line 144: This could be better written as `Name r(api, "", name);`

stuqdog (changes requested):
Nice, excited to see this happen! Got some general suggestions/comments/questions but overall this definitely is looking good.

stuqdog:
> Lots of comments and ideas, but I think this is headed in a good direction. @stuqdog I have some questions here and there for you.
> 
> In some places I've left notes on existing code that looks like it needs help or can be improved. Those things don't necessarily need to be acted on to land this review but if they aren't they should probably be ticketed and `TODO`s added to the review.

👍 I'll take a look through your comments and reply on Monday!

src/viam/sdk/common/utils.cpp
```diff
@@ -21,13 +21,12 @@
 namespace viam {
 namespace sdk {
 
-using viam::common::v1::ResourceName;
 using time_point = std::chrono::time_point<long long, std::chrono::nanoseconds>;
 
-std::vector<ResourceName> resource_names_for_resource(const std::shared_ptr<Resource>& resource) {
+std::vector<Name> resource_names_for_resource(const std::shared_ptr<Resource>& resource) {
```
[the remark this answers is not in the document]
stuqdog: No good reason that I can tell! Agreed, we should move this to be a method on `Resource`.

src/viam/sdk/robot/client.hpp
```diff
@@ -23,13 +25,62 @@ namespace viam {
...
 
 using grpc::Channel;
-using viam::common::v1::ResourceName;
-using viam::common::v1::Transform;
-using viam::robot::v1::FrameSystemConfig;
 using viam::robot::v1::RobotService;
-using viam::robot::v1::Status;
+using time_point = std::chrono::time_point<long long, std::chrono::nanoseconds>;
+
+struct discovery_query {
```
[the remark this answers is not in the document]
stuqdog: Hmm... I personally am not a fan of having a `Robot` base class. In my mind there's only a single canonical implementation of how one should interact with a robot as a client in C++, and so it doesn't really make sense to expose an abstract API the same way it does for the various resources.

That said, there's something very slightly smelly about having these types exist on `RobotClient` (as opposed to `Robot`) that I'm having trouble putting my finger on. So, I'm certainly open to being convinced that an abstract `Robot` base class makes sense.

Either way, I agree that these structs should exist within `RobotClient` (or `Robot` if we create it) rather than as free-floating types.

src/viam/sdk/robot/client.hpp
stuqdog: Oh good catch! I actually think it would be great to change this to an `AttributeMap` now while we're already changing the `RobotClient` API considerably.

src/viam/sdk/robot/client.cpp
stuqdog: I like this. I don't think it's urgent but programmatically enforcing a consistent syntax and having a way of clearly expressing in code which types we expect to have proto conversions seems like a win. I'm not positive I have the look of it precisely right in my head, something like the following?
```
class ProtoConvertible {
  public:
   template <typename T, typename ProtoT>
   struct traits;

   template <typename T, typename ProtoT>
   static T from_proto(const ProtoT& proto) {
     return traits<T, ProtoT>::from_proto(proto);
   }

   template <typename T, typename ProtoT>
   static ProtoT to_proto(const T& t) const {
    return traits<T, ProtoT>::to_proto(t);
  }
...
}
```

and then define in `robot_client.hpp` 
``` 
template <>
struct ProtoConvertible::traits<frame_system_config, FrameSystemConfig> {
  static FrameSystemConfig to_proto(const frame_system_config& fsc);
  static frame_system_config to_proto(const FrameSystemConfig& proto);
}
class RobotClient {
  ...
  struct frame_system_config {
   ...
   FrameSystemConfig to_proto() const {
    return ProtoConvertible::to_proto<frame_system_config, FrameSystemConfig>(this);
   } // actual implementation should go in .cpp

   static frame_system_config from_proto(const FrameSystemConfig& proto) {
    return ProtoConvertible::from_proto<frame_system_config, FrameSystemConfig>(proto);
   } // actual implementation should go in .cpp
  }
...
}
```

And then implement all the actual logic in (for this example) `robot_client.cpp`?

Assuming I've got the gist of it right here, it seems like a bit of boilerplate but probably worthwhile for clarity and consistency.

src/viam/sdk/robot/client.hpp
stuqdog: Yikes, good catch! I think returning a `const&` or copy is totally fine. In my mind `const&` seems better because I'm not sure what value a copy provides to justify the extra cost, but I'm open to being convinced.

src/viam/sdk/robot/client.cpp
stuqdog: Not entirely certain I follow why this needs to return by value, does changing return type to `const&` not solve this? If so, then returning a copy seems totally fine.

src/viam/sdk/tests/test_robot.cpp
stuqdog: Update: I just saw your comment in the description 
> A lot of the testing was done using protos. I was getting errors when trying to compare custom-defined types so continued using protos to compare using the vec_to_string() method. This, arguably, also tests out the to_proto() methods. Should we keep as is or should we change things up? Thought there should be more conversation around this before moving ahead.

I believe my comment speaks to your question @purplenicole730 (let me know if it's not a comprehensive answer!), but just to add an extra note: if there are errors when comparing our types directly then definitely we should investigate and figure out why before merging this.

src/viam/sdk/robot/client.cpp
acmorrow: No, returning by `const&` isn't sufficient, because you could still be doing an (unlocked) read of the state via the reference at the same time a locked mutation was in progress, since the thread holding the reference doesn't hold the lock.

src/viam/sdk/robot/client.hpp
purplenicole730: I get this error: 
```
note: candidate constructor (the implicit copy constructor) not viable: no known conversion from 'std::unique_ptr<viam::robot::v1::RobotService::Stub>' to 'const viam::sdk::RobotClient::Impl' for 1st argument
struct RobotClient::Impl {
```
I had to make the stub a pointer if we wanted to use it with `RobotService::NewStub()`.
```
struct RobotClient::Impl {
    Impl(std::unique_ptr<RobotService::Stub> stub) : stub_(std::move(stub)) {}
    std::unique_ptr<RobotService::Stub> stub_;
};
```
Even so, with this change, `test_robot.cpp` is failing.

src/viam/sdk/tests/test_robot.cpp
```diff
@@ -103,7 +104,29 @@ template <typename T>
...
+    }
+    std::sort(ret.begin(), ret.end());
+    return ret;
+}
+
+std::vector<std::string> name_vec_to_string(std::vector<Name>& vec) {
+    std::vector<std::string> ret;
+    for (auto& v : vec) {
+        auto s = v.to_string();
+        ret.push_back(std::move(s));
```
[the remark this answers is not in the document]
purplenicole730: Good catch. I wrote this function before redefining the `vec_to_string_util` function.

src/viam/sdk/robot/client.hpp
purplenicole730: When I don't have this line, it gives me errors. 
```
/Users/nicolejung/viam/viam-cpp-sdk/src/viam/sdk/../../viam/sdk/robot/client.hpp:67:21: error: use of undeclared identifier 'time_point'; did you mean 'std::chrono::time_point'?
    boost::optional<time_point> last_reconfigured;
                    ^~~~~~~~~~
                    std::chrono::time_point
```

src/viam/sdk/robot/client.hpp
stuqdog: Let's chat offline about this. I'm surprised to hear tests are failing, I suspect there's probably something small to fix it but it'll be a lot easier to find with synchronous communication.

src/viam/sdk/robot/client.hpp
stuqdog: Oh whoops! My mistake, I recall the issue here. `time_point` is `typedef`ed in `utils.*cpp*` so of course it's not accessible in `client.hpp`. There was some back and forth on whether the `typedef` was worthwhile that got stalled on fixing the `time_point` type in general (the fact that the first arg is a `long long` is strange).

We should revisit and fix the `time_point` logic in general. I should've created a ticket for this a while ago; h[ere's one now](https://viam.atlassian.net/browse/RSDK-6574). In the meantime, I think we should avoid having this `using` line in the `.hpp` file entirely, and just give the full type annotation. We can still have a `using` line in the `.cpp` file for ease of use.

src/viam/sdk/robot/client.cpp
acmorrow: Yes, I think that's along the lines of what I had in mind. I filed a ticket so we don't forget - it definitely isn't work to be done in this review.

src/viam/sdk/robot/client.hpp
acmorrow: Ah, well if `RobotService::Stub` is non-copyable you can probalby just require an rvalue reference in the `Impl` ctor:

```
struct RobotClient::Impl {
    Impl(RobotService::Stub&& stub) : stub_(std::move(stub)) {}
    RobotService::Stub stub_;
};
```

If that still doesn't work for some reason then yes you will need the additional `unique_ptr`.

src/viam/sdk/robot/client.hpp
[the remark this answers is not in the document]
acmorrow: Yes I share your unease, and I find the lack of consistency between `Robot` and the other resource types a little troublesome as well. Also unclear what the use case is for `RobotService_`. I can't quite tell what it is for, which makes it hard for me to honestly evaluate whether a `Robot` base is needed. It feels like if `RobotService_` is meaningful, shouldn't be what it is serving be an implementation of `Robot`, not a `RobotClient`?

src/viam/sdk/robot/client.hpp
acmorrow: See the other thread about this on the implementation but it definitely needs to be a copy.

acmorrow (commented):
Just following up on a few comment threads, I'll take another look at this when updates have been pushed.

src/viam/sdk/tests/test_robot.cpp
```diff
@@ -129,28 +157,23 @@ BOOST_AUTO_TEST_CASE(test_get_status) {
...
-
             // ensure that we only get two of the three existing statuses
-            BOOST_CHECK_EQUAL(some_status_strs.size(), 2);
+            BOOST_CHECK_EQUAL(some_statuses.size(), 2);
 
             // unfortunately the sorting is a bit odd so we end up with a mismatch of index,
             // but this ensures that the statuses we received do exist in the mocks, as
             // expected.
-            std::vector<std::string> some_mock_strs{mock_strs[1], mock_strs[2]};
+            std::vector<RobotClient::status> some_mock_statuses{mock_statuses[0], mock_statuses[2]};
```
stuqdog, line 174: The fact that we've seen a change in the order of the statuses is a bit weird! Looking at `mock_status_response`, it looks like you caught and corrected an issue where we were setting statuses out of order (first grabbing element 1, _then_ element zero), thanks for catching that! But, we also changed the order we put the elements in the list. Can we try to create some consistency so that `some_statuses` is equal to `some_mock_statuses{mock_statuses[0], mock_statuses[1]}`? Then we can delete that weird comment above and set that whole issue to bed :)

src/viam/sdk/robot/client.hpp
stuqdog: @purplenicole730 per above comment, we should definitely update this to not return a pointer.

src/viam/sdk/robot/client.hpp
```diff
@@ -44,6 +41,58 @@ using viam::robot::v1::Status;
 /// `with_channel` require a user call to `close()`.
 class RobotClient {
    public:
+    struct discovery_query {
+        viam::robot::v1::DiscoveryQuery to_proto() const;
+        static discovery_query from_proto(const viam::robot::v1::DiscoveryQuery& proto);
```
stuqdog, line 46: It occurs to me that the goal of this PR is to remove proto types from the `robot_client` APIs, but these structs are reintroducing a proto dependency. This isn't great! I wanna think about the best way to go about resolving this... perhaps it is best to introduce a `Robot` base class after all. We can define the structs in `Robot`, and then define their `to_proto`/`from_proto` methodology in `RobotClient`. 

Another possibility: I _believe_ the only users of the `to_proto`/`from_proto` methods here are the robot tests. If that is correct (which will need to be confirmed). That means possibly we could remove those tests, or have them include `robot/client.*cpp*` rather than `robot/client.hpp` in order to access methods not exposed in the `.hpp`. Honestly both these options seem pretty bad and probably worth just discarding (especially when the `Robot` base class option exists), but I'm curious if others have thoughts or other proposals.

src/viam/sdk/robot/client.hpp
```diff
@@ -155,9 +197,10 @@ class RobotClient {
     std::shared_ptr<ViamChannel> viam_channel_;
     std::shared_ptr<Channel> channel_;
     bool should_close_channel_;
-    std::unique_ptr<RobotService::Stub> stub_;
+    struct Impl;
```
stuqdog, line 200: @acmorrow a question for you! you proposed naming this struct `Impl` but that naively seems to go against our standard naming convention for structs (that they be `snake_case`). I realize this is a bit of a unique case in the SDK (`Impl` is private while `struct`s in general are defined by their public-ness, and this is our only current instance of the `pimpl` idiom). Is it more idiomatic given these distinctions to break the normal naming convention and call it `Impl`, or keep with the struct naming convention and call it `impl`?

src/viam/sdk/robot/client.cpp
```diff
@@ -37,18 +37,8 @@
...
 using google::protobuf::RepeatedPtrField;
-using viam::common::v1::PoseInFrame;
-using viam::common::v1::ResourceName;
-using viam::common::v1::Transform;
-using viam::robot::v1::Discovery;
-using viam::robot::v1::DiscoveryQuery;
-using viam::robot::v1::FrameSystemConfig;
-using viam::robot::v1::Operation;
 using viam::robot::v1::RobotService;
-using viam::robot::v1::Status;
```
stuqdog, line 51: Not a big deal either way, but I think maintaining these `using` lines in the `.cpp` file (as opposed to the `.hpp`) is actually fine and may be worth doing if typing out the fully qualified type becomes tedious.

src/viam/sdk/robot/client.hpp
```diff
@@ -81,54 +130,50 @@ class RobotClient {
...
     /// @throws `std::runtime_error` if the requested resource doesn't exist or is the wrong type.
     /// @return a `shared_ptr` to the requested resource.
     std::shared_ptr<T> resource_by_name(std::string name) {
-        ResourceName r;
         API api = API::get<T>();
-        *r.mutable_namespace_() = api.type_namespace();
-        *r.mutable_type() = api.resource_type();
-        *r.mutable_subtype() = api.resource_subtype();
-        *r.mutable_name() = std::move(name);
+        Name r = Name(api, "", name);
```
stuqdog, line 142: (minor) I _think_ we still want to update this to `Name r(api, "", name);`. I'm a little fuzzy on some of these details but I believe what we have now uses the copy constructor (i.e., we're making a `Name`, and then copying it and assigning `r` to equal that copy), where as `Name r(api, "", name)` only constructs one `Name`. I could be wrong given that we're dealing with an rvalue `Name` though! Happy for @acmorrow to come in and correct me on this one :)

stuqdog (changes requested):
Nice, definitely looking better! A couple general comments (including one about `proto` leakage in `robot/client.hpp` that may or may not be out of scope for this PR), but I'm really liking the way this is shaping up. Thanks!

src/viam/sdk/common/utils.cpp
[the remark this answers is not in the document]
acmorrow: I think let's ticket and move on if that is OK, in the interest of landing this review.

src/viam/sdk/robot/client.hpp
acmorrow: Yes, I mentioned this in an earlier round of review too. Ultimately, this review isn't currently achieving the stated goal of removing the proto types from the `Robot` API: it has just pushed them into the `to_proto` and `from_proto` methods on the nested types.

One thing you could do is just make all of them into free functions in an unnamed namespace inside `robot.cpp`:

```
// robot.cpp
...
namespace {
viam::robot::v1::DiscoveryQuery to_proto(const RobotClient::discovery_query& dq) { ... };
discovery_query from_proto(const viam::robot::v1::DiscoveryQuery& proto);

// Repeat for `discovery`, `frame_system_config`, etc.

}  // namespace
```

This will work very well for implementing `Robot` and wouldn't force the introduction of a base class for Robot.

However, it comes with one notable downside, which is that the conversion functions are no longer available to the tests.

That's frustrating, but maybe not the end of the world. After all, the tests are perfectly able to traffic in the proto types. So if you can arrange a suitable situation via mocks, then you can test indirectly: call the API with the vocabulary type, collect the proto object in the mock, and compare. Then do the reverse.

After all, the whole point of having a class like `Robot` is to encapsulate the proto stuff away and provide a better UX to consumers.

src/viam/sdk/robot/client.hpp
acmorrow: Oh, I'm just typing from muscle memory there. Fine to call it `impl` I think.

I truly wish we could bulk convert the whole SDK to `snake_case`.

acmorrow (changes requested):
I'm in agreement with all of @stuqdog's comments and thoughts. I've shared a few of my own around the issue of the visibility of the `proto` conversion functions.

src/viam/sdk/robot/client.hpp
acmorrow: You could just write it `Name r{api, "", name}`, but I don't think you even need `r` anymore. I'd expect the following to work fine:

```
auto resource = this->resource_by_name({API::get<T>(), "", std::move(name)});
```

src/viam/sdk/common/utils.cpp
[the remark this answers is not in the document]
stuqdog: Sure, makes sense! Disregard my ask for the move @purplenicole730 :)

src/viam/sdk/tests/mocks/mock_robot.cpp
```diff
@@ -10,142 +10,122 @@ namespace sdktests {
...
-    *discovery.mutable_query() = query;
-    *discovery.mutable_results() = results;
-    return std::vector<viam::robot::v1::Discovery>{discovery};
+    AttributeMap results =
+        std::make_shared<std::unordered_map<std::string, std::shared_ptr<ProtoType>>>();
+    google::protobuf::Value v;
+    *v.mutable_string_value() = "bar";
+    google::protobuf::MapPair<std::string, std::shared_ptr<ProtoType>> pair(
+        std::move("foo"), std::make_shared<ProtoType>(v));
+    results->emplace(pair);
```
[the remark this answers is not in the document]
purplenicole730: Ah, hahaha, I didn't know that you could just make a shared prototype pointer with the string value. I thought we had to set it as a Value with a string value first like in other use cases. I was following other code examples in making a pair first as well. Thanks for letting me know!

src/viam/sdk/tests/test_types.cpp
```diff
@@ -111,6 +114,69 @@ BOOST_AUTO_TEST_CASE(test_prototype_map_conversion) {
...
+pose default_pose(int offset = 0) {
+    pose pose;
+    pose.coordinates.x = 1 + offset;
+    pose.coordinates.y = 2 + offset;
+    pose.coordinates.z = 3 + offset;
+    pose.orientation.o_x = 2 + offset;
+    pose.orientation.o_y = 3 + offset;
+    pose.orientation.o_z = 4 + offset;
+    pose.theta = 20 + offset;
+    return pose;
```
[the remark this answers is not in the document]
purplenicole730: I was going to ask about this. For some reason, using `robot::mock_function()` works just fine, but calling `robot::default_pose()` does not and gives an error. I'll try again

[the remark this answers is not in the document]
purplenicole730:
```
error: no member named 'default_pose' in namespace 'viam::sdktests::robot'
    pose pose = robot::default_pose();
                ~~~~~~~^
```

src/viam/sdk/tests/test_robot.cpp
```diff
@@ -103,7 +110,29 @@ template <typename T>
 std::vector<std::string> vec_to_string_util(std::vector<T>& vec) {
     std::vector<std::string> ret;
     for (auto& v : vec) {
-        ret.push_back(v.SerializeAsString());
+        auto s = v.to_proto().SerializeAsString();
+        ret.push_back(std::move(s));
```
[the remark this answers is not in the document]
purplenicole730: Originally, I made a new one that handled the custom defined ones, as they had to be converted to proto to use `SerializeAsString()`. This was back before I understood what the `operator==` function was.

src/viam/sdk/tests/test_robot.cpp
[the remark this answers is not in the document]
purplenicole730: I was trying to get this to work before, but I couldn't figure out why it wasn't letting me compare the vecs of `Name` directly. It gives me error messages, even though I think I'm calling it the same way I did for every other custom defined type.

src/viam/sdk/robot/client.cpp
```diff
@@ -37,18 +37,8 @@
...
-
 using google::protobuf::RepeatedPtrField;
-using viam::common::v1::PoseInFrame;
-using viam::common::v1::ResourceName;
-using viam::common::v1::Transform;
-using viam::robot::v1::Discovery;
-using viam::robot::v1::DiscoveryQuery;
-using viam::robot::v1::FrameSystemConfig;
-using viam::robot::v1::Operation;
 using viam::robot::v1::RobotService;
```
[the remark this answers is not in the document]
purplenicole730: Yes, we call it quite often, both to call the `RobotService::NewStub` function, or just as a type `RobotService::Stub`.

src/viam/sdk/robot/client.cpp
```diff
@@ -57,6 +47,147 @@ using viam::robot::v1::Status;
...
+    return op;
+}
+
+bool operator==(const RobotClient::operation& lhs, const RobotClient::operation& rhs) {
+    return lhs.id == rhs.id && lhs.method == rhs.method && lhs.session_id == rhs.session_id &&
+           lhs.arguments == rhs.arguments && lhs.started == rhs.started;
+}
+
+struct RobotClient::Impl {
+    Impl(std::unique_ptr<RobotService::Stub> stub) : stub_(std::move(stub)) {}
```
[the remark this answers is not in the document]
purplenicole730: Yes, after making it so that `stub_` is no longer a pointer, it then gives me an error like so:
```
note: in instantiation of function template specialization 'std::make_unique<viam::sdk::RobotClient::impl, std::unique_ptr<viam::robot::v1::RobotService::Stub>>' requested here
      impl_(std::make_unique<impl>(RobotService::NewStub(channel_))) {}
                 ^
```
called when we try to call: 
```
RobotClient::RobotClient(std::shared_ptr<ViamChannel> channel)
    : viam_channel_(channel),
      channel_(channel->channel()),
      should_close_channel_(false),
      impl_(std::make_unique<impl>(RobotService::NewStub(channel_))) {}
```

src/viam/examples/camera/example_camera.cpp
```diff
@@ -44,11 +44,12 @@ int main() {
             throw;
         }
 
-        std::vector<vs::ResourceName>* resource_names = robot->resource_names();
+        std::vector<vs::Name>* resource_names = robot->resource_names();
```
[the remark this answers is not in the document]
purplenicole730: Does this mean simply returning a `std::vector<Name>`?

src/viam/sdk/robot/client.hpp
purplenicole730: So the move seems to be to make a base class `Robot` in a new `robot.cpp` file. As for testing, `test_robot` doesn't use the protos, but `test_types` does.

@acmorrow When you say "collect the proto object in the mock, and compare", do you mean having `test_types` include the `robot.cpp` file that has the base `Robot` class? I changed the mocks so that they return the custom defined types, but I can easily define them in `test_type`.

src/viam/sdk/tests/test_types.cpp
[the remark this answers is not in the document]
stuqdog: Yeah, that's because we don't expose `default_pose()` in `mock_robot.hpp`. You can add it to the header file and then it will be usable elsewhere.

src/viam/sdk/robot/client.hpp
acmorrow: No, let's stay away from `#include`ing `.cpp` files: it is a quick way to end up with multiply defined symbols.

Instead, my suggestion is to make all of the proto conversion functions private to the `robot.cpp` file. Tests will no longer be able to explicitly test them.

Instead, use the "mock pipeline" approach to connect the `RobotClient` to a mock. Then, inside the mock, you can validate that the vocabulary types work correctly end-to-end without ever using the proto types directly: that'll happen only in the `RobotClient` code. So the testing will still happen, but it will be indirect.

@stuqdog - Does that make sense to you?

src/viam/examples/dial/example_dial.cpp
```diff
@@ -55,7 +54,7 @@ int main() {
     }
 
     // ensure we can send requests for specific resources
-    std::vector<Name> just_one = {resource_names->at(0)};
+    std::vector<Name> just_one = {resource_names[0]};
```
acmorrow, line 57: Just FYI, but there is an important difference between `std::vector::at` and `std::vector::operator[]`: the former is bounds checked but the latter is not. I'm not sure if the old code intended to call the bounds checked version or was just working around the awkwardness of `(*resource_names)[0]`.

src/viam/sdk/common/utils.cpp
```diff
@@ -23,6 +23,7 @@ namespace sdk {
 
 using time_point = std::chrono::time_point<long long, std::chrono::nanoseconds>;
 
+// TODO: RSDK-6627 : move this function to `Resource` class.
```
acmorrow, line 26: 👍

src/viam/sdk/robot/client.cpp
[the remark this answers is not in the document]
acmorrow: Apologies, I meant to leave that comment on the `client.hpp` header, not the `client.cpp` file. I will re-add the comment in the right place.

src/viam/sdk/robot/client.hpp
acmorrow: I believe you can now remove `using viam::robot::v1::RobotService;` from this header.

src/viam/sdk/robot/client.hpp
stuqdog: Yep, completely agree.

src/viam/sdk/robot/client.cpp
```diff
@@ -357,11 +362,11 @@ RobotClient::RobotClient(std::shared_ptr<ViamChannel> channel)
...
       channel_(channel->channel()),
       should_close_channel_(false),
-      impl_(std::make_unique<Impl>(RobotService::NewStub(channel_))) {}
+      impl_(std::make_unique<impl>(RobotService::NewStub(channel_))) {}
 
-std::vector<Name>* RobotClient::resource_names() {
+std::vector<Name> RobotClient::resource_names() {
     const std::lock_guard<std::mutex> lock(lock_);
-    std::vector<Name>* resources = std::move(&resource_names_);
+    std::vector<Name> resources = std::move(resource_names_);
```
acmorrow, line 369: I think this is actually going to destroy the contents of `resource_names_` since they will be moved into `resources` and then returned. I think you can get rid of this line entirely and now write `return resource_names_` which will implicitly copy. It actually wouldn't have been possible to make this error if `RobotClient::resource_names` was declared `const`, which I think it probably should be. Can you please make it `const`? Note that you will probably need to declare `lock_` as mutable to do so. That's fine, it is very common for private `mutex` members to be mutable: https://stackoverflow.com/questions/14131572/always-declare-stdmutex-as-mutable-in-c11

src/viam/sdk/robot/client.cpp
```diff
@@ -398,8 +403,7 @@ std::shared_ptr<RobotClient> RobotClient::at_local_socket(std::string address, O
     const char* uri = addr.c_str();
     const std::shared_ptr<grpc::Channel> channel =
         grpc::CreateChannel(uri, grpc::InsecureChannelCredentials());
-    const std::unique_ptr<viam::robot::v1::RobotService::Stub> st =
-        viam::robot::v1::RobotService::NewStub(channel);
+    const std::unique_ptr<RobotService::Stub> st = RobotService::NewStub(channel);
```
acmorrow, line 406: Is this `st` even used?

src/viam/sdk/robot/client.hpp
```diff
@@ -138,10 +138,7 @@ class RobotClient {
     /// @throws `std::runtime_error` if the requested resource doesn't exist or is the wrong type.
     /// @return a `shared_ptr` to the requested resource.
     std::shared_ptr<T> resource_by_name(std::string name) {
-        API api = API::get<T>();
-        Name r = Name(api, "", name);
-
-        auto resource = this->resource_by_name(std::move(r));
+        auto resource = this->resource_by_name({API::get<T>(), "", name});
```
acmorrow, line 141: I don't think you need the `this->` here. And the function is now simple enough it could be a one liner:

```
std::shared_ptr<T> resource_by_name(std::string name) {
    return std::dynamic_pointer_cast<T>(resource_by_name({API::get<T>(), "", std::move(name)}));
}
```

Note that `name` should be moved from.

src/viam/sdk/robot/client.hpp
```diff
@@ -81,54 +130,47 @@ class RobotClient {
...
         return std::dynamic_pointer_cast<T>(resource);
     }
 
     /// @brief Get the configuration of the frame system of the given robot.
     /// @return The configuration of the calling robot's frame system.
-    std::vector<FrameSystemConfig> get_frame_system_config(
-        std::vector<Transform> additional_transforms = std::vector<Transform>());
+    std::vector<frame_system_config> get_frame_system_config(
+        std::vector<WorldState::transform> additional_transforms =
+            std::vector<WorldState::transform>());
```
acmorrow, line 149: You can make this simpler: `std::vector<WorldState::transform> additional_transforms = {}`

src/viam/sdk/robot/client.hpp
```diff
@@ -81,54 +130,47 @@ class RobotClient {
...
-    /// @return the `PoseInFrame` of the transformed pose.
-    viam::common::v1::PoseInFrame transform_pose(
-        viam::common::v1::PoseInFrame query,
-        std::string destination,
-        std::vector<Transform> additional_transforms = std::vector<Transform>());
+    /// @return the `pose_in_frame` of the transformed pose.
+    pose_in_frame transform_pose(pose_in_frame query,
+                                 std::string destination,
+                                 std::vector<WorldState::transform> additional_transforms =
+                                     std::vector<WorldState::transform>());
```
acmorrow, line 173: Ditto as `= {}` for the defaulted parameter.

src/viam/sdk/robot/client.cpp
```diff
@@ -57,6 +53,146 @@ using viam::robot::v1::Status;
 // NOLINTNEXTLINE
 const std::string kStreamRemoved("Stream removed");
 
+// TODO: add a traits class for proto to type and back conversion
+DiscoveryQuery RobotClient::discovery_query::to_proto() const {
+    DiscoveryQuery proto;
+    *proto.mutable_subtype() = subtype;
+    *proto.mutable_model() = std::move(model);
```
acmorrow, line 60: This `std::move` probably doesn't do anything: the method is `const` so `model` is const, so can't be moved from.

src/viam/sdk/robot/client.cpp
```diff
@@ -57,6 +53,146 @@ using viam::robot::v1::Status;
...
+
+RobotClient::discovery RobotClient::discovery::from_proto(const Discovery& proto) {
+    discovery discovery;
+    discovery.query = discovery_query::from_proto(proto.query());
+    discovery.results = struct_to_map(proto.results());
+    return discovery;
+}
+
+bool operator==(const RobotClient::discovery& lhs, const RobotClient::discovery& rhs) {
+    return lhs.query == rhs.query && map_to_struct(lhs.results).SerializeAsString() ==
```
acmorrow, line 90: Why do `results` get turned back into `proto Struct`'s in order to compare them with `SerializeAsString`? I'd expect `(*lhs.results == *rhs.results)` to work, since then you are comparing unordered maps.

Ohhh.... I think I see. It won't work because it will end up comparing the embedded `shared_ptr<ProtoType>` rather than the pointed-to `ProtoType`. That's unfortunate.

I think we should prioritize fixing things up around `AttributeMap` / `ProtoType`. I think if we do that we will end up back here because I'd expect to see `map_to_struct` and `struct_to_map` change.

OK to go ahead with what you have here unless @stuqdog disagrees.

acmorrow (changes requested):
This is coming along well.

Please take care of the items I've called out here, and then I think it is time to tackle moving the `to_proto` and `from_proto` methods out of the header and refactoring the testing to use a mock pipeline per discussion.

src/viam/examples/dial/example_dial.cpp
stuqdog: I don't think there was a strong desire for the bounds checking particularly, especially in example code. This looks fine to me.

src/viam/sdk/robot/client.cpp
stuqdog: No disagreement here. Agreed also that `AttributeMap` could use another look. Do you recall offhand if we made a ticket for that already?

src/viam/sdk/robot/client.cpp
purplenicole730: I guess not. Not sure why it's there. Removed

src/viam/sdk/robot/client.cpp
acmorrow: The `mutable` should go on the declaration of `RobotClient::lock_`.

src/viam/sdk/robot/client.hpp
stuqdog: Actually to clarify: I think the proto conversion functions can't be just private in `robot.cpp` if `Robot` is an abstract base class. Instead I think they should be private in `robot/client.cpp` where they are actually being used.

src/viam/sdk/robot/client.cpp
acmorrow: No, I don't think we have. Could you make one?

stuqdog: Done. https://viam.atlassian.net/browse/RSDK-6636

src/viam/sdk/robot/client.hpp
stuqdog: Yep agreed, a project for a future ticket.

src/viam/sdk/tests/test_robot.cpp
```diff
@@ -171,30 +218,48 @@ BOOST_AUTO_TEST_CASE(test_get_operations) {
...
+        auto proto = protos[0];
+        auto proto_results = proto.results().fields().begin();
+
+        BOOST_CHECK_EQUAL(component.query.subtype, proto.query().subtype());
+        BOOST_CHECK_EQUAL(component.query.model, proto.query().model());
+        BOOST_CHECK_EQUAL(results->first, proto_results->first);
+        std::cout << component.results->begin()->second->proto_value().number_value() << " ------ "
+                  << proto.results().fields().begin()->second.string_value() << std::endl;
+        // BOOST_CHECK_EQUAL(results->second->proto_value().string_value(),
+        //                   proto_results->second.number_value());
```
purplenicole730, line 241: I've been chasing this down for a while to get to the bottom of a test failing. I think I finally narrowed it down enough to make a little sense of what is going on.

Goal: to get `results->second` and `proto_results->second` both to return "bar".
```
        auto components = mock_discovery_response();
        auto component = components[0];
        auto results = component.results->begin();
        auto protos = mock_proto_discovery_response();
        auto proto = protos[0];
        auto proto_results = proto.results().fields().begin();
        BOOST_CHECK_EQUAL(results->second->proto_value().string_value(),
                           proto_results->second.number_value());
```
For some reason, this does not work. (note: the `test_get_operations` test case is also failing, and I'm assuming it's because of this results value.) When printing out `results->second->proto_value().string_value()`, it is empty. After doing tests, I realized that if I set the value to the number 4, it works. `results->second->proto_value().number_value()` returns 4, and all tests pass. 

The issue is with this:
```
    AttributeMap results =
        std::make_shared<std::unordered_map<std::string, std::shared_ptr<ProtoType>>>();
    results->emplace("foo", std::make_shared<ProtoType>("bar"));
```
For some reason, making the pair like this does not work, but making the pair with `std::make_shared<ProtoType>(4)` works just fine.

I'm not sure why this is the case. The only difference I can find is that in `proto_type.cpp`, I see that number and string values are set differently:
```
        case 2: {
            const std::string s = boost::get<std::string>(proto_type_);
            *v.mutable_string_value() = s;
            break;
        }
        case 3: {
            const int i = boost::get<int>(proto_type_);
            v.set_number_value(i);
            break;
        }
```

stuqdog: I have a solution for this. Short version is: we were creating a `ProtoType` not with a `std::string` but with a `const char*`, which we don't have a constructor for. The compiler was then interpreting this as a `bool`. This makes some sense (`const char*` is inherently truth-y or false-y) but also confuses me a bit because the `ProtoType` constructors are all explicit. In either case, adding a `const char*` constructor or explicitly creating the ProtoType with a `std::string({value})` solves the problem. Probably we should do both.

stuqdog (approved):
Nice! This looks great. There are a couple minor comments/requested changes I left but I don't think they're significant enough to require another round of review after being fixed. Thanks so much for putting this together!

One note, however: tests will now fail in this branch because we've made `clang-tidy` a bit stricter recently. So you'll need to make a couple changes around `const` correctness to address those. To avoid having to go through the CI cycle over and over, I suggest invoking `cmake` with the `DVIAMCPPSDK_CLANG_TIDY=ON` flag and building so you can see all places that may require fixes.

src/viam/sdk/robot/client.cpp
```diff
@@ -57,6 +53,101 @@ using viam::robot::v1::Status;
...
+           lhs.last_reconfigured == rhs.last_reconfigured;
+}
+
+bool operator==(const RobotClient::operation& lhs, const RobotClient::operation& rhs) {
+    return lhs.id == rhs.id && lhs.method == rhs.method && lhs.session_id == rhs.session_id &&
+           lhs.arguments == rhs.arguments && lhs.started == rhs.started;
+}
+
+struct RobotClient::impl {
+    impl(std::unique_ptr<RobotService::Stub> stub) : stub_(std::move(stub)) {}
```
acmorrow, line 147: I still think you can write this as `impl(RobotService::Stub&& stub) : stub_(std::move(stub)` and get rid of the `unique_ptr`, but, at the same time, what you have here isn't wrong, it just has an extra level of indirection.

acmorrow (approved):
LGTM with one optional suggestion / thing to try. I only took a quick look at the mock / test stuff because I think Ethan has a good handle on what is going on there.

This was quite a review but I think the final result looks great, with a much cleaner RobotClient API!

src/viam/sdk/robot/client.cpp
stuqdog: fwiw I get errors when I try to make this change since the `RobotService::NewStub` method returns a `unique_ptr` and not a raw `Stub` type. I can get rid of the `unique_ptr` by dereferencing it in the constructor but I'm not sure that's better.

purplenicole730: I tried this twice before and it gave me errors even though I changed everything. It's a little hard to find the two conversations we had on this since I was asked not to resolve the comments.
I'm guessing it's because of what Ethan just said.

PR #208, opened 2024-01-31
RSDK-3030 - Refactor registry

stuqdog:
****High Level Changes****

- Remove `init` method for registering resources pre-main. Instead, call a `Registry::initialize()` method to handle all registration.
- Simplify `ResourceRegistration2`, move it into a `register_resource` method.
- Move all resources onto `ResourceRegistration2`, removing the need to create unique `{ResourceType}Registration` classes.
- Ensure that resource servers are properly registered when creating a `RobotService` such that resources acquired from that robot service can actually make gRPC calls.
  - In order to accomplish this, I moved management of `ResourceServers` onto the `Server` class. I considered creating a top level `Viam` class (which would own both the servers and the `initialize()` method currently on `Registry`), but decided against this because it would then need to be passed around to various methods and thus would complicate user-facing APIs.
  - this in turn led to a refactor of the `ModuleService`. Previously it was managing services by itself but this ended up being pretty duplicative of the new `server`-based ownership module, so I moved it over to just looking to the `Server` to manage everything.


src/viam/examples/modules/complex/main.cpp
```diff
@@ -28,33 +28,32 @@
...
 
 int main(int argc, char** argv) {
+    Registry::initialize();
     API base_api = API::get<Base>();
+    API gizmo_api = API::get<Gizmo>();
+    API summation_api = API::get<Summation>();
     Model mybase_model("viam", "base", "mybase");
 
+    // // CR erodkin: add note about how we've reduced the comment here.
+    // Make sure to explicity register resources with custom APIs.
```
stuqdog, line 38: Removed the "must be done in `main` and not in resource implementation" bit of this comment because the idea of doing registration pre-main doesn't really make sense anymore.

src/viam/sdk/robot/service.hpp
```diff
@@ -49,6 +49,8 @@ class RobotService_ : public ResourceServer, public viam::robot::v1::RobotServic
                            ::viam::robot::v1::StopAllResponse* response) override;
 
    private:
+    // CR erodkin: this is a bit of a hacky workaround, leave a comment about it
+    API api() const override;
```
stuqdog, line 53: We added the ability to look up `API` for a `ResourceServer` so that we can register them properly within the server. Unfortunately the `RobotService` is technically a `ResourceServer`! It doesn't really make sense for it to have an API that people can lookup so I just threw it in private to try and hide it. Curious if folks have thoughts on a better approach.

src/viam/sdk/rpc/server.hpp
```diff
@@ -28,13 +32,23 @@ class Server {
     /// repeated calls.
     void start();
 
-    // TODO: make `register_service` take one of our types as an arg rather than a
-    // grpc service type, and convert under the hood
```
stuqdog, line 32: I tried to do this but ran into weird issues. We can't just go to registering a `ResourceServer` type unless we make it clear under the hood that it's a `grpc::Service`. But, we can't have the `ResourceServer` base type inherit from `grpc::Service` because, e.g., `MotorServer` needs to inherit from `MotorService`, and then we get an ambiguous type hierarchy inference when trying to register. We _could_ solve this by adding a `to_grpc_service` method that does appropriate typecasting or something like that, but this is pretty hacky and I don't love it. Add to this the fact that users (even those implementing new services) should never have to touch this method and it seemed not worth looking into this todo anymore.

Happy to revert if folks disagree.

src/viam/sdk/module/service.cpp
```diff
@@ -263,28 +262,9 @@ ModuleService::~ModuleService() {
     }
 }
 
-void ModuleService::add_api_from_registry_inlock_(API api, const std::lock_guard<std::mutex>&) {
```
stuqdog, line 266: This was adding managed resource servers/resources from the registry to the module_. But, since all resource servers are now owned/maintained by the `Server`, this is no longer necessary.

src/viam/sdk/registry/registry.cpp
```diff
@@ -72,12 +123,24 @@ std::shared_ptr<ResourceRegistration> Registry::lookup_resource(API api) {
     return apis_.at(api);
 }
 
-std::unordered_map<std::string, std::shared_ptr<ModelRegistration>> Registry::registered_models() {
-    std::unordered_map<std::string, std::shared_ptr<ModelRegistration>> registry;
-    for (auto& resource : resources_) {
-        registry.emplace(resource.first, resource.second);
+const google::protobuf::ServiceDescriptor* Registry::get_service_descriptor_(
```
stuqdog, line 126: Just moving some logic from the header to the implementation file.

src/viam/sdk/registry/registry.cpp
```diff
@@ -92,8 +155,65 @@ const google::protobuf::ServiceDescriptor* ResourceRegistration::service_descrip
     return service_descriptor_;
 }
 
+void register_resources() {
```
stuqdog, line 158: A bit of a gotcha for Viam developers, we'll now need to add new resources here. It should be hard to miss; `client_to_mock_pipeline` handles `Registry` initialization and so tests will fail if resources aren't registered.

src/viam/sdk/rpc/server.cpp
```diff
@@ -1,39 +1,66 @@
...
 
+#include <viam/sdk/registry/registry.hpp>
+
 namespace viam {
 namespace sdk {
 
-Server::Server() : builder_(std::make_unique<grpc::ServerBuilder>()) {}
+Server::Server() : builder_(std::make_unique<grpc::ServerBuilder>()) {
+    auto new_manager = std::make_shared<ResourceManager>();
+    for (const auto& rr : Registry::registered_resources()) {
```
stuqdog, line 13: Register all known resources when creating the server, to ensure proper gRPC reflection.

src/viam/sdk/rpc/server.cpp
```diff
@@ -1,39 +1,69 @@
...
 #include <grpcpp/security/server_credentials.h>
 
+#include <viam/sdk/registry/registry.hpp>
+
 namespace viam {
 namespace sdk {
 
-Server::Server() : builder_(std::make_unique<grpc::ServerBuilder>()) {}
+Server::Server() : builder_(std::make_unique<grpc::ServerBuilder>()) {
+    Registry::initialize();
```
stuqdog, line 14: I don't like having this logic (Registry initializing, creating of managed resource servers) happening inside the `Server` constructor. But, it does have the nice property of creating only a single point where all this is handled, and not requiring users to call initialize themselves (which could be a bit of a gotcha).

I tried moving this logic into the `Server::start()` call where I think it makes a bit more sense, but this created a problem where it was impossible to add a managed resource until the server had already started. This made it impossible, e.g., for the `RobotService` to make sure that all _its_ managed resources were taken care of at construction. Given that this is relatively hidden from users (they should never have to deal with registration at this level, or deal with the look of this implementation), I thought the usefulness properties and lack of user-facing issues outweighed the slight code-smell that comes from it. Happy to revisit if others disagree.

src/viam/sdk/rpc/server.cpp
```diff
@@ -1,39 +1,69 @@
...
+    for (const auto& rr : Registry::registered_resources()) {
+        auto new_manager = std::make_shared<ResourceManager>();
+        auto server = rr.second->create_resource_server(new_manager, *this);
+        managed_servers_.emplace(server->api(), std::move(server));
```
stuqdog, line 18: Register all known resources when creating the server, to ensure proper gRPC reflection.

I don't like having this logic (Registry initializing, creating of managed resource servers) happening inside the Server constructor. But, it does have the nice property of creating only a single point where all this is handled, and not requiring users to call initialize themselves (which could be a bit of a gotcha).

I tried moving this logic into the `Server::start()` call where I think it makes a bit more sense, but this created a problem where it was impossible to add a managed resource until the server had already started. This made it impossible, e.g., for the `RobotService` to make sure that all its managed resources were taken care of at construction. Given that this is relatively hidden from users (they should never have to deal with registration at this level, or deal with the look of this implementation), I thought the usefulness properties and lack of user-facing issues outweighed the slight code-smell that comes from it. Happy to revisit if others disagree.

acmorrow (commented):
@stuqdog - I'm starting into this review today. I've just taken a quick skim through so far. Is it fair to say that as far as the changes under `src/viam/sdk/{components,services}/*` that it is more or less identical changes for each of `base`, `board`, etc.? If so that will save me some time reviewing as I can take a close look at the changes for one component and one service and basically mark-as-read the others. It will be easier to see the real changes that way.

stuqdog:
> @stuqdog - I'm starting into this review today. I've just taken a quick skim through so far. Is it fair to say that as far as the changes under `src/viam/sdk/{components,services}/*` that it is more or less identical changes for each of `base`, `board`, etc.? If so that will save me some time reviewing as I can take a close look at the changes for one component and one service and basically mark-as-read the others. It will be easier to see the real changes that way.

@acmorrow hi yes, that is a fair assessment!

src/viam/examples/modules/complex/main.cpp
```diff
@@ -29,32 +29,28 @@ using namespace viam::sdk;
 
 int main(int argc, char** argv) {
     API base_api = API::get<Base>();
+    API gizmo_api = API::get<Gizmo>();
+    API summation_api = API::get<Summation>();
     Model mybase_model("viam", "base", "mybase");
 
+    // Make sure to explicity register resources with custom APIs.
+    Registry::register_resource<GizmoClient, GizmoServer, GizmoService>(gizmo_api);
+    Registry::register_resource<SummationClient, SummationServer, SummationService>(summation_api);
```
acmorrow, line 38: I find it a little odd that in `complex/client`, the registration must make reference to `[Gizmo,Summation]Server` despite not really needing anything to do with servers, and that in the module main that the registration must make reference to `[Gizmo,Summation]Client`, which it will presumably never use.

src/viam/sdk/resource/resource_server_base.hpp
```diff
@@ -1,14 +1,15 @@
...
 
 #include <viam/sdk/resource/resource_manager.hpp>
-#include <viam/sdk/rpc/server.hpp>
 
 namespace viam {
 namespace sdk {
 
 class ResourceServer {
    public:
+    virtual API api() const = 0;
```
acmorrow, line 10: Why do `ResourceServer`'s need this now?

src/viam/sdk/rpc/server.cpp
acmorrow: Yeah I'm really feeling the lack of clarity around what the root of initialization and ownership is in the SDK. I don't think we are going to get there in this review.

src/viam/sdk/resource/resource_server_base.cpp
```diff
@@ -6,5 +6,9 @@ const std::shared_ptr<ResourceManager>& ResourceServer::resource_manager() const
     return manager_;
 };
 
+std::shared_ptr<ResourceManager>& ResourceServer::resource_manager() {
```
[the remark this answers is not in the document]
stuqdog: Ahh yeah, good catch. Upon reflection I'm not sure why I added this at all. We already had a `const` version, and since the inner `ResourceManager` isn't const we can still call the setter methods. Will remove.

src/viam/sdk/robot/service.hpp
stuqdog: There is no `Robot` base class, and there is no `Robot` API. Certainly not in the SDK, but (to my knowledge) not anywhere in the codebase. The `type` of an API (either `component` or `service`) breaks down with `robot`, because `component` or `service` is a resource type on the particular robot. It doesn't really make sense in my head to think of the `API` of a robot.

src/viam/sdk/components/base/base.hpp
```diff
@@ -52,9 +38,6 @@ class Base : public Component, public Stoppable {
     friend std::ostream& operator<<(std::ostream& os, const properties& v);
     friend bool operator==(const properties& lhs, const properties& rhs);
 
-    // functions shared across all components
-    static std::shared_ptr<ResourceRegistration> resource_registration();
```
acmorrow, line 56: BTW, I should add that I am happy to see the registration entry points removed from the base classes. It was a little weird that the base classes were responsible for registering the client subclasses and associated servers types.

src/viam/sdk/rpc/server.cpp
stuqdog: 😞 yeah that is fair. I suspect a second PR that focuses broadly on this specific question will be important in the near future.

src/viam/sdk/registry/registry.cpp
```diff
@@ -92,8 +155,61 @@ const google::protobuf::ServiceDescriptor* ResourceRegistration::service_descrip
...
+    // Register all components
+    Registry::register_resource<BaseClient, BaseServer, component::base::v1::BaseService>(
```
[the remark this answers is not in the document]
stuqdog: Yeah, I tried to figure out some way to do that with `boost::mpl::list` but got pretty lost in the weeds and walked away with the sense that `boost::mpl` wasn't able to do what I needed. I'm very very far from an expert on C++ meta-programming however, and more than happy to be pointed in a better direction or otherwise proven wrong!

src/viam/sdk/registry/registry.hpp
```diff
@@ -169,11 +166,24 @@ class Registry {
     /// @return A `shared_ptr` to the registered api's `ResourceRegistration`.
     static std::shared_ptr<ResourceRegistration> lookup_resource(API api);
 
-    /// @brief Provide information on registered resources.
-    /// @return A map from name to `ModelRegistration` of all registered resources.
-    static std::unordered_map<std::string, std::shared_ptr<ModelRegistration>> registered_models();
+    /// @brief Provide information on registered resource models.
+    /// @return A map from name to `ModelRegistration` of all registered resource models.
+    static const std::unordered_map<std::string, std::shared_ptr<ModelRegistration>>&
```
acmorrow, line 171: I have some concerns about these return types. First, it is returning a reference to internal state. Are there concurrency concerns? Second, `shared_ptr` is const, but the `ModelRegistration` it points to is not.

src/viam/sdk/resource/resource_server_base.hpp
stuqdog: My initial thinking was: when initializing the `server` we store all supported `ResourceServer` instances, but we want to be able to look them up so we store them by `API`. This is how we get the API from the `ResourceRegistration`.

However! I now realize that the key in `ResourceRegistration` is an `API` already 🤦 so I think we can get rid of this (and do away with the question of the `Robot` `API`) entirely.

src/viam/sdk/registry/registry.hpp
stuqdog: Ahh good shout out. I think this overlaps with your concerns on the `resource_manager` method on `ResourceServer`. I think some revisiting of why we need to expose these types at all, how they're being used, and how to avoid doing so is warranted. I'll do some digging/refactoring.

src/viam/examples/modules/complex/client.cpp
```diff
@@ -39,8 +39,9 @@ int main() {
 
     // Register custom gizmo and summation API so robot client can access resources
     // of that type from the server.
-    Registry::register_resource(API::get<Gizmo>(), Gizmo::resource_registration());
-    Registry::register_resource(API::get<Summation>(), Summation::resource_registration());
+    Registry::register_resource<GizmoClient, GizmoServer, GizmoService>(API::get<Gizmo>());
```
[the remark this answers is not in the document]
stuqdog: Yeah, I like the look of this. Is there a reason to even include `GizmoClient` in the `register_resource` call? If we're fine with knowledge of server side forcing knowledge of client side then I think we can get away with having  
```
using client_type = FooClient;
using interface_type = Foo;
```
 in the `FooServer` declaration, and then reducing down to `Registry::register_resource<FooServer>();` for registration.

src/viam/examples/modules/complex/main.cpp
stuqdog: Hmm yeah I agree. Will revisit and see if we can clean that up some.

acmorrow: It makes me wonder whether client and server registration should be separate activities.

src/viam/examples/modules/complex/client.cpp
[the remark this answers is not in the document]
acmorrow: As I mentioned in my other comment, I'm curious about whether `register_resource` is trying to do too much. Am I correct that both `FooClient` and `FooServer` do need to be registered, but the former only for client usage and the latter only in a relevant module? If that's the case then I'd be pretty happy keeping FooServer even unaware of FooClient, and just having both know the associated base type.

src/viam/sdk/registry/registry.hpp
```diff
@@ -11,84 +11,54 @@
...
 // TODO(RSDK-3030): one class per header
-/// @class ResourceRegistration registry.hpp "registry/registry.hpp"
-/// @brief Defines registered `Resource` creation functionality.
-class ResourceRegistration {
+class ResourceServerRegistration {
    public:
-    virtual ~ResourceRegistration();
-
-    /// @brief Add `Reconfigure` functionality to a resource.
-    std::function<std::shared_ptr<Resource>(std::shared_ptr<Resource>, Name)> create_reconfigurable;
```
stuqdog, line 35: This wasn't doing anything and the (theoretical) semantics of it were confusing. So I got rid of it. Per RSDK-6484, we should just use a `Reconfigurable` base class instead.

src/viam/sdk/registry/registry.hpp
```diff
@@ -11,86 +11,56 @@
...
-            p->FindServiceByName(ProtoServiceT::service_full_name());
-        if (!sd) {
-            throw std::runtime_error("Unable to get service descriptor");
-        }
-        return std::make_shared<ResourceRegT>(sd);
-    }
+    ResourceClientRegistration() = default;
 };
 
+// TODO(RSDK-3030): instead of std::functions, consider making these functions
```
benjirewis, line 62: Should we use a different ticket number if this is still a `TODO`?

src/viam/sdk/rpc/server.hpp
benjirewis: I think it's fine to just take in the `grpc::Service*`.

benjirewis (approved):
Sorry for the delay on review, but this generally LGTM! I have no big comments to add beyond Drew's.

src/viam/sdk/registry/registry.hpp
stuqdog: Ooh yes, good catch! Updated.

acmorrow (approved):
I'm approving this, more or less as is. I have a few small comments here and there but I don't think they amount to enough that we should do another round of review. I also don't think this is the end of the resource registration story, nor of the driver initialization story. I still have a lot of questions about who owns what, and it feels like there are maybe more concepts in play than are really needed. I also have some lingering concerns around thread safety and potentially racy visibility into internal state. But I'm very pleased with the reduction in boilerplate in the client and server files, and I think the API for registering clients and servers is dramatically improved.

src/viam/sdk/registry/registry.cpp
```diff
@@ -32,7 +84,7 @@ const Model& ModelRegistration::model() const {
     return model_;
 };
 
-void Registry::register_model(std::shared_ptr<ModelRegistration> resource) {
+void Registry::register_model(std::shared_ptr<const ModelRegistration> resource) {
```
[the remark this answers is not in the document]
stuqdog: Yep will do!

PR #209, opened 2024-02-01
RSDK-2758 Add an exception system

benjirewis:
[RSDK-2758](https://viam.atlassian.net/browse/RSDK-2758)

Redo of #100. Copied much of the base code from @zaporter-work's original implementation, and the goals of this PR remain basically the same as the last one:
> This provides a base `ViamException` class and a few derived classes. The purpose of these exceptions is threefold:
    1. To provide a base to build on if we wish to pursue more advanced error creation in the future.
    2. To help users easily understand if an error came from our code or not.
    3. To facilitate error-catching and recovery behavior (ex: a `DuplicateResourceException` might be handled differently than a `ResourceNotFoundException`)

Replaces all instances of `throw std::runtime_error`, `throw std::invalid_argument`, `throw std::length_error` and `throw "foo"` with `throw`s of the new exception classes. Note that none of the `throw` calls are touched in the `examples` directory, as those represent example client or module code that should not be using our internal exception system.

[RSDK-2758]: https://viam.atlassian.net/browse/RSDK-2758?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/viam/sdk/common/client_helper.hpp
+++ b/src/viam/sdk/common/client_helper.hpp
@@ -17,7 +17,7 @@ class ClientHelper {
     static void default_rsc_(RequestType&) {}
     static void default_rhc_(const ResponseType&) {}
     static void default_ehc_(const ::grpc::Status& status) {
-        throw GRPCException(status.error_message(), status.error_code());
+        throw ViamException(status.error_message(), ViamErrorCode::GRPC);
     }
 
    public:
--- a/src/viam/sdk/common/client_helper.hpp
+++ b/src/viam/sdk/common/client_helper.hpp
@@ -17,7 +17,7 @@ class ClientHelper {
     static void default_rsc_(RequestType&) {}
     static void default_rhc_(const ResponseType&) {}
     static void default_ehc_(const ::grpc::Status& status) {
-        throw ViamException(status.error_message(), ViamErrorCode::GRPC);
+        throw GRPCException(status);
     }
 
    public:
--- a/src/viam/sdk/common/client_helper.hpp
+++ b/src/viam/sdk/common/client_helper.hpp
@@ -1,6 +1,7 @@
 #pragma once
 
 #include <grpcpp/client_context.h>
+#include <viam/sdk/common/exception.hpp>
 #include <viam/sdk/common/proto_type.hpp>
 #include <viam/sdk/common/utils.hpp>
 
@@ -16,7 +17,7 @@ class ClientHelper {
     static void default_rsc_(RequestType&) {}
     static void default_rhc_(const ResponseType&) {}
     static void default_ehc_(const ::grpc::Status& status) {
-        throw std::runtime_error(status.error_message());
+        throw GRPCException(status.error_message(), status.error_code());
     }
 
    public:
--- a/src/viam/sdk/common/exception.cpp
+++ b/src/viam/sdk/common/exception.cpp
@@ -1,0 +1,39 @@
+#include <viam/sdk/common/exception.hpp>
+
+namespace viam {
+namespace sdk {
+
+ViamException::ViamException(const std::string& what)
+    : std::runtime_error("ViamException: " + what){};
+
+ViamException::ViamException(const std::string& what, const std::string& type)
+    : std::runtime_error("ViamException(" + type + "): " + what){};
+
+ViamException::~ViamException() = default;
+
+ConnectionException::ConnectionException(const std::string& what)
+    : ViamException(what, "Connection"){};
+
+DuplicateResourceException::DuplicateResourceException(const std::string& what, Name resource_name)
+    : ViamException(what, "DuplicateResource"), resource_name_(std::move(resource_name)){};
+
+Name DuplicateResourceException::resource_name() const noexcept {
+    return resource_name_;
+}
+
+GRPCException::GRPCException(const std::string& what, int grpc_error_code)
+    : ViamException(what, "gRPC"), grpc_error_code_(std::move(grpc_error_code)){};
+
+int GRPCException::error_code() const noexcept {
+    return grpc_error_code_;
+}
+
+ResourceNotFoundException::ResourceNotFoundException(const std::string& what, Name resource_name)
+    : ViamException(what, "ResourceNotFound"), resource_name_(std::move(resource_name)){};
+
+Name ResourceNotFoundException::resource_name() const noexcept {
+    return resource_name_;
+}
+
+}  // namespace sdk
+}  // namespace viam
--- a/src/viam/sdk/common/exception.cpp
+++ b/src/viam/sdk/common/exception.cpp
@@ -14,12 +14,11 @@ ViamException::~ViamException() = default;
 ConnectionException::ConnectionException(const std::string& what)
     : ViamException(what, "Connection"){};
 
-DuplicateResourceException::DuplicateResourceException(const std::string& what, Name resource_name)
-    : ViamException(what, "DuplicateResource"), resource_name_(std::move(resource_name)){};
+DuplicateRegistrationException::DuplicateRegistrationException(const std::string& what)
+    : ViamException(what, "DuplicateRegistration"){};
 
-Name DuplicateResourceException::resource_name() const noexcept {
-    return resource_name_;
-}
+DuplicateResourceException::DuplicateResourceException(const std::string& what)
+    : ViamException(what, "DuplicateResource"){};
 
 GRPCException::GRPCException(const std::string& what, int grpc_error_code)
     : ViamException(what, "gRPC"), grpc_error_code_(std::move(grpc_error_code)){};
@@ -28,12 +27,11 @@ int GRPCException::error_code() const noexcept {
     return grpc_error_code_;
 }
 
-ResourceNotFoundException::ResourceNotFoundException(const std::string& what, Name resource_name)
-    : ViamException(what, "ResourceNotFound"), resource_name_(std::move(resource_name)){};
+NotSupportedException::NotSupportedException(const std::string& what)
+    : ViamException(what, "NotSupported"){};
 
-Name ResourceNotFoundException::resource_name() const noexcept {
-    return resource_name_;
-}
+ResourceNotFoundException::ResourceNotFoundException(const std::string& what)
+    : ViamException(what, "ResourceNotFound"){};
 
 }  // namespace sdk
 }  // namespace viam
--- a/src/viam/sdk/common/exception.cpp
+++ b/src/viam/sdk/common/exception.cpp
@@ -46,7 +46,7 @@ std::string ErrorCategory::message(int ev) const {
 }
 
 std::error_condition make_error_condition(ErrorCondition e) {
-    static ErrorCategory errorCategory{};
+    static const ErrorCategory errorCategory{};
     return {static_cast<int>(e), errorCategory};
 }
 
--- a/src/viam/sdk/common/exception.cpp
+++ b/src/viam/sdk/common/exception.cpp
@@ -3,52 +3,58 @@
 namespace viam {
 namespace sdk {
 
-ViamException::ViamException(const std::string& what)
-    : std::runtime_error("ViamException: " + what), code_(ViamErrorCode::Unknown){};
+Exception::Exception(const std::string& what)
+    : std::runtime_error("ViamException: " + what), condition_(ErrorCondition::k_general){};
 
-ViamException::ViamException(const std::string& what, ViamErrorCode code)
-    : std::runtime_error("ViamException: " + what), code_(code){};
+Exception::Exception(ErrorCondition condition, const std::string& what)
+    : std::runtime_error("ViamException: " + what), condition_(condition){};
 
-ViamException::~ViamException() = default;
+Exception::~Exception() = default;
 
-std::error_code ViamException::code() const noexcept {
-    return code_;
+std::error_condition Exception::condition() const noexcept {
+    return condition_;
 };
 
-struct ViamErrorCategory : std::error_category {
+struct ErrorCategory : std::error_category {
     const char* name() const noexcept override;
     std::string message(int ev) const override;
 };
 
-const char* ViamErrorCategory::name() const noexcept {
-    return "viam";
+const char* ErrorCategory::name() const noexcept {
+    return "viam::sdk";
 }
 
-std::string ViamErrorCategory::message(int ev) const {
-    switch (static_cast<ViamErrorCode>(ev)) {
-        case ViamErrorCode::Unknown:
-            return "unknown";
-        case ViamErrorCode::Connection:
+std::string ErrorCategory::message(int ev) const {
+    switch (static_cast<ErrorCondition>(ev)) {
+        case ErrorCondition::k_general:
+            return "general";
+        case ErrorCondition::k_connection:
             return "connection establishment failure";
-        case ViamErrorCode::DuplicateRegistration:
+        case ErrorCondition::k_duplicate_registration:
             return "duplicate registration";
-        case ViamErrorCode::DuplicateResource:
+        case ErrorCondition::k_duplicate_resource:
             return "duplicate resource";
-        case ViamErrorCode::GRPC:
+        case ErrorCondition::k_grpc:
             return "gRPC";
-        case ViamErrorCode::NotSupported:
+        case ErrorCondition::k_not_supported:
             return "not supported";
-        case ViamErrorCode::ResourceNotFound:
+        case ErrorCondition::k_resource_not_found:
             return "resource not found";
         default:
             return "unknown";
     }
 }
 
-const ViamErrorCategory viamErrorCategory{};
+std::error_condition make_error_condition(ErrorCondition e) {
+    static ErrorCategory errorCategory{};
+    return {static_cast<int>(e), errorCategory};
+}
 
-std::error_code make_error_code(ViamErrorCode e) {
-    return {static_cast<int>(e), viamErrorCategory};
+GRPCException::GRPCException(grpc::Status status)
+    : Exception(ErrorCondition::k_grpc, status.error_message()), status_(std::move(status)){};
+
+grpc::Status GRPCException::status() const noexcept {
+    return status_;
 }
 
 }  // namespace sdk
--- a/src/viam/sdk/common/exception.cpp
+++ b/src/viam/sdk/common/exception.cpp
@@ -3,49 +3,44 @@
 namespace viam {
 namespace sdk {
 
-Exception::Exception(const std::string& what)
-    : std::runtime_error("ViamException: " + what), condition_(ErrorCondition::k_general){};
-
 Exception::Exception(ErrorCondition condition, const std::string& what)
-    : std::runtime_error("ViamException: " + what), condition_(condition){};
+    : std::runtime_error("viam::sdk::Exception: " + what), condition_(condition){};
+
+Exception::Exception(const std::string& what) : Exception(ErrorCondition::k_general, what){};
 
 Exception::~Exception() = default;
 
-std::error_condition Exception::condition() const noexcept {
+const std::error_condition& Exception::condition() const noexcept {
     return condition_;
 };
 
-struct ErrorCategory : std::error_category {
-    const char* name() const noexcept override;
-    std::string message(int ev) const override;
-};
-
-const char* ErrorCategory::name() const noexcept {
-    return "viam::sdk";
-}
-
-std::string ErrorCategory::message(int ev) const {
-    switch (static_cast<ErrorCondition>(ev)) {
-        case ErrorCondition::k_general:
-            return "general";
-        case ErrorCondition::k_connection:
-            return "connection establishment failure";
-        case ErrorCondition::k_duplicate_registration:
-            return "duplicate registration";
-        case ErrorCondition::k_duplicate_resource:
-            return "duplicate resource";
-        case ErrorCondition::k_grpc:
-            return "gRPC";
-        case ErrorCondition::k_not_supported:
-            return "not supported";
-        case ErrorCondition::k_resource_not_found:
-            return "resource not found";
-        default:
-            return "unknown";
-    }
-}
-
 std::error_condition make_error_condition(ErrorCondition e) {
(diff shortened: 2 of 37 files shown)


src/viam/sdk/common/client_helper.hpp
```diff
@@ -16,7 +17,7 @@ class ClientHelper {
     static void default_rsc_(RequestType&) {}
     static void default_rhc_(const ResponseType&) {}
     static void default_ehc_(const ::grpc::Status& status) {
-        throw std::runtime_error(status.error_message());
+        throw GRPCException(status.error_message(), status.error_code());
```
benjirewis, line 20: I'm curious what to do about `@throws` doc comments. Any method that uses the client helper with the `default_ehc_` can now feasibly throw a `GRPCException` here (they all _were_ able to throw a `std::runtime_error` before we just didn't document it). Should I document that on all client functions? How important is `@throws`?

src/viam/sdk/common/exception.hpp
```diff
@@ -0,0 +1,84 @@
+/// @file common/exception.hpp
```
benjirewis, line 1: The classes in this file are mostly a subset of the [error types from the Python SDK](https://github.com/viamrobotics/viam-python-sdk/blob/main/src/viam/errors.py#L1). I added both `DuplicateRegistrationException` _and_ `DuplicateResourceException` because they seemed like different cases to me.

src/viam/sdk/common/exception.hpp
```diff
@@ -0,0 +1,84 @@
...
+
+/// @class GRPCException
+/// @brief An exception thrown when a machine has sent an error back to an SDK
+/// client over gRPC.
+/// @ingroup Exception
+class GRPCException : public ViamException {
+   public:
+    explicit GRPCException(const std::string& what, int grpc_error_code);
+
+    int error_code() const noexcept;
```
benjirewis, line 60: I'm not sure how important this method nor the `grpc_error_code_` field are. I can get rid of them if we see no use case for someone catching a `GRPCException` and accessing its error code.

benjirewis (commented):
I need to add `test_exception.cpp` with some tests, but thought I'd gather feedback before continuing.

src/viam/sdk/common/exception.cpp
```diff
@@ -0,0 +1,37 @@
...
+    : ViamException(what, "Connection"){};
+
+DuplicateRegistrationException::DuplicateRegistrationException(const std::string& what)
+    : ViamException(what, "DuplicateRegistration"){};
+
+DuplicateResourceException::DuplicateResourceException(const std::string& what)
+    : ViamException(what, "DuplicateResource"){};
+
+GRPCException::GRPCException(const std::string& what, int grpc_error_code)
+    : ViamException(what, "gRPC"), grpc_error_code_(std::move(grpc_error_code)){};
```
stuqdog, line 24: Since `grpc_error_code` is an int (which doesn't have a move constructor), I think this `std::move` doesn't do anything.

src/viam/sdk/components/base/base.cpp
```diff
@@ -34,7 +34,7 @@ std::shared_ptr<ResourceRegistration> Base::resource_registration() {
     const google::protobuf::ServiceDescriptor* sd =
         p->FindServiceByName(viam::component::base::v1::BaseService::service_full_name());
     if (!sd) {
-        throw std::runtime_error("Unable to get service descriptor for the base service");
+        throw ViamException("Unable to get service descriptor for the base service");
```
stuqdog, line 37: just a heads up that all of these "Unable to get service descriptor for {resourcetype} service" errors in `{resourcetype}.cpp` are being removed in the `refactor-registry` PR so some rebase conflict will be happening!

stuqdog (approved):
(I noticed this is in draft but figured I'd take a look. If I was over-eager, apologies! Please re-request review if there's anything still coming). 

One or two minor comments but looks good! Note that I expect there will be some conflict between this PR and the refactor registry PR, so we should make sure that whichever one gets merged second properly converts over any lingering `runtime_exception`s.

acmorrow:
So, when this came up before (per @zaporter-work), and again now, it sends me down a deep rabbit hole of studying the gRPC error model and [error_details](https://github.com/googleapis/googleapis/blob/master/google/rpc/error_details.proto) and [C++ system_error](https://en.cppreference.com/w/cpp/header/system_error) subsystem and thinking about `std::error_condition` and `std::error_category`, and `std::expected` too, and how it should all fit together. And then I stall out because I'm worried it is all too complex and we shouldn't bother.

Do we actually *need* sub-classes of `ViamException`? Or should there just be one which carries a `std::error_condition` with it? Should we create a `std::error_category` for gRPC errors? For ours? Could we use `error_details` to transmit category information from server to client? That might allow the client side to actually throw the server side error with fidelity. Could `generic` components and services have their own error spaces?

Getting the error handling model for the SDK right is very important, because once one is adopted it is more or less impossible to change it later. On the other hand, our current informal setup is less than ideal, and maybe we shouldn't let the great be the enemy of the good, etc.

stuqdog:
> So, when this came up before (per @zaporter-work), and again now, it sends me down a deep rabbit hole of studying the gRPC error model and [error_details](https://github.com/googleapis/googleapis/blob/master/google/rpc/error_details.proto) and [C++ system_error](https://en.cppreference.com/w/cpp/header/system_error) subsystem and thinking about `std::error_condition` and `std::error_category`, and `std::expected` too, and how it should all fit together. And then I stall out because I'm worried it is all too complex and we shouldn't bother.
> 
> Do we actually _need_ sub-classes of `ViamException`? Or should there just be one which carries a `std::error_condition` with it? Should we create a `std::error_category` for gRPC errors? For ours? Could we use `error_details` to transmit category information from server to client? That might allow the client side to actually throw the server side error with fidelity. Could `generic` components and services have their own error spaces?
> 
> Getting the error handling model for the SDK right is very important, because once one is adopted it is more or less impossible to change it later. On the other hand, our current informal setup is less than ideal, and maybe we shouldn't let the great be the enemy of the good, etc.

I am definitely not an expert on this and a lot of your concerns hadn't even occurred to me so happy to be corrected, but: it seems to me we can accomplish pretty much everything we have here by carrying forward the `std::error_condition` and having an `enum` of the common error cases to match on for when we care about what kind of error we received. At a minimum it offers extensibility and specificity without creating child class proliferation.

src/viam/sdk/common/client_helper.hpp
benjirewis: Sounds good; there are some `throws` doc comments I should still fix, but I'll handle that after we agree on the general structure (codes vs. conditions vs. child classes).

src/viam/sdk/common/exception.hpp
benjirewis: After switching to using `std::error_code`, I've lost the gRPC error code and only maintained the gRPC status message. I can re-add `grpc_error_code_` as an optional field on `ViamException` that's only present if it's a gRPC exception if desired?

benjirewis:
I switched to a combination of `std::error_code` and `std::error_category` and got rid of the proliferation of child classes. Let me know what you think! I didn't really know what I was doing and used [this article](https://akrzemi1.wordpress.com/examples/error_condition-example/) for reference.

src/viam/sdk/common/exception.hpp
```diff
@@ -0,0 +1,52 @@
...
+    NotSupported,           // Behavior not supported by the SDK
+    ResourceNotFound        // Resource does not exist
+};
+
+std::error_code make_error_code(ViamErrorCode e);
+
+/// @class ViamException
+/// @brief Defines an exception type for the SDK.
+/// @ingroup Exception
+class ViamException : public std::runtime_error {
```
acmorrow, line 34: You could probably just call this `Exception`, since we are already in `viam::sdk`.

src/viam/sdk/common/exception.hpp
```diff
@@ -0,0 +1,52 @@
...
+    Unknown = 0,            // Default code
+    Connection,             // Issue during connection establishment
+    DuplicateRegistration,  // API or API/Model pair has already been registered
+    DuplicateResource,      // Resource has already been added
+    GRPC,                   // gRPC error from remote machine
+    NotSupported,           // Behavior not supported by the SDK
+    ResourceNotFound        // Resource does not exist
+};
+
+std::error_code make_error_code(ViamErrorCode e);
```
acmorrow, line 29: `std::error_code` or `std::error_condition`? I always struggle with the distinction.

src/viam/sdk/common/exception.hpp
```diff
@@ -0,0 +1,52 @@
...
+namespace viam {
+namespace sdk {
+
+/// @defgroup Exception Classes related to SDK exceptions.
+
+/// @class ViamErrorCode
+/// @brief Defines a set of a error codes to be used in conjunction with
+/// ViamException.
+/// @ingroup Exception
+enum class ViamErrorCode {
```
acmorrow, line 19: Let's give this a defined width.

src/viam/sdk/common/exception.hpp
```diff
@@ -0,0 +1,52 @@
...
+    Unknown = 0,            // Default code
```
acmorrow, line 20: I think for constants we would have `k_unknown`, `k_connection`, etc.

src/viam/sdk/common/exception.hpp
acmorrow, line 19: You probably don't need `Viam` on the name.

src/viam/sdk/common/exception.hpp
```diff
@@ -0,0 +1,52 @@
...
+
+std::error_code make_error_code(ViamErrorCode e);
+
+/// @class ViamException
+/// @brief Defines an exception type for the SDK.
+/// @ingroup Exception
+class ViamException : public std::runtime_error {
+   public:
+    explicit ViamException(const std::string& what);
+    explicit ViamException(const std::string& what, ViamErrorCode code);
```
acmorrow, line 37: I think the code is more fundamental, so I think it should be the first argument:

src/viam/sdk/components/board/client.cpp
```diff
@@ -101,7 +102,7 @@ Board::analog_value BoardClient::read_analog(const std::string& analog_reader_na
 
     const grpc::Status status = stub_->ReadAnalogReader(ctx, request, &response);
     if (!status.ok()) {
-        throw std::runtime_error(status.error_message());
+        throw ViamException(status.error_message(), ViamErrorCode::GRPC);
```
acmorrow, line 105: I'd love to come up with a scheme whereby we could entirely preserve the `grpc::Status` object into the exception.

src/viam/sdk/common/exception.hpp
```diff
@@ -0,0 +1,52 @@
...
+/// @class ViamErrorCode
+/// @brief Defines a set of a error codes to be used in conjunction with
+/// ViamException.
+/// @ingroup Exception
+enum class ViamErrorCode {
+    Unknown = 0,            // Default code
+    Connection,             // Issue during connection establishment
+    DuplicateRegistration,  // API or API/Model pair has already been registered
+    DuplicateResource,      // Resource has already been added
+    GRPC,                   // gRPC error from remote machine
```
acmorrow, line 24: If we formulated an error category for gRPC's error codes, could we do more?

src/viam/sdk/common/exception.hpp
acmorrow, line 20: Will we always need `Unknown` (or `k_unknown`) or will we eventually mandate that all exceptions have a known error code / condition?

src/viam/sdk/common/exception.cpp
```diff
@@ -0,0 +1,55 @@
...
+    return code_;
+};
+
+struct ViamErrorCategory : std::error_category {
+    const char* name() const noexcept override;
+    std::string message(int ev) const override;
+};
+
+const char* ViamErrorCategory::name() const noexcept {
+    return "viam";
```
acmorrow, line 24: Maybe `viam::sdk`?

src/viam/sdk/common/exception.cpp
```diff
@@ -0,0 +1,55 @@
...
+ViamException::ViamException(const std::string& what, ViamErrorCode code)
+    : std::runtime_error("ViamException: " + what), code_(code){};
+
+ViamException::~ViamException() = default;
+
+std::error_code ViamException::code() const noexcept {
+    return code_;
+};
+
+struct ViamErrorCategory : std::error_category {
```
acmorrow, line 18: Let's put this in the unnamed namespace - it doesn't need linkage as far as I can see.

src/viam/sdk/common/exception.cpp
```diff
@@ -0,0 +1,55 @@
...
+            return "resource not found";
+        default:
+            return "unknown";
+    }
+}
+
+const ViamErrorCategory viamErrorCategory{};
+
+std::error_code make_error_code(ViamErrorCode e) {
+    return {static_cast<int>(e), viamErrorCategory};
```
acmorrow, line 51: Maybe pull the static error category in here as a thread-safe function-scoped static:

```
std::error_code make_error_code(ViamErrorCode e) {
    static ViamErrorCategory category{};
    return {static_cast<int>(e), category);
}
```

acmorrow (changes requested):
I like where this is headed. My biggest questions are around whether this should all be in terms of `error_condition` rather than `error_code`, and whether we can do more with gRPC errors. I suspect two of my comments on that front are somewhat in tension, but that's mostly because I'm not yet 100% sure what a good solution should look like or how it should behave.

src/viam/sdk/common/exception.hpp
stuqdog: Hmm... I think it's a bit clunkier in the case where we don't have a `GRPCException` type specifically. Perhaps we could `string`ify the entire `grpc::Status` and pass that as the `what`? I want to think about it a bit but I'm a bit hesitant on the optional "only-relevant-if-grpc" field.

src/viam/sdk/common/exception.cpp
benjirewis: Sure!

src/viam/sdk/common/exception.hpp
benjirewis: So I decided to leave the `k_grpc` error condition enum but made a new `GRPCException` class that is a child class of `Exception` and carries a private `grpc::Status` field. Let me know what you think.

src/viam/sdk/common/exception.hpp
benjirewis: I _think_ `std::error_condition` is probably better here. The [distinction](https://en.cppreference.com/w/cpp/error/error_condition) seems to be that `std::error_code` is "platform-dependent" while `std::error_condition` is "platform-independent". I believe we want the latter.

src/viam/sdk/common/exception.hpp
benjirewis: Done.

src/viam/sdk/common/exception.hpp
benjirewis: Move the `ErrorCondition` to be the first argument and also moved this constructor above the other.

src/viam/sdk/components/board/client.cpp
benjirewis: See what I did with the `GRPCException` class.

src/viam/sdk/common/exception.hpp
benjirewis: Removed.

src/viam/sdk/common/exception.hpp
benjirewis: Sounds good; done.

src/viam/sdk/common/exception.hpp
benjirewis: Cool went with `uint8_t`.

src/viam/sdk/common/exception.cpp
```diff
@@ -0,0 +1,61 @@
+#include <viam/sdk/common/exception.hpp>
+
+namespace viam {
+namespace sdk {
+
+Exception::Exception(const std::string& what)
+    : std::runtime_error("ViamException: " + what), condition_(ErrorCondition::k_general){};
```
acmorrow, line 7: Probably `"viam::sdk::Exception: "` here.

acmorrow, line 7: Maybe better to use constructor delegation here.

src/viam/sdk/common/exception.cpp
```diff
@@ -0,0 +1,61 @@
...
+
+Exception::Exception(ErrorCondition condition, const std::string& what)
+    : std::runtime_error("ViamException: " + what), condition_(condition){};
```
acmorrow, line 10: As above, `"viam::sdk::Exception: "`

src/viam/sdk/common/exception.hpp
```diff
@@ -0,0 +1,66 @@
...
+
+/// @defgroup Exception Classes related to SDK exceptions.
+
+/// @class ErrorCondition
+/// @brief Defines a set of a error conditions to be used in conjunction with
+/// Exception.
+/// @ingroup Exception
+enum class ErrorCondition : uint8_t {
+    k_general = 0,             // Default condition
+    k_connection,              // Issue during connection establishment
```
acmorrow, line 22: Let's give all of these explicit values for API stability reasons. Yes, they will get them by default, but if someone later (unwisely) added one in the middle, bad things would result.

src/viam/sdk/common/exception.hpp
```diff
@@ -0,0 +1,66 @@
...
+/// @class Exception
+/// @brief Defines an exception type for the SDK.
+/// @ingroup Exception
+class Exception : public std::runtime_error {
+   public:
+    explicit Exception(ErrorCondition condition, const std::string& what);
+    explicit Exception(const std::string& what);
+    virtual ~Exception();
+
+    std::error_condition condition() const noexcept;
```
acmorrow, line 41: This could return by `const&` too. Did you have a reason for the value return?

src/viam/sdk/common/exception.hpp
```diff
@@ -0,0 +1,66 @@
...
+};
+
+/// @class GRPCException
+/// @brief Defines an exception from a gRPC interaction.
+/// @ingroup Exception
+class GRPCException : public Exception {
+   public:
+    explicit GRPCException(grpc::Status status);
+
+    grpc::Status status() const noexcept;
```
acmorrow, line 54: Same thought re `const&` on the return.

src/viam/sdk/common/exception.hpp
```diff
@@ -0,0 +1,66 @@
+/// @file common/exception.hpp
+///
+/// @brief Defines custom exceptions for the SDK.
+#pragma once
+#include <grpcpp/support/status.h>
```
acmorrow, line 5: I want to think about how I feel about having Status become an embedded part of our API.

src/viam/sdk/common/exception.cpp
```diff
@@ -0,0 +1,61 @@
...
+
+Exception::~Exception() = default;
+
+std::error_condition Exception::condition() const noexcept {
+    return condition_;
+};
+
+struct ErrorCategory : std::error_category {
```
acmorrow, line 18: Since you are in `.cpp` scope here, you could just as well pull the implementations inline.

acmorrow (changes requested):

src/viam/sdk/common/exception.cpp
benjirewis: Ah good point; threw the struct declaration and its methods in the `make_error_condition` function.

src/viam/sdk/common/exception.hpp
benjirewis: Another option would be to have the error details, code and message as fields instead of a `grpc::Status`.

src/viam/sdk/common/exception.hpp
benjirewis: Done.

src/viam/sdk/common/exception.hpp
benjirewis: Ah no reason; returned by const reference instead.

src/viam/sdk/common/exception.hpp
benjirewis: Cool; done.

src/viam/sdk/common/exception.cpp
benjirewis: Done.

src/viam/sdk/common/exception.cpp
benjirewis: Yeah good point; done.

acmorrow (commented):
I like this, I think. I want to contemplate a bit how the gRPC error code and GRPCException thing should work w.r.t. to grpc::Status a bit more. Everything else here looks good though.

Probably time to promote it from draft because I expect this will merge in a form close to what is here now.

benjirewis:
> Probably time to promote it from draft because I expect this will merge in a form close to what is here now.

SGTM; I'll take some time to write tests in the meantime.

acmorrow (approved):
LGTM. We have other uses of grpc in our API currently so I'm not too concerned about one more.

src/viam/sdk/components/camera/camera.cpp
```diff
@@ -37,7 +37,7 @@ std::shared_ptr<ResourceRegistration> Camera::resource_registration() {
     const google::protobuf::ServiceDescriptor* sd =
         p->FindServiceByName(viam::component::camera::v1::CameraService::service_full_name());
     if (!sd) {
-        throw std::runtime_error("Unable to get service descriptor for the camera service");
+        throw Exception("Unable to get service descriptor for the camera service");
```
stuqdog, line 40: This Exception (and analogous in other resources) will disappear on rebase (fyi).

stuqdog (approved):
Sorry for review delay, lgtm!

PR #210, opened 2024-02-12
RSDK-3019 - add clang-tidy performance-move-const-arg

stuqdog:
- No longer disable `performance-move-const-arg` in `clang-tidy`. 
- Necessary fixes for `clang-tidy` to pass.

### DIFF
--- a/src/viam/sdk/registry/registry.cpp
+++ b/src/viam/sdk/registry/registry.cpp
@@ -57,13 +57,13 @@ const Model& ModelRegistration::model() const {
 };
 
 void Registry::register_model(std::shared_ptr<const ModelRegistration> resource) {
-    const std::string reg_key = resource->api().to_string() + "/" + resource->model().to_string();
+    std::string reg_key = resource->api().to_string() + "/" + resource->model().to_string();
     if (resources_.find(reg_key) != resources_.end()) {
         const std::string err = "Cannot add resource with name " + reg_key + "as it already exists";
         throw std::runtime_error(err);
     }
 
-    resources_.emplace(reg_key, std::move(resource));
+    resources_.emplace(std::move(reg_key), std::move(resource));
 }
 
 void Registry::register_resource_server_(
--- a/src/viam/sdk/registry/registry.cpp
+++ b/src/viam/sdk/registry/registry.cpp
@@ -95,14 +95,14 @@ std::shared_ptr<const ModelRegistration> Registry::lookup_model_inlock_(
 
 std::shared_ptr<const ModelRegistration> Registry::lookup_model(const std::string& name) {
     const std::lock_guard<std::mutex> lock(lock_);
-    return lookup_model_inlock_(name, std::move(lock));
+    return lookup_model_inlock_(name, lock);
 }
 
 std::shared_ptr<const ModelRegistration> Registry::lookup_model(const API& api,
                                                                 const Model& model) {
     const std::lock_guard<std::mutex> lock(lock_);
     const std::string name = api.to_string() + "/" + model.to_string();
-    return lookup_model_inlock_(name, std::move(lock));
+    return lookup_model_inlock_(name, lock);
 }
 
 std::shared_ptr<const ResourceServerRegistration> Registry::lookup_resource_server(const API& api) {
--- a/src/viam/sdk/registry/registry.cpp
+++ b/src/viam/sdk/registry/registry.cpp
@@ -63,7 +63,7 @@ void Registry::register_model(std::shared_ptr<const ModelRegistration> resource)
         throw std::runtime_error(err);
     }
 
-    resources_.emplace(std::move(reg_key), std::move(resource));
+    resources_.emplace(reg_key, std::move(resource));
 }
 
 void Registry::register_resource_server_(
@@ -102,7 +102,7 @@ std::shared_ptr<const ModelRegistration> Registry::lookup_model(const API& api,
                                                                 const Model& model) {
     const std::lock_guard<std::mutex> lock(lock_);
     const std::string name = api.to_string() + "/" + model.to_string();
-    return lookup_model_inlock_(std::move(name), std::move(lock));
+    return lookup_model_inlock_(name, std::move(lock));
 }
 
 std::shared_ptr<const ResourceServerRegistration> Registry::lookup_resource_server(const API& api) {
--- a/src/viam/sdk/resource/resource_manager.cpp
+++ b/src/viam/sdk/resource/resource_manager.cpp
@@ -47,9 +47,9 @@ void ResourceManager::replace_all(
     this->resources_ = new_resources;
     this->short_names_ = new_short_names;
 
-    for (const auto& resource : resources) {
+    for (auto resource : resources) {
         try {
-            do_add(resource.first, resource.second);
+            do_add(resource.first, std::move(resource.second));
         } catch (std::exception& exc) {
             BOOST_LOG_TRIVIAL(error) << "Error replacing all resources" << exc.what();
             return;
@@ -65,36 +65,35 @@ std::string get_shortcut_name(const std::string& name) {
     return name_split.at(name_split.size() - 1);
 }
 
-void ResourceManager::do_add(const Name& name, const std::shared_ptr<Resource>& resource) {
+void ResourceManager::do_add(Name name, std::shared_ptr<Resource> resource) {
     if (name.name().empty()) {
         throw "Empty name used for resource: " + name.to_string();
     }
-    const std::string short_name = name.short_name();
+    std::string short_name = name.short_name();
 
-    do_add(short_name, resource);
+    do_add(std::move(short_name), std::move(resource));
 }
 
-void ResourceManager::do_add(const std::string& name, const std::shared_ptr<Resource>& resource) {
+void ResourceManager::do_add(std::string name, std::shared_ptr<Resource> resource) {
     if (resources_.find(name) != resources_.end()) {
         throw "Attempted to add resource that already existed: " + name;
     }
 
-    resources_.emplace(name, resource);
-
-    const std::string shortcut = get_shortcut_name(name);
+    std::string shortcut = get_shortcut_name(name);
     if (shortcut != name) {
         if (short_names_.find(shortcut) != short_names_.end()) {
-            short_names_.emplace(shortcut, "");
+            short_names_.emplace(std::move(shortcut), "");
         } else {
-            short_names_.emplace(shortcut, name);
+            short_names_.emplace(std::move(shortcut), name);
         }
     }
+    resources_.emplace(std::move(name), std::move(resource));
 }
 
-void ResourceManager::add(const Name& name, const std::shared_ptr<Resource>& resource) {
+void ResourceManager::add(Name name, std::shared_ptr<Resource> resource) {
     const std::lock_guard<std::mutex> lock(lock_);
     try {
-        do_add(name, resource);
+        do_add(std::move(name), std::move(resource));
     } catch (std::exception& exc) {
         BOOST_LOG_TRIVIAL(error) << "Error adding resource to subtype service: " << exc.what();
     }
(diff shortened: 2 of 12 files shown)


src/viam/sdk/services/motion/server.cpp
```diff
@@ -103,37 +98,18 @@ ::grpc::Status MotionServer::GetPose(
...
+        const auto& component_name = Name::from_proto(request->component_name());
+        const std::string& destination_frame = request->destination_frame();
+        std::vector<WorldState::transform> supplemental_transforms;
+        for (const auto& proto_transform : request->supplemental_transforms()) {
+            supplemental_transforms.push_back(WorldState::transform::from_proto(proto_transform));
+        }
+        const pose_in_frame pose = motion->get_pose(
+            component_name, destination_frame, supplemental_transforms, helper.getExtra());
+        *response->mutable_pose() = pose.to_proto();
+    });
```
stuqdog, line 112: (flyby) Somehow missed this one when converting everything over to using `service_helper`.

src/viam/sdk/services/mlmodel/client.cpp
```diff
@@ -69,7 +69,7 @@ std::shared_ptr<MLModelService::named_tensor_views> MLModelServiceClient::infer(
         aav->views.emplace(kv.first, std::move(tensor));
     }
     auto* const tsav_views = &aav->views;
-    return {std::move(aav), tsav_views};
+    return {aav, tsav_views};
```
acmorrow, line 72: Surprising, but actually correct. The shared_ptr aliasing constructor takes its first arg by `const&`! I did not know that. Apparently they added the one taking an rvalue reference for C++20. So the code would be correct there. I'm inclined to NOLINT this one with a comment. The move is harmless up to C++17 since no move will happen, but desirable for C++20 where it saves a refcount jiggle.

src/viam/sdk/rpc/server.cpp
```diff
@@ -47,8 +47,8 @@ void Server::add_resource(std::shared_ptr<Resource> resource) {
                << " but no matching resource server as found";
         throw std::runtime_error(buffer.str());
     }
-    auto resource_server = managed_servers_.at(std::move(api));
-    resource_server->resource_manager()->add(resource->name(), std::move(resource));
+    auto resource_server = managed_servers_.at(api);
+    resource_server->resource_manager()->add(resource->name(), resource);
```
acmorrow, line 51: Surprising. It should be fine to move `resource` here.

src/viam/sdk/services/motion/server.cpp
```diff
@@ -34,11 +34,8 @@ ::grpc::Status MotionServer::Move(::grpc::ServerContext* context,
                 Motion::constraints::from_proto(request->constraints()));
         }
 
-        const bool success = motion->move(std::move(destination),
-                                          std::move(name),
-                                          std::move(ws),
-                                          std::move(constraints),
-                                          helper.getExtra());
+        const bool success = motion->move(
```
acmorrow, line 37: Only because `destination` and `name` were declared `const` above.

src/viam/sdk/services/motion/server.cpp
```diff
@@ -52,10 +49,8 @@ ::grpc::Status MotionServer::MoveOnMap(
         const auto& destination = pose::from_proto(request->destination());
         const auto& component_name = Name::from_proto(request->component_name());
         const auto& slam_name = Name::from_proto(request->slam_service_name());
-        const bool success = motion->move_on_map(std::move(destination),
-                                                 std::move(component_name),
-                                                 std::move(slam_name),
-                                                 helper.getExtra());
+        const bool success =
+            motion->move_on_map(destination, component_name, slam_name, helper.getExtra());
```
acmorrow, line 53: Ditto

acmorrow: Also, why are those variables declared as `&`s up on 49-51?

src/viam/sdk/services/motion/server.cpp
```diff
@@ -87,10 +82,10 @@ ::grpc::Status MotionServer::MoveOnGlobe(
                 motion_configuration::from_proto(request->motion_configuration()));
         }
 
-        const std::string execution_id = motion->move_on_globe(std::move(destination),
-                                                               std::move(heading),
-                                                               std::move(component_name),
-                                                               std::move(movement_sensor_name),
+        const std::string execution_id = motion->move_on_globe(destination,
```
acmorrow, line 85: Ditto, also why `&` again.

src/viam/sdk/services/motion/server.cpp
acmorrow: Nice catch!

src/viam/sdk/spatialmath/geometry.cpp
```diff
@@ -170,22 +170,22 @@ viam::common::v1::Geometry GeometryConfig::to_proto() const {
...
-    pose_.orientation = std::move(orientation);
+    pose_.orientation = orientation;
 }
 void GeometryConfig::set_theta(double theta) {
-    pose_.theta = std::move(theta);
+    pose_.theta = theta;
 }
 void GeometryConfig::set_geometry_type(GeometryType type) {
-    geometry_type_ = std::move(type);
+    geometry_type_ = type;
```
acmorrow, line 188: These all look like should be fine to move. What is the complaint?

src/viam/sdk/services/mlmodel/server.cpp
```diff
@@ -91,8 +91,8 @@ ::grpc::Status MLModelServiceServer::Metadata(
             target.Reserve(source.size());
             for (auto&& s : source) {
                 auto& new_entry = *target.Add();
-                *new_entry.mutable_name() = std::move(s.name);
-                *new_entry.mutable_description() = std::move(s.description);
+                *new_entry.mutable_name() = s.name;
+                *new_entry.mutable_description() = s.description;
```
acmorrow, line 95: Arguably, the correct fix here and below is to make the `source` parameter non-const.

acmorrow (changes requested):

src/viam/sdk/services/mlmodel/client.cpp
benjirewis: Speaking of: what C++ version is clang tidy using to generate its warnings? I suppose it's not C++20 given this comment, but was curious if it uses C++14 because of the cmakelists value?

benjirewis (approved):
LGTM % Drew's comments on non-consting instead of moving.

src/viam/sdk/rpc/server.cpp
stuqdog: (per error message) `add` takes a `const&` so no move will happen. But that seems a bit strange to me. I've updated for `add` to take by value and consistently move.

src/viam/sdk/services/motion/server.cpp
stuqdog: Good question on `&`s! Not sure, deleted.

I've un-`const`ed the variables and added back the moves (except for `destination`; `pose` is trivially copyable so I maintained the `const`ness of it).

src/viam/sdk/services/motion/server.cpp
stuqdog: `destination` (`struct geo_point`) is trivially copyable, as is `boost::optional<double>`. So `destination` and `heading` remain unmoved. 

`&`ness removed.

src/viam/sdk/spatialmath/geometry.cpp
stuqdog: These types are all trivially copyable, so the move has no effect.

src/viam/sdk/services/motion/server.cpp
stuqdog: As it turns out, we get stuck in a bit of a jam with `destination` and `name`. Making them un-const causes `misc-const-correctness` to complain, but leaving them const makes `performance-move-const-arg` complain! Thankfully they're only used in this call so we can just avoid assigning an lvalue.

src/viam/sdk/rpc/server.cpp
```diff
@@ -47,8 +47,9 @@ void Server::add_resource(std::shared_ptr<Resource> resource) {
                << " but no matching resource server as found";
         throw std::runtime_error(buffer.str());
     }
-    auto resource_server = managed_servers_.at(std::move(api));
-    resource_server->resource_manager()->add(resource->name(), std::move(resource));
+    auto resource_server = managed_servers_.at(api);
+    auto name = resource->name();
+    resource_server->resource_manager()->add(std::move(name), std::move(resource));
```
stuqdog, line 52: need to name and `move` `name` or else we get an error about `use` and `move` being unsequenced.

src/viam/sdk/rpc/dial.cpp
```diff
@@ -60,7 +60,7 @@ void DialOptions::set_entity(boost::optional<std::string> entity) {
 }
 
 void DialOptions::set_timeout(std::chrono::duration<float> timeout) {
-    timeout_ = std::move(timeout);
+    timeout_ = timeout;
```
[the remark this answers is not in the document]
acmorrow: OK I guess. I think it is overzealous. What if the type changed in the future and was no longer trivially copyable? Unlikely, I'll admit, for `std::chrono::duration`, but not for types we control. I don't see that the `move` has any downsides.

https://devblogs.microsoft.com/oldnewthing/20220512-00/?p=106651

Overall, I sort of feel like maybe we should disable that particular clang tidy check and just allow moves of trivial types. It is much easier to get SDK developers used to using `move` consistently if there isn't a gotcha about trivial types.

src/viam/sdk/registry/registry.cpp
```diff
@@ -102,7 +102,7 @@ std::shared_ptr<const ModelRegistration> Registry::lookup_model(const API& api,
                                                                 const Model& model) {
     const std::lock_guard<std::mutex> lock(lock_);
     const std::string name = api.to_string() + "/" + model.to_string();
-    return lookup_model_inlock_(std::move(name), std::move(lock));
```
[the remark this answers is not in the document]
acmorrow: No, you are right. Disregard my suggestion, I think it is fine how it is.

src/viam/sdk/registry/registry.cpp
```diff
@@ -102,7 +102,7 @@ std::shared_ptr<const ModelRegistration> Registry::lookup_model(const API& api,
...
+    return lookup_model_inlock_(name, std::move(lock));
```
acmorrow, line 105: Do you need the `move` on lock either? It's const.

src/viam/sdk/resource/resource_manager.cpp
```diff
@@ -47,9 +47,9 @@ void ResourceManager::replace_all(
     this->resources_ = new_resources;
     this->short_names_ = new_short_names;
 
-    for (const auto& resource : resources) {
+    for (auto resource : resources) {
```
acmorrow, line 50: I disagree with this one, looking at it. This works because it now copies the pair out, and then moves from it. That doesn't seem worthwhile vs just viewing the pair and copying the second in the call to `do_add`

src/viam/sdk/resource/resource_manager.cpp
```diff
@@ -65,36 +65,35 @@ std::string get_shortcut_name(const std::string& name) {
     return name_split.at(name_split.size() - 1);
 }
 
-void ResourceManager::do_add(const Name& name, const std::shared_ptr<Resource>& resource) {
+void ResourceManager::do_add(Name name, std::shared_ptr<Resource> resource) {
```
acmorrow, line 68: This could be `Name const& name, std::shared_ptr<Resource> resource)` because the first parameter is only observed, but the second one is actually moved.

acmorrow (approved):
This LGTM mod a few small things that don't, I think, warrant another round of review. I feel a bit like the warning on move of trivial is overzealous, but I don't feel strongly about it.

src/viam/sdk/rpc/dial.cpp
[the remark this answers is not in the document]
stuqdog: Yeah, this makes sense. I've set `performance-move-const-arg.CheckTriviallyCopyableMove` to false and reinstated the `move` here, as well as the `move`s in `geometry.cpp`

src/viam/sdk/registry/registry.cpp
stuqdog: Ah, nope! Surprised it let us keep that. Fixed.

PR #211, opened 2024-02-13
RSDK-3007 - add clang-tidy performance-unnecessary-value-param

stuqdog:
- adds performance-unnecessary-value-param checks to clang-tidy
- (flyby) implement a previously-unimplemented `to_proto` method
- (flyby) change `Vector3::to_proto` to no longer be static

### DIFF
--- a/src/viam/sdk/module/handler_map.cpp
+++ b/src/viam/sdk/module/handler_map.cpp
@@ -37,7 +37,7 @@ viam::module::v1::HandlerMap HandlerMap_::to_proto() const {
 HandlerMap_::HandlerMap_(){};
 
 // NOLINTNEXTLINE(readability-const-return-type)
-const HandlerMap_ HandlerMap_::from_proto(viam::module::v1::HandlerMap proto) {
+const HandlerMap_ HandlerMap_::from_proto(const viam::module::v1::HandlerMap& proto) {
     HandlerMap_ hm;
 
     const google::protobuf::RepeatedPtrField<viam::module::v1::HandlerDefinition>& handlers =
@@ -65,7 +65,7 @@ const HandlerMap_ HandlerMap_::from_proto(viam::module::v1::HandlerMap proto) {
     return hm;
 }
 
-void HandlerMap_::add_model(Model model, RPCSubtype subtype) {
+void HandlerMap_::add_model(const Model& model, const RPCSubtype& subtype) {
     handles_[subtype].push_back(model);
 }
 
--- a/src/viam/sdk/module/handler_map.cpp
+++ b/src/viam/sdk/module/handler_map.cpp
@@ -65,8 +65,8 @@ const HandlerMap_ HandlerMap_::from_proto(const viam::module::v1::HandlerMap& pr
     return hm;
 }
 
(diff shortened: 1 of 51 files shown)


src/viam/sdk/robot/client.cpp
```diff
@@ -1,17 +1,11 @@
...
 #include <cstddef>
-#include <iostream>
 #include <memory>
 #include <mutex>
-#include <ostream>
-#include <set>
 #include <string>
 #include <thread>
-#include <tuple>
-#include <type_traits>
```
stuqdog, line 14: (flyby) unused imports

src/viam/sdk/module/handler_map.cpp
```diff
@@ -65,7 +65,7 @@ const HandlerMap_ HandlerMap_::from_proto(viam::module::v1::HandlerMap proto) {
     return hm;
 }
 
-void HandlerMap_::add_model(Model model, RPCSubtype subtype) {
+void HandlerMap_::add_model(const Model& model, const RPCSubtype& subtype) {
```
acmorrow, line 68: I wonder if it would still want this as a `&` if you `std::move`'d `model` below.

src/viam/sdk/robot/client.cpp
```diff
@@ -222,7 +216,7 @@ void RobotClient::refresh_every() {
     }
 };
 
-RobotClient::RobotClient(std::shared_ptr<ViamChannel> channel)
+RobotClient::RobotClient(const std::shared_ptr<ViamChannel>& channel)
```
acmorrow, line 219: Could also leave it a value and `std::move` it on 220 probably.

src/viam/sdk/robot/client.cpp
```diff
@@ -234,8 +228,8 @@ std::vector<ResourceName>* RobotClient::resource_names() {
     return resources;
 }
 
-std::shared_ptr<RobotClient> RobotClient::with_channel(std::shared_ptr<ViamChannel> channel,
-                                                       Options options) {
+std::shared_ptr<RobotClient> RobotClient::with_channel(const std::shared_ptr<ViamChannel>& channel,
```
acmorrow, line 231: Ditto re maybe a `std::move` below makes if OK to pass by value?

acmorrow (approved):
This LGTM. There are a few places where you opted to transform an existing value parameter into a `const&` but where the object looks like it is then copied from the `&`. In principle those could probably be moves. Basically:

If you start with:
```
void foo(Value v) {
    do_something(v);
}
``` 

You could either go to:
```
void foo(const Value& v) {  // No copy here
    do_something(v);            // Copy here
}
```

Or
```
void foo(Value v) {                         // Copy OR move here; depends on caller
    do_something(std::move(v));  // Move here
}
```

So the former solution is a guaranteed copy in the implementation. The latter solution is either a copy and a move or two moves, depending on whether the caller provides an lvalue or an rvalue.

Anyway, I'm not proposing you go through this review and adjust. I think what you have here is fine especially if you are doing it primarily to placate the linter.

src/viam/sdk/robot/client.cpp
stuqdog: Ahh good catch, thanks! Fixed.

PR #212, opened 2024-02-14
RSDK-6484 - add reconfigurable base class

stuqdog:
Adds `Reconfigurable` base class, analogized from `Stoppable`

### DIFF
--- a/src/viam/sdk/resource/reconfigurable.cpp
+++ b/src/viam/sdk/resource/reconfigurable.cpp
@@ -1,5 +1,6 @@
-#include <memory>
 #include <viam/sdk/resource/reconfigurable.hpp>
+
+#include <memory>
 
 namespace viam {
 namespace sdk {
--- a/src/viam/sdk/resource/reconfigurable.cpp
+++ b/src/viam/sdk/resource/reconfigurable.cpp
@@ -1,0 +1,20 @@
+#include <memory>
+#include <viam/sdk/resource/reconfigurable.hpp>
+
+namespace viam {
+namespace sdk {
+
+Reconfigurable::~Reconfigurable() = default;
+Reconfigurable::Reconfigurable() = default;
+
+void Reconfigurable::reconfigure_if_reconfigurable(const std::shared_ptr<Resource>& resource,
+                                                   const Dependencies& deps,
+                                                   const ResourceConfig& cfg) {
+    auto reconfigurable_res = std::dynamic_pointer_cast<Reconfigurable>(resource);
+    if (reconfigurable_res) {
+        reconfigurable_res->reconfigure(deps, cfg);
+    }
+}
+
+}  // namespace sdk
+}  // namespace viam
--- a/src/viam/sdk/resource/reconfigurable.hpp
+++ b/src/viam/sdk/resource/reconfigurable.hpp
@@ -1,0 +1,31 @@
+#pragma once
+
+#include <viam/sdk/config/resource.hpp>
+#include <viam/sdk/resource/resource.hpp>
+
+namespace viam {
+namespace sdk {
+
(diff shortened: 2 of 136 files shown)


src/viam/sdk/resource/reconfigurable.hpp
```diff
@@ -0,0 +1,31 @@
...
+namespace sdk {
+
+class Reconfigurable {
+   public:
+    virtual ~Reconfigurable();
+
+    /// @brief Reconfigures a resource.
+    /// @param deps Dependencies of the resource.
+    /// @param cfg The resource's config.
+    virtual void reconfigure(const Dependencies& deps, const ResourceConfig& cfg) = 0;
```
stuqdog, line 16: I opted to change the signature of `reconfigure` to take its args by `const&`. I think this especially makes sense for the cfg; I could theoretically imagine someone wanting non-`const` access to `deps` but I think passing `deps` by value would be a bit strange, and I suspect that `const&` dependencies will be correct and manageable. In the worst case, we can add an override of `reconfigure` that isn't `const&`. Happy to revisit if others have thoughts.

purplenicole730 (approved):
LGTM

src/viam/sdk/resource/reconfigurable.cpp
```diff
@@ -0,0 +1,20 @@
+#include <memory>
+#include <viam/sdk/resource/reconfigurable.hpp>
```
acmorrow, line 2: The "associated" include should always be first, in a standalone block.

src/viam/sdk/resource/reconfigurable.hpp
```diff
@@ -0,0 +1,31 @@
...
+
+    /// @brief Reconfigures a resource if it is Reconfigurable.
+    /// @param resource the Resource to reconfigure.
+    /// @param deps Dependencies of the resource.
+    /// @param cfg The resource's config.
+    static void reconfigure_if_reconfigurable(const std::shared_ptr<Resource>& resource,
```
acmorrow, line 22: Not sure if we discussed this before or not, but I wonder if you could just make this a non-virtual method of `Resource`:

```
class Resource {
    ...
    void reconfigure(const Dep..., const Res...);
    ...
};

void Resource::reconfigure(...) {
    auto reconfigurable_res = std::dynamic_pointer_cast<Reconfigurable>(resource);
    if (reconfigurable_res) {
        reconfigurable_res->reconfigure(...);
    }
```

Same could be done for `stop` / `Stoppable`?

src/viam/sdk/resource/reconfigurable.hpp
acmorrow: I think it is fine.

acmorrow (approved):
One nit, and a question. I'm happy for the answer to the question to be "let's do that in a subsequent step/review".

src/viam/sdk/resource/reconfigurable.hpp
stuqdog: I like this, it seems nice to have `stop` and `reconfigure` sit on the `Resource` class. Happy enough to make the change here, it seems pretty simple.

One clarifying point, I _believe_ the methods will have to be `static` because otherwise the `this` arg is a raw pointer, but `dynamic_pointer_cast` requires a `shared_ptr`. Is there some cleverness I'm missing for how to have a non-static member method allow for a `dynamic_pointer_cast` of `this`?

acmorrow: Oh, sorry - I hadn't really focused on the fact that it took a `shared_ptr` arg. But I don't think it is an obstacle: you can just use plain `dynamic_cast` instead of `dynamic_pointer_cast`.

PR #213, opened 2024-02-15
RSDK-6530: Add depth map encode decode methods to C++ SDK

hexbabe:
From https://viam.atlassian.net/browse/RSDK-6530 :

> Sean Yu wrote a deserialize_depth_map function in order to consume depth maps in cpp-land for the [rgb-d-overlay](https://github.com/viam-labs/rgb-d-overlay) module. It is best to make this a static Camera method in C++ SDK


src/viam/sdk/components/camera/camera.hpp
```diff
@@ -83,6 +83,11 @@ class Camera : public Component {
         response_metadata metadata;
     };
 
+    /// @brief Decode image data of custom MIME type FORMAT_RAW_DEPTH into a standard
+    /// representation.
+    static std::tuple<uint64_t, uint64_t, std::vector<uint16_t>> deserialize_depth_map(
```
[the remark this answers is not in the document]
hexbabe: I think we should. It will require writing an equivalent `encode_depth_map` function, but I think that's a good idea I will do it

src/viam/sdk/components/camera/camera.cpp
```diff
@@ -24,6 +25,68 @@ API API::traits<Camera>::api() {
     return {kRDK, kComponent, "camera"};
 }
 
+std::vector<unsigned char> Camera::encode_depth_map(const Camera::depth_map& m) {
+    if (m.depth_values.size() != m.width * m.height) {
```
hexbabe, line 29: Does this stay here, or move into the `Camera::depth_map` constructor? Not sure if assertions like this belong in structs

stuqdog: That's a good question! My instinct is to keep it here, I think it makes documenting the error cases more consistent with what's happening in the code base currently. But I'm actually not sure what's more idiomatic for C++ broadly speaking, @acmorrow likely knows more!

stuqdog:
> Wait, does the `run-tests` action run unit tests?

yep! it's a little funky though, if one test fails in a file then often it says they all failed. Lemme know if you want any help with debugging.

purplenicole730 (commented):
I'm removing the request for my review since although I read through it, the camera logic is lost on me.

src/viam/sdk/components/camera/camera.hpp
```diff
@@ -83,6 +85,64 @@ class Camera : public Component {
...
+    /// format consists of "magic number" header (UTF-8 encoding for 'DEPTHMAP' in big-endian),
+    /// then the width and height encoded as 64-bit unsigned integers, followed by the depth
+    /// values encoded as 16-bit unsigned integers (big-endian format). This format is suitable
+    /// for serialization and transmission of depth map data through gRPC.
+    ///
+    /// @param width The width of the depth map in pixels.
+    /// @param height The height of the depth map in pixels.
+    /// @param depth_values A vector containing the depth values for each pixel in the depth
+    /// map.
+    ///                     Each depth value is a 16-bit unsigned integer.
```
stuqdog, line 125: `@param` is now just a `depth_map`

src/viam/sdk/components/camera/camera.hpp
```diff
@@ -3,8 +3,10 @@
 /// @brief Defines a `Camera` component.
 #pragma once
 
+#include <arpa/inet.h>
 #include <bitset>
 #include <chrono>
+#include <cstring>
```
stuqdog, line 9: Do we actually need these new includes? I don't see where they're being used here.

In particular the `arpa/inet.h` include is something I'd like to avoid if possible, since it isn't part of the standard library and so creates an additional dependency.

Note it's possible I'm being overly paranoid here! I'm not sure how standard the `arpa/inet.h` include is, so happy to have some pushback on this :)

src/viam/sdk/components/camera/camera.cpp
stuqdog: Ahh another reason to do it here is, someone could theoretically modify the `depth_map` after constructing it, thus calling here with an invalid depth_map even if they constructed a valid one. So yeah, let's leave it as is!

src/viam/sdk/components/camera/camera.cpp
```diff
@@ -7,9 +7,15 @@
 #include <viam/api/component/camera/v1/camera.grpc.pb.h>
 #include <viam/api/component/camera/v1/camera.pb.h>
 
+#include <viam/sdk/common/exception.hpp>
 #include <viam/sdk/common/utils.hpp>
 #include <viam/sdk/resource/resource.hpp>
 
+#ifndef htonll
+#define htonll(b) \
+    ((1 == htonl(1)) ? (b) : ((uint64_t)htonl((b)&0xFFFFFFFF) << 32) | htonl((b) >> 32))
```
stuqdog, line 16: Ahh I see where the `arpa` include comes into play here. If we must `include` it I'd prefer it to be in the `cpp` file rather than the `hpp`. But, it looks like we can do `endian` conversions with [boost](https://www.boost.org/doc/libs/1_84_0/libs/endian/doc/html/endian.html#conversion), which we already are relying on. can we swap to that?

stuqdog (commented):
Very cool, though I'll admit the camera logic flew past me a bit! Just a small comment update and a request to swap to `boost` for endian conversion, otherwise looks good!

src/viam/sdk/components/camera/camera.cpp
hexbabe: Good call. I didn't realize boost could do that. Might be best to refactor the Realsense to use that too... diff conversation though

src/viam/sdk/components/camera/camera.hpp
```diff
@@ -4,6 +4,7 @@
 #pragma once
 
 #include <bitset>
+#include <boost/endian/conversion.hpp>
```
stuqdog, line 7: can we put this `include` in the `.cpp`? Seems like an implementation detail that users don't need to know about.

Also note that the `include`s are sorted into different groups that might be a bit unintuitive. This should go with the `google/protobuf` includes in the `.cpp` as a third party include.

src/viam/sdk/components/camera/camera.hpp
```diff
@@ -83,6 +84,66 @@ class Camera : public Component {
...
+    /// for serialization and transmission of depth map data through gRPC.
+    ///
+    /// @param depth_map the dimensions and depth values of a depth map.
+    /// @return A std::vector<unsigned char> representing the encoded binary data of the depth
+    /// map.
+    ///         The vector includes 8 bytes for width, 8 bytes for height, followed by 2 bytes
+    ///         per depth value.
+    /// @throws Exception: if the depth data values do not correspond to height and width.
+    ///
+    static std::vector<unsigned char> encode_depth_map(const Camera::depth_map& m);
```
stuqdog, line 133: Hmm... I think probably this should be a non-static method on `depth_map`. While I like the shorter path of the method being on `Camera`, it seems a bit strange to have it be a static method that only takes a single `depth_map` object as an argument. And if `depth_map` ever were used by other resources such that we wanted to decouple it from `camera`, it would make sense to move the `encode`/`decode` methods along with it.

If we do end up going that route, it would make sense for `decode_depth_map` to similarly move onto `depth_map` (though of course that must remain static!).

stuqdog (approved):
Nice! A couple small fixes requested (no need to re-request review unless you're uncertain on any of it or have pushback), otherwise LGTM!

src/viam/sdk/components/camera/camera.hpp
```diff
@@ -83,6 +84,66 @@ class Camera : public Component {
...
+    static const auto HEADER_SIZE = 8 /*magic num*/ + 8 /*width*/ + 8 /*height*/;
+
+    /// @struct depth_map
+    /// @brief Represents the dimensions and depth values of a depth map.
+    ///
+    /// depth_map holds the width and height of a depth map, along with a vector
+    /// of actual depth values. Each depth value is a 16-bit unsigned integer representing
+    /// the distance from the camera to a point in the scene.
+    struct depth_map {
+        uint64_t width;                      ///< Width of the depth map in pixels.
```
acmorrow, line 100: That's a neat doxygen trick I didn't know.

src/viam/sdk/components/camera/camera.hpp
```diff
@@ -83,6 +84,66 @@ class Camera : public Component {
...
+        uint64_t height;                     ///< Height of the depth map in pixels.
+        std::vector<uint16_t> depth_values;  ///< A flat vector of depth values.
```
acmorrow, line 102: The `depth_map` type is the vocabulary type for the SDK. So I think it is worth ensuring that the interface that it presents offers good UX. As written, I still need to do index computation in order to obtain the depth value at a given `(x, y)` by doing `dm.depth_values[dm.height * x + y]` or something like that.

The C++ SDK already has a dependency on the xtensor library via the mlmodel service API. You could trivially represent this 2d array of uint16 with xtensor, and then you would get indexed access for free.

If xtensor feels like overkill for a 2d array there are probably other ways we can get a good 2d array concept rather than pushing the need to index into a linear array to the user.

src/viam/sdk/components/camera/camera.hpp
acmorrow: Also interested to know if this needs to be part of the public API? It feels like an implementation detail. That might argue for keeping it a free function rather than part of `depth_map` (if that class even continues to exist, per my xtensor comment).

src/viam/sdk/components/camera/camera.hpp
```diff
@@ -83,6 +84,66 @@ class Camera : public Component {
...
+    ///
+    /// This function processes a binary blob representing a depth map in a specific
+    /// format and extracts the dimensions and depth values contained within.
+    ///
+    /// @param data A vector of unsigned chars representing the binary data of the depth map.
+    /// @return A depth_map struct containing the width, height, and depth values extracted from the
+    /// data.
+    /// @throws Exception: if the data is misformatted e.g. doesn't contain valid depth information,
+    ///         or if the data size does not match the expected size based on the width and height.
+    static Camera::depth_map decode_depth_map(const std::vector<unsigned char>& data);
```
acmorrow, line 145: Ditto: not sure if this needs to be part of the public API or not.

src/viam/sdk/components/camera/camera.cpp
```diff
@@ -24,6 +25,96 @@ API API::traits<Camera>::api() {
     return {kRDK, kComponent, "camera"};
 }
 
+// Appends a uint64_t value in big-endian format to a byte vector and updates the offset.
+void append_uint64_big_endian(std::vector<unsigned char>& data, size_t& offset, uint64_t value) {
+    const uint64_t value_be = boost::endian::native_to_big(value);
+    if (data.size() < offset + 8) {
+        throw Exception("Incorrect data size-- attempted to write beyond data bounds");
```
acmorrow, line 32: `size--` looks like a C++ expression here. I'd write this as `"Incorrect data size: attempted ..."`

src/viam/sdk/components/camera/camera.cpp
```diff
@@ -24,6 +25,96 @@ API API::traits<Camera>::api() {
     return {kRDK, kComponent, "camera"};
 }
 
+// Appends a uint64_t value in big-endian format to a byte vector and updates the offset.
```
acmorrow, line 28: Just curious, but what "format" are we following here? How did big-endian get into the picture (hah!) at all?

src/viam/sdk/components/camera/camera.cpp
```diff
@@ -24,6 +25,96 @@ API API::traits<Camera>::api() {
...
+void append_uint64_big_endian(std::vector<unsigned char>& data, size_t& offset, uint64_t value) {
```
acmorrow, line 29: Utility functions like this should be in the unnamed namespace so they don't have linkage.

acmorrow (changes requested):

acmorrow:
Just a general feeling here, but this seems like a bit of a leaky abstraction. For instance, why are the depths uint16_t's? It feels like that's an artifact of an implementation somewhere, but is somewhat unnatural in the UX. Ditto the flat array presented with height/width. Also unclear to me where the endian-ness aspect comes in.

src/viam/sdk/components/camera/camera.cpp
hexbabe: Here is the context that I have: one of the reasons we are concerned with endianness is because the [Intel Realsense returns big-endian by default](https://github.com/viamrobotics/viam-camera-realsense?tab=readme-ov-file#attributes). Similarly, [the Python SDK also assumes big-endian when decoding](https://github.com/hexbabe/sean-viam-python-sdk/blob/5accda204293685954eb7382859f97c7f57b964b/src/viam/media/video.py#L43). I think @bhaney has more accurate and precise info on why we do this.

src/viam/sdk/components/camera/camera.cpp
hexbabe: Thanks for the review! New to C++ so learning a lot :D

hexbabe:
> Just a general feeling here, but this seems like a bit of a leaky abstraction. For instance, why are the depths uint16_t's? It feels like that's an artifact of an implementation somewhere, but is somewhat unnatural in the UX. Ditto the flat array presented with height/width. Also unclear to me where the endian-ness aspect comes in.

https://github.com/viamrobotics/rdk/blob/5df6d046388e79ab499a2f476d7e2b1fcc09d9e8/rimage/depth_map_raw.go#L311

Above is the specification in RDK about our custom depth map format. I can add a similar blurb about it in `camera.cpp`?

src/viam/sdk/components/camera/camera.hpp
hexbabe: I may be wrong but I think we should keep it as a flat vector. ML frameworks like pytorch and tensorflow has utilities to convert vectors to the native data structs e.g. `torch.from_blob` or `tf.convert_to_tensor`. And then you can use something like `.reshape` with the width and height. I looked it up and it's actually quite a bit harder to make this conversion with xtensor. 

I feel like direct indexing is not a common use case with depth maps? Again, not a CV guy could be wrong lol

src/viam/sdk/components/camera/camera.hpp
hexbabe: > not sure if this needs to be part of the public API or not.

Would the alternative to being part of the public API be to only include it in the cpp file as a private function? Would it just be part of get_image i.e. hook it up to the logic at the end to output the decoded value instead of the current raw bytes? That would be a breaking change then

src/viam/sdk/components/camera/camera.cpp
bhaney: endianness in this context is the endian-ness of a 16bit number, not of a system. Each pixel in the depth image is 2 bytes. For the two bytes, you can either read [first second] or [second first] within the byte stream. When I talk about big-endian, I mean it within this context

bhaney: calling `$xxd -g 1 test_image.viam.dep` on a test depth file, where a 20 x 10 image has each pixel value as x_location*y_location, this is what you would expect as the byte stream (with a space between every byte):
```
00000000: 44 45 50 54 48 4d 41 50 00 00 00 00 00 00 00 14  DEPTHMAP........
00000010: 00 00 00 00 00 00 00 0a 00 00 00 00 00 00 00 00  ................
00000020: 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00  ................
00000030: 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00  ................
00000040: 00 00 00 01 00 02 00 03 00 04 00 05 00 06 00 07  ................
00000050: 00 08 00 09 00 0a 00 0b 00 0c 00 0d 00 0e 00 0f  ................
00000060: 00 10 00 11 00 12 00 13 00 00 00 02 00 04 00 06  ................
00000070: 00 08 00 0a 00 0c 00 0e 00 10 00 12 00 14 00 16  ................
00000080: 00 18 00 1a 00 1c 00 1e 00 20 00 22 00 24 00 26  ......... .".$.&
```
`44 45 50 54 48 4d 41 50` is the string DEPTHMAP in ascii
`00 00 00 00 00 00 00 14` says the image has a width of 20 pixels
`00 00 00 00 00 00 00 0a` says the image has a height of 10 pixels

These need to be big-endian encoded numbers too, because if instead you encoded the width as `14 00 00 00 00 00 00 00`, that would imply an image width of 1441151880758558720

bhaney (dismissed):
Thanks for adding this! Just some small changes for clarity

acmorrow:
@hexbabe -

I think I had sort of misunderstood the point of this review. I expected it was creating a new type like `raw_image` that would actually be _used_ in the `Camera` api (e.g. `raw_image Camera::get_image(...)`). But that isn't what is going on here: the `depth_map` type isn't returned or consumed by `Camera`, and the two new functions exist only to transform a byte buffer into a `depth_map` and vice versa.

So, what is the flow of control here? How does the user of the SDK end up with a byte buffer which happens to contain an encoded depth map, and know that it should be passed to `Camera::decode_depth_map`? Or is this just a "part 1" review and subsequent changes will add methods that traffic in `depth_map` to `Camera`?

src/viam/sdk/components/camera/camera.cpp
```diff
@@ -24,6 +25,96 @@ API API::traits<Camera>::api() {
...
+    const uint64_t value_be = boost::endian::native_to_big(value);
+    if (data.size() < offset + 8) {
```
acmorrow, line 31: You can get rid of a lot of these explicit `8`'s by using `sizeof`

src/viam/sdk/components/camera/camera.hpp
```diff
@@ -83,6 +84,66 @@ class Camera : public Component {
         response_metadata metadata;
     };
 
+    /// @brief UTF-8 encoding of 'DEPTHMAP' used in the header of FORMAT_RAW_DEPTH bytes payload.
+    static const uint64_t MAGIC_NUMBER = 0x44455054484D4150ULL;
```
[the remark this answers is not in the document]
acmorrow: These constants belong in the `.cpp` file. Also you don't need to use `SHOUTY` constants in C++. Often in the C++SDK we prefix constants with `k_` to indicate that they are constant.

src/viam/sdk/components/camera/camera.hpp
acmorrow: I take back my comment about it being part of the public API. My comment was based on my assumption that `Camera` methods were actually going to be trading in `depth_map`s, in which case I wouldn't have expected this part of the transformation pair to need to be visible to users. However, per my newer understanding about what is being done here, I agree that it needs to be public. I still have some questions though about how users are expected to know what is and isn't a buffer containing a depth map that they should be passing to `Camera::decode_depth_map`, or what they are expected to do with a buffer containing an encoded depth map after they call `Camera::encode_depth_map`.

src/viam/sdk/components/camera/camera.hpp
acmorrow: As above.

src/viam/sdk/components/camera/camera.cpp
```diff
@@ -24,6 +25,96 @@ API API::traits<Camera>::api() {
...
+    uint64_t value;
+    std::memcpy(&value, &data[offset], 8);
+    value = boost::endian::big_to_native(value);
+    offset += 8;
+
+    return value;
+}
+
+std::vector<unsigned char> Camera::encode_depth_map(const Camera::depth_map& m) {
+    if (m.depth_values.size() != m.width * m.height) {
```
acmorrow, line 54: Couldn't the `depth_map` constructor enforce this invariant?

src/viam/sdk/components/camera/camera.cpp
```diff
@@ -24,6 +25,96 @@ API API::traits<Camera>::api() {
...
+            std::to_string(m.width * m.height) +
+            ". Actual: " + std::to_string(m.depth_values.size()));
+    }
+
+    const size_t total_byte_count = HEADER_SIZE + m.depth_values.size() * sizeof(uint16_t);
+    std::vector<unsigned char> data(total_byte_count);
+    size_t offset = 0;
+
+    // Network data is stored in big-endian, while most host systems are little endian.
+    append_uint64_big_endian(data, offset, MAGIC_NUMBER);
```
acmorrow, line 66: Might be better if `offset` was passed by pointer here to make it clearer to a reader that it was going to be modified.

src/viam/sdk/components/camera/camera.cpp
```diff
@@ -24,6 +25,96 @@ API API::traits<Camera>::api() {
...
+        const uint16_t depth_value = static_cast<uint16_t>(
+            data[data_index] << 8 | data[data_index + 1]);  // Assemble from big endian into uint16
+        arr.push_back(depth_value);
+    }
+
+    Camera::depth_map depth_map;
+    depth_map.width = width;
+    depth_map.height = height;
+    depth_map.depth_values = std::move(arr);
+    return depth_map;
```
acmorrow, line 115: I'll bet you can write this entire thing as just `return {width, height, std::move(arr)};`

src/viam/sdk/components/camera/camera.hpp
```diff
@@ -83,6 +84,66 @@ class Camera : public Component {
...
+    /// @brief Number of bytes of the header for FORMAT_RAW_DEPTH payloads
+    static const auto HEADER_SIZE = 8 /*magic num*/ + 8 /*width*/ + 8 /*height*/;
+
+    /// @struct depth_map
+    /// @brief Represents the dimensions and depth values of a depth map.
+    ///
+    /// depth_map holds the width and height of a depth map, along with a vector
+    /// of actual depth values. Each depth value is a 16-bit unsigned integer representing
+    /// the distance from the camera to a point in the scene.
+    struct depth_map {
```
acmorrow, line 99: There are invariants for a valid depth_map, notably that depth_values.size() == width * height. That'd argue for making this a class so that the invariant can be enforced on construction and then preserved.

acmorrow (changes requested):
Overall, the code looks good. I have some small comments and suggestions.

I do wonder whether 1) the depth_map stuff belongs on `Camera` at all given that the rest of the Camera API doesn't make use of it in any way (yet?), and 2) how exactly it is expected for clients or module authors to use these APIs.

hexbabe:
> @hexbabe -
> 
> I think I had sort of misunderstood the point of this review. I expected it was creating a new type like `raw_image` that would actually be _used_ in the `Camera` api (e.g. `raw_image Camera::get_image(...)`). But that isn't what is going on here: the `depth_map` type isn't returned or consumed by `Camera`, and the two new functions exist only to transform a byte buffer into a `depth_map` and vice versa.
> 
> So, what is the flow of control here? How does the user of the SDK end up with a byte buffer which happens to contain an encoded depth map, and know that it should be passed to `Camera::decode_depth_map`? Or is this just a "part 1" review and subsequent changes will add methods that traffic in `depth_map` to `Camera`?

The scope of this PR is to put decode and encode methods into the SDK as sort of util methods for depth map handling in modules and clients. For decoding, I wanted to mirror how they are handled in the Python SDK. In the Python SDK, we leave it to the client to call `raw_img.bytes_to_depth_array()` to decode the depth map. The goal here is to provide similar decode util for C++ SDK users. For encoding, we would expect depth camera module creators to use the encode function to convert their raw bytes into the Viam depth mime type format.

Sorry for all the confusion. My mistake was thinking that this ticket would be as simple as copying over [the function I wrote for the RGB-D overlay module](https://github.com/viam-labs/rgb-d-overlay/blob/4d102284647d39d62bfd47bc44b14dae3b7f6ccd/src/main.cpp#L29) into `Camera` to be used as util functions. If I were to redo this PR, I would've communicated the requirements and thought harder about how to integrate the methods into the exiting C++ APIs.

To amend this mistake, I think we should chat briefly to discuss where it makes sense to place these functions in a less haphazard way.

acmorrow:
@hexbabe and I met to discuss this. The plan is to move forward with this review. The `depth_map` type will be re-worked in terms of `xtensor` (either it will be an `xtensor` type alias, or it will wrap an `xtensor`). There will then be a subsequent review where the Camera interface is adjusted in order to integrate depth map information. This might look like having `Camera::get_image` return a variant over `raw_image` and `depth_map`, or adding a `get_depth_map` function that returns an `Optional<depth_map>`. Similar adjustments will need to be made to `get_images` to incorporate `depth_map`s.

src/viam/sdk/components/camera/camera.hpp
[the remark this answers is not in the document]
hexbabe: From the gcc [manual](http://gcc.gnu.org/onlinedocs/gcc/Long-Long.html):

> ISO C99 supports data types for integers that are at least 64 bits wide ( . . . ) . To make an integer constant of type long long int, add the suffix LL to the integer. To make an integer constant of type unsigned long long int, add the suffix ULL to the integer.

It should ensure that the literal has the correct type and size @bhaney

src/viam/sdk/components/camera.cpp
```diff
@@ -24,6 +29,106 @@ API API::traits<Camera>::api() {
...
+    append_uint64_big_endian(data, &offset, k_magic_number);
+    append_uint64_big_endian(data, &offset, width);
+    append_uint64_big_endian(data, &offset, height);
+
+    for (size_t i = 0; i < height; ++i) {
+        for (size_t j = 0; j < width; ++j) {
+            const uint16_t value = m(i, j);
+            const uint16_t value_be = boost::endian::native_to_big(value);
+            std::memcpy(&data[offset], &value_be, sizeof(uint16_t));
+            offset += sizeof(uint16_t);
```
acmorrow, line 88: If you make `append_uint64_big_endian` a template on the value type, you could re-use it here:

```
template<typename T>
append_big_endian(std::vector<unsigned char>& data, size_t* offset, T value);
```

src/viam/sdk/components/camera.cpp
```diff
@@ -24,6 +29,106 @@ API API::traits<Camera>::api() {
...
+    std::vector<uint16_t> depth_values;
+    depth_values.reserve(width * height);
+    for (size_t i = 0; i < width * height; ++i) {
+        const size_t data_index = k_header_size + i * sizeof(uint16_t);
+        const uint16_t depth_value = static_cast<uint16_t>(
+            data[data_index] << 8 | data[data_index + 1]);  // Assemble from big endian into uint16
+        depth_values.push_back(depth_value);
+    }
+
+    return xt::adapt(depth_values, std::array<size_t, 2>{height, width});
```
acmorrow, line 129: I think this is going to cause problems: I think `xt::adapt` is only going to view the memory in `depth_values`, and it will go out of scope when this function returns.

I think you will need another approach to construct an xtensor that owns the memory here. Let me know if you need me to look into how to do that best.

acmorrow (changes requested):
This looks really good, though I think there is a memory error that needs to be addressed.

src/viam/sdk/components/camera.cpp
hexbabe: Wow that's so elegant it brings a tear to my eye

src/viam/sdk/components/camera.cpp
```diff
@@ -24,6 +25,104 @@ API API::traits<Camera>::api() {
...
+    }
+
+    std::vector<uint16_t> depth_values;
+    depth_values.reserve(width * height);
+    for (size_t i = 0; i < width * height; ++i) {
+        depth_values.push_back(read_big_endian<uint16_t>(data, &offset));
+    }
+
+    xt::xarray<uint16_t> m = xt::xarray<uint16_t>::from_shape({height, width});
+    std::copy(depth_values.begin(), depth_values.end(), m.begin());
```
hexbabe, line 122: I looked up a method to copy the memory in. Let me know if there's a better way to do it @acmorrow

acmorrow: Hey, I was actually just looking at that. So, what you have here is fine, though it does cost you a copy. If you wanted to avoid the copy though, I think that's totally doable.

The trick would be to not create a `std::vector`, but just to directly create the `xt::xarray` you want to return, of the appropriate size and shape, before the loop, and then populate it in the loop.

```
xt::xarray<uint16_t> m = xt::xarray<uint16_t>::from_shape({height, width});
for (size_t i = 0; i < width * height; ++i) {
    m(<ii>, <jj>) = read_big_endian<uint16_t>(data, &offset);
```

Now, that'd require a little hassle to compute the correct `ii` and `jj` w.r.t. `height` and `width` and `i`, which is sort of a hassle. However, `xtensor` lets you reshape and view. So I think you could do it something like this:

```
xt::xarray<uint16_t> m = xt::xarray<uint16_t>::from_shape({height, width});
auto m_linear_view = xt::flatten(m);
for (size_t i = 0; i < width * height; ++i) {
    m_linear_view[i] = read_big_endian<uint16_t>(data, &offset));
return m;
```

It may not be exactly `flatten`, you might need `reshape_view`? I'd need to spend a little time with the `xtensor` docs to be sure. But something from https://xtensor.readthedocs.io/en/latest/view.html.

I'd say spend no more than 15 minutes on it. If you can make it work, great. If not, put in a TODO and leave it as a copy.

acmorrow (approved):
LGTM if you want to keep the copy. If you elect to use reshape/view to build directly into the output tensor, please post an update with that so I can give it a look.

PR #214, opened 2024-02-16
fix bitrot in example tflite module

acmorrow:


src/viam/examples/modules/tflite/main.cpp
```diff
@@ -93,7 +93,7 @@ class MLModelServiceTFLite : public vsdk::MLModelService, public vsdk::Stoppable
         }
     }
 
-    void reconfigure(vsdk::Dependencies dependencies, vsdk::ResourceConfig configuration) final
+    void reconfigure(const vsdk::Dependencies& dependencies, const vsdk::ResourceConfig& configuration) final
```
stuqdog, line 96: 👍 sorry I missed this!

stuqdog (approved):
lgtm!

src/viam/examples/modules/tflite/main.cpp
acmorrow: No worries - I actually forgot about it too, but then I was going to actually use this module for something and realized it was out of date!

PR #215, opened 2024-02-16
Move component client/server classes out of the API

acmorrow:
I'm still working through the implications of this, particularly w.r.t. to mocks and testing, but this is the general idea:
- Move `sdk/components/x/x.h` to `skd/components/x.h`.
- Move `sdk/components/x/client.h` to `sdk/components/private/x_client.h`.
- Move `sdk/components/x/server.h` to `sdk/components/private/x_server.h`.
And then adjust `#include`'s as necessary.

After this change, the `viam/sdk/components` directory looks like this in the source directory:
```
$ ls src/viam/sdk/components
base.cpp            camera.cpp          encoder.cpp         motor.cpp           power_sensor.cpp    sensor.hpp
base.hpp            camera.hpp          encoder.hpp         motor.hpp           power_sensor.hpp    servo.cpp
board.cpp           component.cpp       generic.cpp         movement_sensor.cpp private             servo.hpp
board.hpp           component.hpp       generic.hpp         movement_sensor.hpp sensor.cpp
```

And like this in the installation directory (i.e.`$PREFIX/include/viam/sdk/components`):
```
$ ls build/install/include/viam/sdk/components
base.hpp            camera.hpp          encoder.hpp         motor.hpp           power_sensor.hpp    servo.hpp
board.hpp           component.hpp       generic.hpp         movement_sensor.hpp sensor.hpp
```

If we like this approach, I'll do the same for `services`. In hindsight I should have done that one first as there are fewer services than components, but oh well.


acmorrow:
> Awesome, this looks great! Definitely a fan of the new look.

Cool! I'll get this done for `services` and promote this to a real review.

src/viam/sdk/tests/mocks/camera_mocks.hpp
```diff
@@ -4,9 +4,9 @@
 #include <viam/api/component/camera/v1/camera.grpc.pb.h>
 #include <viam/api/component/camera/v1/camera.pb.h>
 
-#include <viam/sdk/components/camera/camera.hpp>
-#include <viam/sdk/components/camera/client.hpp>
-#include <viam/sdk/components/camera/server.hpp>
+#include <viam/sdk/components/camera.hpp>
+#include <viam/sdk/components/private/camera_client.hpp>
+#include <viam/sdk/components/private/camera_server.hpp>
```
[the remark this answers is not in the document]
acmorrow: Yes. I want to understand whether tests / mocks actually require access to the concrete `FooServer` and `FooClient` types: it would be nice if they don't, because then we can actually not only make them "header private",  by not installing the headers, but we can make those types have hidden visibility so they aren't even part of the symbol table for the SDK library.

src/viam/sdk/tests/test_utils.hpp
```diff
@@ -61,11 +61,13 @@ void client_to_mock_pipeline(std::shared_ptr<Resource> mock, F&& test_case) {
     grpc::ChannelArguments args;
     auto test_server = TestServer(server);
     auto grpc_channel = test_server.grpc_in_process_channel(args);
-    ClientType resource_client(mock->name(), grpc_channel);
+
+    auto resource_client = Registry::lookup_resource_client(API::get<ClientType>())
```
acmorrow, line 65: This is a non-mechanical change, so that tests no longer have a dependency on the concrete `FooClient` type. Instead of directly constructing `ClientType`, we ask the registry to do it for us, then downcast. 

Potentially we should rename the `ClientType` template parameter?

src/viam/sdk/tests/test_mlmodel.cpp
```diff
@@ -12,17 +12,15 @@
...
 #include <memory>
 #include <tuple>
 #include <unordered_map>
 
 #include <boost/variant/get.hpp>
 
-#include <viam/sdk/services/mlmodel/client.hpp>
-#include <viam/sdk/services/mlmodel/private/proto.hpp>
-#include <viam/sdk/services/mlmodel/server.hpp>
+#include <viam/sdk/services/private/proto.hpp>
```
acmorrow, line 23: This needs some thought: this test currently tests directly some internal proto to vocab conversions. I do want to keep testing that code as it is particularly tricky. I'll give it some thought, as there will probably be other places we need something like this.

src/viam/sdk/tests/test_encoder.cpp
```diff
@@ -7,14 +7,8 @@
 
 #include <boost/test/included/unit_test.hpp>
 
-#include <viam/api/common/v1/common.pb.h>
```
acmorrow, line 10: I stripped lots of direct `#include <viam/api` where I could. There are still some in our public headers. We should strive to eliminate them all.

src/viam/sdk/services/private/proto.cpp
```diff
@@ -12,7 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-#include <viam/sdk/services/mlmodel/private/proto.hpp>
+#include <viam/sdk/services/private/proto.hpp>
```
acmorrow, line 15: This file and the associated header should be renamed somehow.

acmorrow (commented):
@stuqdog - Updated, now services are handled the same way. I ran into one small issue with testing in the mlmodel stuff, so at least one more update is on the way. But please give this a look. I expect given that you were happy with the `components` changes that you will also be fine with the `services` changes. The only really interesting change in this review was changing how the mock pipeline stuff worked so it doesn't need to have access to the concrete client types.

src/viam/sdk/tests/test_utils.hpp
stuqdog: Yeah, I think `ResourceType` is probably better at this point.

src/viam/sdk/tests/test_encoder.cpp
stuqdog: Just created a [ticket](https://viam.atlassian.net/browse/RSDK-6734) for this.

acmorrow:
@stuqdog - This is ready for another round of review. Changes since last time:

- Pushed all "private" stuff into an `impl` namespace. I like this because it gives us a way to see at a glance in the symbols (or in source usage for that matter) if an internal type is being exported or used somehow. Happy to bikeshed a name. Ideally I'd call it `private` but that's a keyword of course.
- Fixed the mlmodel tests to not use the internal "proto" functions anymore but test them indirectly by way of a mock and a client pipeline.
- Renamed the `components/private/proto` headers to `components/private/mlmodel` and renamed the `mlmodel_details` namespace (which used to be public) to `mlmodel` inside the `impl` namespace.

src/viam/sdk/tests/test_mlmodel.cpp
acmorrow: The tests aren't part of the `libviamsdk` library. Eventually, the types and symbols under `private` won't even be exported from `libviamsdk`, so the tests won't be able to use them. Better to clean it up now.

stuqdog: Got it, thanks for clarifying.

PR #216, opened 2024-02-20
RSDK-3008 - add misc-unused-params

stuqdog:


src/viam/sdk/tests/mocks/mock_motion.cpp
```diff
@@ -28,11 +28,12 @@ bool MockMotion::move(const pose_in_frame& destination,
...
-    this->peek_current_pose = std::move(destination);
-    this->peek_component_name = std::move(component_name);
-    this->peek_slam_name = std::move(slam_name);
-    this->current_location.pose = std::move(destination);
+                             const AttributeMap&) {
+    // CR erodkin: flyby
+    this->peek_current_pose = destination;
+    this->peek_component_name = component_name;
+    this->peek_slam_name = slam_name;
+    this->current_location.pose = destination;
```
stuqdog, line 36: (flyby) here and below, `move`ing these args is meaningless because they're `const&`, and we can't make them not `const&` here without modifying the signature in the `Motion` base class. So, we get rid of the `move`.

acmorrow (approved):
LGTM

PR #219, opened 2024-03-01
RSDK-6726 - update client types in c++ code samples

stuqdog:
`apis.json` was still passing `FooClient` as the client type for code samples, leading to `resource_by_name<FooClient>(...)` rather than the now-correct `resource_by_name<Foo>(...)`. This PR fixes that.

### DIFF
--- a/src/viam/sdk/robot/client.cpp
+++ b/src/viam/sdk/robot/client.cpp
@@ -311,7 +311,9 @@ RobotClient::RobotClient(std::shared_ptr<ViamChannel> channel)
     : channel_(channel->channel()),
       viam_channel_(std::move(channel)),
       should_close_channel_(false),
-      impl_(std::make_unique<impl>(RobotService::NewStub(channel_))) {}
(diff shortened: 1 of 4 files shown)


purplenicole730 (approved):

src/viam/sdk/robot/client.cpp
```diff
@@ -311,7 +311,9 @@ RobotClient::RobotClient(std::shared_ptr<ViamChannel> channel)
     : channel_(channel->channel()),
       viam_channel_(std::move(channel)),
       should_close_channel_(false),
-      impl_(std::make_unique<impl>(RobotService::NewStub(channel_))) {}
+      impl_(std::make_unique<impl>(RobotService::NewStub(channel_))) {
+    Registry::initialize();
+}
```
stuqdog, line 316: (flyby) we need to ensure that the `Registry` is properly initialized even when functioning as a pure client. This isn't a great long-term fix but @acmorrow has some future work on library initialization that will help clear this up.

acmorrow (approved):
LGTM

PR #221, opened 2024-03-13
RSDK-6636 - refactor AttributeMap and ProtoType

stuqdog:
Currently this is just changes to `proto_type.hpp`, still a WIP

Assuming we're happy with the shape of this, I will go about renaming the file and updating the rest of the code base.

### DIFF
--- a/src/viam/sdk/common/proto_type.hpp
+++ b/src/viam/sdk/common/proto_type.hpp
@@ -10,8 +10,6 @@
 namespace viam {
 namespace sdk {
 
-class AttributeMap;
-
 namespace value_type_details {
 
 template <typename T>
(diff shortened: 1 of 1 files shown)


src/viam/sdk/common/proto_type.hpp
```diff
@@ -27,64 +27,82 @@ struct get_helper {
...
-
-    template <typename V>
-    static std::shared_ptr<const AttributeMap::element_type> get(const V& v) {
-        auto* const result = boost::get<AttributeMap>(&v);
-        return result ? *result : nullptr;
-    }
-};
-
-}  // namespace prototype_details
+}  // namespace value_type_details
```
acmorrow, line 30: The purpose of `get_helper` was to specialize for the `AttributeMap` case to deal with the extra layer of pointer with an automatic dereference in the populated case. Since the specialization has been removed, it suggests that `get_helper` is no longer needed. But I have a suspicion it still may need to exist, but with different specializations.

src/viam/sdk/common/proto_type.hpp
```diff
@@ -27,64 +27,82 @@ struct get_helper {
...
+        value_type();
+        value_type(std::string s);
+        value_type(const char* c);
+        value_type(int i);
+        value_type(float f);
+        value_type(double d);
+        value_type(bool b);
+        value_type(std::vector<value_type> v);
+        value_type(map_type);
+        value_type(google::protobuf::Value value);
```
acmorrow, line 66: For at least these types, copying is pretty expensive. I recommend providing both a `const&` and a `&&` overload.

Alternatively, you could write a universally forwarding constructor:

```
template<typename T>
value_type(T&& t)
```

However, that tends to be too greedy without some SFINAE to make sure it doesn't get invoked when it shouldn't.

src/viam/sdk/common/proto_type.hpp
```diff
@@ -27,64 +27,82 @@ struct get_helper {
...
+
+        value_type();
+        value_type(std::string s);
+        value_type(const char* c);
+        value_type(int i);
+        value_type(float f);
+        value_type(double d);
+        value_type(bool b);
+        value_type(std::vector<value_type> v);
+        value_type(map_type);
```
acmorrow, line 65: This constructor is going to prove to be troublesome due to the containers vs incomplete types issue from above.

src/viam/sdk/common/proto_type.hpp
```diff
@@ -27,64 +27,82 @@ struct get_helper {
...
-    }
+    class value_type;
+
+    using map_type = std::unordered_map<std::string, value_type>;
+
+    class value_type {
+       public:
+        template <typename T>
+        auto get() {
+            return value_type_details::get_helper<T>::get(value_);
```
acmorrow, line 42: Without any specialization for `get_helper<T>`, these just decay into `boost::get`. But I think you may want new specializations? If a user asks for the `std::vector<value_type>` member, does it get you it, or do you get the wrapper? I'm not actually sure.

src/viam/sdk/common/proto_type.hpp
```diff
@@ -27,64 +27,82 @@ struct get_helper {
...
+    AttributeMap();
+    ~AttributeMap();
+    // CR erodkin: add comments to all of this probably
+    google::protobuf::Struct to_proto() const;
+
+    bool contains(const std::string& key) const;
+
+    // CR erodkin: consider how to return a ref or pointer to the value type such that we don't have
+    // to copy the `value_type`, while retaining guarantees of const-ness.
+    value_type get(const std::string& key) const;
```
acmorrow, line 83: Is there a problem with just having:
```
const value_type& get(...) const;
value_type& get(...);
```

acmorrow, line 83: What `value_type` gets returned if the item is not found? Or does it throw?

src/viam/sdk/common/proto_type.hpp
```diff
@@ -27,64 +27,82 @@ struct get_helper {
...
+    void set(std::string key, value_type::value_types value) const;
+    void set(std::string key, value_type value) const;
+
+    // append a single value to a list type within the AttributeMap
+    void append(const std::string& key, value_type::value_types value);
+    void append(const std::string& key, value_type value);
+
+    // add a single key/value pair to a map type within the AttributeMap
+    void insert(const std::string& key, std::pair<std::string, value_type>);
+    void insert(const std::string& key, std::pair<std::string, value_type::value_types>);
```
acmorrow, line 95: It may just be better to expose the `map_` rather than try to define your own wrapper above the unordered_map interface.

acmorrow (changes requested):

src/viam/sdk/common/proto_type.hpp
```diff
@@ -27,64 +27,82 @@ struct get_helper {
...
-        return prototype_details::get_helper<T>::get(proto_type_);
-    }
-
-    template <typename T>
-    auto get() const {
-        return prototype_details::get_helper<T>::get(proto_type_);
-    }
+    class value_type;
+
+    using map_type = std::unordered_map<std::string, value_type>;
```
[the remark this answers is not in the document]
stuqdog: ack, will change to a `unique_ptr`

src/viam/sdk/common/proto_type.hpp
```diff
@@ -27,64 +27,82 @@ struct get_helper {
...
+
+        template <typename T>
+        auto get() const {
+            return value_type_details::get_helper<T>::get(value_);
+        }
+
+        using base_value_types_ =
+            boost::variant<boost::blank, std::string, const char*, int, float, double, bool>;
+        using value_types = boost::make_recursive_variant<
+            base_value_types_,
```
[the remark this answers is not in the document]
stuqdog: Hmm... I agree that seeing `boost::variant` and then `boost::make_recursive_variant` on it immediately after is a little strange. My thinking though was that we still wanted the root type to be a variant. So in essence the three cases for a value type are 
```
| boost::variant<...>,
| std::vector<boost::variant<...>>,
| std::unordered_map<std::string, boost::variant<...>>,
```

In playing around with this definition I was able to build a `value_type` with raw values, with lists, with maps, with nested lists/maps, etc. etc. I also got the sense that `make_recursive_variant_over` was largely equivalent to `make_recursive_variant`, per description [here](https://www.boost.org/doc/libs/1_84_0/doc/html/boost/make_recurs_1_3_44_5_5_1_3.html).* That said, I can definitely believe I'm missing something and that use of `mpl`/`make_recursive_variant_over` would produce superior results!

* `type has behavior equivalent in every respect to [make_recursive_variant](https://www.boost.org/doc/libs/1_84_0/doc/html/boost/make_recursive_variant.html)< Sequence[0], Sequence[1], ... >::type (where Sequence[i] denotes the i-th element of Sequence), except that no upper limit is imposed on the number of types.`

src/viam/sdk/common/proto_type.hpp
stuqdog: Ahh yeah, wrapping these in `unique_ptr` now.

src/viam/sdk/common/proto_type.hpp
stuqdog: Yeah good question. My intuition is returning a `boost::blank` as the default is best, people can check for the "not found" case pretty easily (`if (!foo)`) and it saves having to potentially large chunks of code in a `try/catch`.

src/viam/sdk/common/proto_type.hpp
stuqdog: Yeah, I agree this is trying to do to much. I'm wondering though if it's better to keep the `AttributeMap` class around and expose `map_` or to just define `value_type` at the top level and have `AttributeMap` just be a typedef on the map type. The latter seems more intuitive to me personally, though it does perhaps make `get` helpers slightly more awkward if they're just free standing functions rather than a class method.

src/viam/sdk/common/proto_type.hpp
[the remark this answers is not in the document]
acmorrow: I don't think you need to change to unique ptr. You just can't declare this typedef with `value_type` as an incomplete type. But the recursive_wrapper stuff can handle it. Adding unique_ptr is just doubling up on the indirection that recursive_wrapper already provides.

src/viam/sdk/common/proto_type.hpp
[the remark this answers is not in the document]
acmorrow: Yes, make_recursive_variant_over is equivalent. What I'm puzzled about is why there are two levels of variant, rather than just:

```
boost::make_recursive_variant<boost::blank, ..., bool, std::unordered_map<std::string, boost::recursive_variant_>, std::vector<boost::recursive_variant_>>
```

Why the second level of `variant`? I wouldn't expect it to be needed.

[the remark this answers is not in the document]
stuqdog: Ahh I see, that makes sense. Yeah I don't think the second level of variant is needed, will fix.

src/viam/sdk/common/proto_type.hpp
stuqdog: I see, I was misunderstanding. Thanks for clarifying! If I put the `using map_type = ...` inside `value_type` class proper, is that allowed?

njooma:
What's the status of this? Stale?

acmorrow:
More "stalled" than "stale". I was investigating some alternative strategies with hopes of limiting the use of boost in the API, or at least reducing it to a point where it could be poly-filled away with `std::variant`. My ideas didn't quite work out, or, didn't quite work out to a point where I had full confidence in pursuing it. We should probably pull @lia-viam into the discussion and see if we can get it moving forward again.

PR #222, opened 2024-03-15
RSDK-6627 - move resource-names-for-resource to be a method on resource

stuqdog:


src/viam/sdk/resource/resource.hpp
```diff
@@ -28,6 +28,8 @@ class Resource {
     /// @brief Return the resource's name.
     virtual std::string name() const;
 
+    std::vector<Name> resource_names() const;
```
acmorrow, line 31: Let's please add a doc comment for this while we are moving it.

Also, is `resource_names` the best name for this method? I'm actually a little unclear on what the meaning of the returned names is. It seems to be all the registered models that share the same API as this resource?

acmorrow (approved):
If you want to keep this to pure code motion to limit scope then LGTM. If you are willing to put a few more minutes into it, I think it'd be nice to see a doc comment, and maybe bikeshed the name of the function? I'm a little unclear on what it is intended to do.

src/viam/sdk/resource/resource.hpp
stuqdog: Done! Though, it appears the only place this method is ever used is on a private method in `RobotService`. I wonder if it might make sense to just have this be a private method and make `RobotService` a friend class?

src/viam/sdk/robot/service.cpp
```diff
@@ -42,8 +42,8 @@ RobotService_::RobotService_(const std::shared_ptr<ResourceManager>& manager, Se
 std::vector<ResourceName> RobotService_::generate_metadata_() {
     std::vector<ResourceName> metadata;
     for (const auto& key_and_val : resource_manager()->resources()) {
-        for (const Name& resource : resource_names_for_resource(key_and_val.second)) {
-            metadata.push_back(resource.to_proto());
+        for (const Name& name : key_and_val.second->registered_models()) {
```
acmorrow, line 45: If this is really the only usage of `registered_models`, I'd suggest just moving the function into this file in the unnamed namespace and changing it to take a `const Resource&`

acmorrow (approved):
LGTM

purplenicole730 (approved):
Looks good! 😁

PR #226, opened 2024-04-10
RSDK-7132 Stream Ticks

oliviamiller:


src/viam/sdk/components/board.hpp
```diff
@@ -44,6 +45,15 @@ class Board : public Component {
         std::unordered_map<std::string, digital_value> digital_interrupt_values;
     };
 
+    /// @struct tick
+    /// A board digital interrupt that contains high/low value and the time the digital interrupt
+    /// occured.
+    struct tick {
+        std::string pin_name;
+        bool high;
+        uint64_t time;
```
acmorrow, line 54: - Slight preference for putting the bool last here.
- Consider using std::chrono for `time`

src/viam/sdk/components/board.hpp
```diff
@@ -221,6 +231,22 @@ class Board : public Component {
     virtual digital_value read_digital_interrupt(const std::string& digital_interrupt_name,
                                                  const AttributeMap& extra) = 0;
 
+    /// @brief Returns a stream of digital interrupt ticks.
+    /// @param digital_interrupt_names digital interrupts to stream
+    /// @param ticks queue to put the ticks in
+    inline void stream_ticks(const std::vector<std::string> digital_interrupt_names,
```
acmorrow, line 237: - The `digital_interrupt_names` parameter should probably be passed by `const&` to avoid the cost of copying it.

- I'm not clear on two things about the `ticks` parameter. First, why is it wrapped in a `shared_ptr`? Second, why is it a parameter rather than a return value? Is the idea that you should be able to repeatedly call this with the same queue to gather results periodically? If so, is there a reason it can't just be `std::queue<tick>& ticks` or `std::queue<tick>& ticks`?

- `std::queue` is a really narrow interface. I'm not sure it is helpful, in the sense that if you wanted to act on all the ticks you would need to do so destructively. I'd suggest `std::vector` as the accumulator type.

src/viam/sdk/components/private/board_client.cpp
```diff
@@ -137,6 +139,33 @@ Board::digital_value BoardClient::read_digital_interrupt(const std::string& digi
     return response.value();
 }
 
+void BoardClient::stream_ticks(const std::vector<std::string> digital_interrupt_names,
+                               std::shared_ptr<std::queue<tick>> ticks,
+                               const AttributeMap& extra) {
+    viam::component::board::v1::StreamTicksRequest request;
```
acmorrow, line 145: Could this implementation use the `make_client_helper` used elsewhere in this file? It is OK if it can't for some reason (though I'd like to know why), since it is a helper and not intended to cover 100% of cases, but I'm always interested to know if there are cases that it currently can't cover but could be made to.

src/viam/sdk/components/private/board_client.cpp
```diff
@@ -137,6 +139,33 @@ Board::digital_value BoardClient::read_digital_interrupt(const std::string& digi
...
+    viam::component::board::v1::StreamTicksResponse response;
+    ClientContext ctx;
+
+    request.set_name(this->name());
+
+    for (unsigned int i = 0; i < digital_interrupt_names.size(); i++) {
```
acmorrow, line 151: You could use a more compact syntax here I think:

```
for (const auto& item : digital_interrupt_names) {
```

src/viam/sdk/components/private/board_client.cpp
```diff
@@ -137,6 +139,33 @@ Board::digital_value BoardClient::read_digital_interrupt(const std::string& digi
...
+        request.add_pin_names(digital_interrupt_names[i]);
+    }
+    *request.mutable_extra() = map_to_struct(extra);
+
+    std::unique_ptr<
+        ::grpc::ClientReaderInterface<::viam::component::board::v1::StreamTicksResponse>>
```
acmorrow, line 157: Can probably just say `auto reader = stub_->StreamTicks`

src/viam/sdk/components/private/board_client.cpp
```diff
@@ -137,6 +139,33 @@ Board::digital_value BoardClient::read_digital_interrupt(const std::string& digi
...
+        reader = stub_->StreamTicks(ctx, request);
+
+    while (reader->Read(&response)) {
```
acmorrow, line 160: How long does this loop run?

src/viam/sdk/components/private/board_client.cpp
```diff
@@ -137,6 +139,33 @@ Board::digital_value BoardClient::read_digital_interrupt(const std::string& digi
...
+        Board::tick tick;
+        tick.pin_name = response.pin_name();
+        tick.high = response.high();
+        tick.time = response.time();
+        ticks->push(tick);
```
acmorrow, line 165: This can probably be simplified to:
```
while (...) {
    ticks->push({std::move(response.pin_name()), std::move(response.high()), std::move(response.time()};
}
```

src/viam/sdk/components/private/board_server.cpp
```diff
@@ -174,6 +177,51 @@ ::grpc::Status BoardServer::GetDigitalInterruptValue(
...
+    }
+
+    const std::shared_ptr<std::queue<Board::tick>> ticks;
+    const std::vector<std::string> digital_interrupt_names(request->pin_names().begin(),
+                                                           request->pin_names().end());
+
+    board->stream_ticks(digital_interrupt_names, ticks, extra);
+
+    ::viam::component::board::v1::StreamTicksResponse response;
+    while (true) {
```
acmorrow, line 208: Could this just be `while (!ticks->empty())`?

src/viam/sdk/components/private/board_server.cpp
```diff
@@ -174,6 +177,51 @@ ::grpc::Status BoardServer::GetDigitalInterruptValue(
...
+        if (!ticks->empty()) {
+            const Board::tick tick = ticks->front();
+            ticks->pop();
+            response.set_pin_name(tick.pin_name);
+            response.set_high(tick.high);
+            response.set_time(tick.time);
+            writer->Write(response);
+        }
+        if (context->IsCancelled()) {
```
acmorrow, line 217: Maybe this should happen before doing the Write?

src/viam/sdk/components/private/board_server.cpp
```diff
@@ -174,6 +177,51 @@ ::grpc::Status BoardServer::GetDigitalInterruptValue(
     return ::grpc::Status();
 }
 
+::grpc::Status BoardServer::StreamTicks(
```
acmorrow, line 180: Is there something that prevents use of `make_service_helper` here? Thoughts much like my request about the client helper.

src/viam/sdk/components/private/board_server.cpp
```diff
@@ -174,6 +177,51 @@ ::grpc::Status BoardServer::GetDigitalInterruptValue(
...
+
+    ::viam::component::board::v1::StreamTicksResponse response;
+    while (true) {
+        if (!ticks->empty()) {
+            const Board::tick tick = ticks->front();
+            ticks->pop();
+            response.set_pin_name(tick.pin_name);
+            response.set_high(tick.high);
+            response.set_time(tick.time);
+            writer->Write(response);
```
acmorrow, line 215: There is some extra copying goign on here. You are copying the element out of the front of the queue, then destroying the copy in the queue, then copying into the response. I think you can do fewer copies by keeping a reference to `front`, moving items out of it into the response, the popping at the end:
```
const auto& tick = ticks->front();
response.set_pin_name(std::move(tick.pin_name));
response.set_high(std::move(tick.high));
response.set_time(std::move(tick.time));
ticks->pop();
writer->Write(response);
```

src/viam/sdk/tests/mocks/mock_board.cpp
```diff
@@ -71,6 +71,14 @@ Board::digital_value MockBoard::read_digital_interrupt(const std::string& digita
     return this->peek_read_digital_interrupt_ret;
 }
 
+void MockBoard::stream_ticks(const std::vector<std::string> digital_interrupt_names,
+                             std::shared_ptr<std::queue<tick>> ticks,
+                             const AttributeMap& extra) {
+    for (unsigned int i = 0; i < digital_interrupt_names.size(); i++) {
```
acmorrow, line 77: Use range for here (e.g. `(for const auto& name : ...)`)

acmorrow (changes requested):
This looks like a good start. I do have some suggestions for how to improve it though. I think maybe focus on the API issues in `board.hpp` first, because I think changes there may obsolete or affect some of the other comments, but I wanted to share everything that stood out to me anyway.

src/viam/sdk/components/private/board_server.cpp
```diff
@@ -1,12 +1,15 @@
...
+#include <queue>
 #include <viam/sdk/common/service_helper.hpp>
 #include <viam/sdk/common/utils.hpp>
 #include <viam/sdk/components/board.hpp>
 #include <viam/sdk/config/resource.hpp>
 #include <viam/sdk/resource/resource_manager.hpp>
 #include <viam/sdk/rpc/server.hpp>
 
+using grpc::ServerWriter;
+
```
stuqdog, line 12: It doesn't look like we're actually taking advantage of this `using` line here.

src/viam/sdk/components/private/board_server.cpp
stuqdog, line 208: Is there any risk that this will become blocking on any other calls to a board server? Should we be putting this loop in a separate thread and returning right away?

src/viam/sdk/components/private/board_server.hpp
```diff
@@ -74,6 +74,11 @@ class BoardServer : public ResourceServer,
         const ::viam::component::board::v1::GetDigitalInterruptValueRequest* request,
         ::viam::component::board::v1::GetDigitalInterruptValueResponse* response) override;
 
+    ::grpc::Status StreamTicks(
+        ::grpc::ServerContext* context,
+        const ::viam::component::board::v1::StreamTicksRequest* request,
+        ::grpc::ServerWriter<::viam::component::board::v1::StreamTicksResponse>* writer) override;
+
```
stuqdog, line 81: Assuming we're able to act on @acmorrow's suggestion about using `make_service_helper`, this should be marked `noexcept`. If that assumption is false then disregard :)

stuqdog (changes requested):
Generally looks good! I have a couple questions/suggestions and agree with all of @acmorrow's comments.

src/viam/sdk/components/board.hpp
oliviamiller: 1.) sounds good will do!

2.) The idea is that the end user will call `StreamTicks` with the queue parameter, and the ticks that occur will continuously be added to the queue and then consumed by the user. It  I tried to do `std::queue<tick>&` but when I tested with a client script the ticks never got put into the queue, making it a `shared_ptr` worked. Not clear on why that is so possible i'm implementing something wrong and don't actually need shared_ptr.

3.) I used `queue` rather than vector because most likely a user or component will be processing the ticks in the order they came in, and I don't see a super easy way to consume the first in element in a vector and then remove it. I don't see a scenario where someone would want to act on all the ticks at once, and we have the `DigitalInterruptValue` API if someone wants to see how many ticks have occurred.

src/viam/sdk/components/private/board_client.cpp
oliviamiller: I don't think it works, the line `make_client_helper(this, *stub_, &StubType::StreamTicks)` has error  `no instance of function template "viam::sdk::make_client_helper" matches the argument list`. Maybe because its a streaming API?

src/viam/sdk/components/private/board_client.cpp
oliviamiller: It should run continuously until the call is cancelled by client.

src/viam/sdk/components/private/board_server.cpp
oliviamiller: No because this should still run even if the tick queue has nothing in it. If the queue is empty this should still be running as the server should also be listening for new ticks to come into the queue.

src/viam/sdk/components/private/board_client.cpp
oliviamiller: so maybe this should have a try catch to catch any errors and stop running.

src/viam/sdk/components/private/board_server.cpp
oliviamiller: `stream_ticks` should continue running continuously until cancelled. I'm not sure if this will block other calls to the server, can you not have multiple server calls running at the same time?

src/viam/sdk/components/private/board_server.cpp
oliviamiller: changed to use it here.

src/viam/sdk/components/private/board_server.cpp
oliviamiller: linter complained that we don't need the `std::move` but otherwise done.

src/viam/sdk/components/private/board_client.cpp
acmorrow: Quite likely. Would you mind filing a ticket to investigate why make_client_helper doesn't work with streaming? It should be made to, if possible.

src/viam/sdk/components/private/board_server.cpp
acmorrow: Unfortunately, I don't think the API as currently designed is going to work with that model. I will leave additional comments on `board.hpp` about why.

src/viam/sdk/components/private/board_server.cpp
acmorrow: I think there is a bunch of code that can now be removed here, then.

src/viam/sdk/components/board.hpp
```diff
@@ -221,6 +232,22 @@ class Board : public Component {
...
+    inline void stream_ticks(std::vector<std::string> const& digital_interrupt_names,
+                             std::shared_ptr<std::queue<tick>> ticks) {
+        return stream_ticks(digital_interrupt_names, ticks, {});
+    }
+
+    /// @brief Returns a stream of digital interrupt ticks.
+    /// @param digital_interrupt_names digital interrupts to stream
+    /// @param extra Any additional arguments to the method
+    /// @param ticks queue to put the ticks in
+    virtual void stream_ticks(std::vector<std::string> const& digital_interrupt_names,
```
acmorrow, line 247: Now that I understand a little more about the intention here, I think a change of design is likely needed.

My understanding, and please let me know if I have this wrong, is that the caller of `stream_ticks` should not return until the request is cancelled (how?) and that all ticks that are returned during that period should be available to act on _as they occur_.

However, this interface cannot safely be made to do that. The calling thread is blocked, so some other thread must be reacting to the arriving events. To do so, they would need to be consuming those events from the same queue that the thread which called `stream_ticks` is producing the ticks into.

But the queue passed here has no means to synchronize, nor does it have a way to communicate wakeups from the producer (the thread blocked in the streaming call) to the consumer (the thread that wants to pull events from the queue).

Otherwise, with the API here, the only safe thing to do is to:
1) Have one thread call `stream_ticks`, it blocks
2) Someone else cancels the call (how?)
3) The call to `stream_ticks` unblocks, and now you have a queue with all of the ticks that happened.

But I'm fairly sure that isn't what is desired.

I see two ways forward. The first would be to modify this such that the argument to `stream_ticks` is a self-synchronized single-consumer single-producer queue, probably over `Optional<tick>` so there is a way to indicate completion:

```
pc_queue<Optional<Tick>> tick_queue;
std::thread consumer([&tick_queue] {
    auto next = tick_queue.wait();
    if (!next) break;
    std::cout << "Saw a tick: " << ... ;
});
auto board_client = ...
board_client->stream_ticks(pc_queue);
consumer.join();
```

The other way would be to have `stream_ticks` take an on-tick callback.

```
virtual void stream_ticks(std::function<bool(Tick tick)>& tick_handler) = 0;
```

Then you could use it like this:

```
auto board_client = ...
board_client->stream_ticks([](Tick tick) {
    std::cout << "Saw a tick: " << ...;
    if (...) { // Don't want more ticks
       return false;
    return true;
});
```

Where the callback and return a value indicating that it wants no more ticks. Here, the thread doing the streaming also invokes the callback, so some care needs to be taken not to block. The callback, for instance, could then decide to push the ticks to a queue that some other thread is consuming from:

```
pc_queue<Tick> tick_queue;
std::thread consumer([&tick_queue] {
    auto next = tick_queue.wait();
    std::cout << "Saw a tick: " << ... ;
});

auto board_client = ...
board_client->stream_ticks([&queue](Tick tick) {
    queue.push(std::move(tick));
});
consumer.join();
```

There are probably other ways. I'd like to get @stuqdog's thoughts on the matter too so we can design the right API.

acmorrow (changes requested):

src/viam/sdk/components/board.hpp
```diff
@@ -44,6 +45,15 @@ class Board : public Component {
         std::unordered_map<std::string, digital_value> digital_interrupt_values;
     };
 
+    /// @struct tick
+    /// A board digital interrupt that contains high/low value and the time the digital interrupt
+    /// occured.
+    struct Tick {
+        std::string pin_name;
+        std::chrono::nanoseconds time;
```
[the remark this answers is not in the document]
acmorrow: I take it back - keep it `std::chrono::nanoseconds`, but with a comment that it doesn't represent any absolute time.

src/viam/sdk/components/private/board_client.cpp
oliviamiller: https://viam.atlassian.net/browse/RSDK-7318

src/viam/sdk/tests/test_board.cpp
```diff
@@ -146,6 +146,24 @@ BOOST_AUTO_TEST_CASE(test_read_digital_interrupt) {
     });
 }
 
+BOOST_AUTO_TEST_CASE(test_stream_ticks) {
+    const auto mock = std::make_shared<MockBoard>("mock_board");
+
+    client_to_mock_pipeline<Board>(mock, [&](Board& client) {
+        std::function<bool(Board::Tick tick)> tick_handler = [](Board::Tick tick) -> bool {
```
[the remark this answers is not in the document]
oliviamiller: it was complaining before that it wasn't the right type when passed into `stream_ticks`, but now that its passed by reference auto is working .

acmorrow (approved):
A few small things but nothing that needs another round of review for my part.

I would however like @stuqdog to have another look. He had some open comments from an earlier iteration of this work.

stuqdog (approved):
Sorry for delay, was OOO Monday and Tuesday. this lgtm, thanks!

PR #231, opened 2024-05-03
[RSDK-7310] Update the board interface

penguinland:
This turned out to be two problems combined! 1) update the SDK to use a more recent version of the protobufs in the API repo, and then 2) remove the `Status` RPC from the boards. The PR isn't quite done: I'm unsure how this should interact with https://github.com/viamrobotics/viam-cpp-sdk/pull/230, which is where the compiled version of the updated protos is. but the rest I think is ready for a look.

Updating the protobufs was harder than I expected: https://github.com/viamrobotics/api/pull/470 (unrelated to boards) added a new type that had not been used previously, which meant that getting it to work required updating CMakeLists.txt to include the Google libraries for the new type.

Changes to boards:
- `get_status` doesn't exist any more. You can get the standard status from a component the same way you would for a motor, but there isn't a special board-specific one any more.
- In the component status you _can_ get, analog and digital interrupt values are stored as ints, not wrappers around ints. No need to unwrap.
- `get_analog_names` and `get_digital_interrupt_names` can't easily be implemented any more. They don't exist in the Flutter and TypeScript SDKs, and are hopefully being removed from the Go and Python SDKs soon (they have a purpose on the server side, but no obvious use on the client side, so probably shouldn't be in these SDKs). So, I removed them. 
- Similarly, I removed `get_analog_readers` and `get_digital_interrupts`, as they're not present in the Go SDK and would be very hard to implement without the old status.

Everything compiles against a recent version of the API repo, but not against the protobufs copied into this repo itself. How should that be accomplished?

### DIFF
--- a/src/viam/api/CMakeLists.txt
+++ b/src/viam/api/CMakeLists.txt
@@ -341,6 +341,7 @@ target_sources(viamapi
       ${PROTO_GEN_DIR}/../../viam/api/google/api/annotations.pb.h
       ${PROTO_GEN_DIR}/../../viam/api/google/api/http.pb.h
       ${PROTO_GEN_DIR}/../../viam/api/google/api/httpbody.pb.h
+      ${PROTO_GEN_DIR}/../../viam/api/google/rpc/status.pb.h
       ${PROTO_GEN_DIR}/../../viam/api/module/v1/module.grpc.pb.h
       ${PROTO_GEN_DIR}/../../viam/api/module/v1/module.pb.h
       ${PROTO_GEN_DIR}/../../viam/api/robot/v1/robot.grpc.pb.h
--- a/src/viam/api/CMakeLists.txt
+++ b/src/viam/api/CMakeLists.txt
@@ -181,12 +181,98 @@ if (VIAMCPPSDK_USE_DYNAMIC_PROTOS)
     ${PROTO_GEN_DIR}/component/servo/v1/servo.grpc.pb.h
     ${PROTO_GEN_DIR}/component/servo/v1/servo.pb.cc
     ${PROTO_GEN_DIR}/component/servo/v1/servo.pb.h
+    ${PROTO_GEN_DIR}/google/api/annotations.grpc.pb.cc
+    ${PROTO_GEN_DIR}/google/api/annotations.grpc.pb.h
     ${PROTO_GEN_DIR}/google/api/annotations.pb.cc
     ${PROTO_GEN_DIR}/google/api/annotations.pb.h
-    ${PROTO_GEN_DIR}/google/api/httpbody.pb.cc
-    ${PROTO_GEN_DIR}/google/api/httpbody.pb.h
+    ${PROTO_GEN_DIR}/google/api/client.grpc.pb.cc
+    ${PROTO_GEN_DIR}/google/api/client.grpc.pb.h
+    ${PROTO_GEN_DIR}/google/api/client.pb.cc
+    ${PROTO_GEN_DIR}/google/api/client.pb.h
+    ${PROTO_GEN_DIR}/google/api/expr/v1alpha1/checked.grpc.pb.cc
+    ${PROTO_GEN_DIR}/google/api/expr/v1alpha1/checked.grpc.pb.h
+    ${PROTO_GEN_DIR}/google/api/expr/v1alpha1/checked.pb.cc
+    ${PROTO_GEN_DIR}/google/api/expr/v1alpha1/checked.pb.h
+    ${PROTO_GEN_DIR}/google/api/expr/v1alpha1/eval.grpc.pb.cc
+    ${PROTO_GEN_DIR}/google/api/expr/v1alpha1/eval.grpc.pb.h
+    ${PROTO_GEN_DIR}/google/api/expr/v1alpha1/eval.pb.cc
+    ${PROTO_GEN_DIR}/google/api/expr/v1alpha1/eval.pb.h
+    ${PROTO_GEN_DIR}/google/api/expr/v1alpha1/explain.grpc.pb.cc
+    ${PROTO_GEN_DIR}/google/api/expr/v1alpha1/explain.grpc.pb.h
+    ${PROTO_GEN_DIR}/google/api/expr/v1alpha1/explain.pb.cc
+    ${PROTO_GEN_DIR}/google/api/expr/v1alpha1/explain.pb.h
+    ${PROTO_GEN_DIR}/google/api/expr/v1alpha1/syntax.grpc.pb.cc
+    ${PROTO_GEN_DIR}/google/api/expr/v1alpha1/syntax.grpc.pb.h
+    ${PROTO_GEN_DIR}/google/api/expr/v1alpha1/syntax.pb.cc
+    ${PROTO_GEN_DIR}/google/api/expr/v1alpha1/syntax.pb.h
+    ${PROTO_GEN_DIR}/google/api/expr/v1alpha1/value.grpc.pb.cc
+    ${PROTO_GEN_DIR}/google/api/expr/v1alpha1/value.grpc.pb.h
+    ${PROTO_GEN_DIR}/google/api/expr/v1alpha1/value.pb.cc
+    ${PROTO_GEN_DIR}/google/api/expr/v1alpha1/value.pb.h
+    ${PROTO_GEN_DIR}/google/api/expr/v1beta1/decl.grpc.pb.cc
+    ${PROTO_GEN_DIR}/google/api/expr/v1beta1/decl.grpc.pb.h
+    ${PROTO_GEN_DIR}/google/api/expr/v1beta1/decl.pb.cc
+    ${PROTO_GEN_DIR}/google/api/expr/v1beta1/decl.pb.h
+    ${PROTO_GEN_DIR}/google/api/expr/v1beta1/eval.grpc.pb.cc
+    ${PROTO_GEN_DIR}/google/api/expr/v1beta1/eval.grpc.pb.h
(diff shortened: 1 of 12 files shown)


src/viam/examples/modules/complex/proto/buf.lock
```diff
@@ -4,5 +4,4 @@ deps:
   - remote: buf.build
     owner: googleapis
     repository: googleapis
-    commit: ee48893a270147348e3edc6c1a03de0e
-    digest: shake256:a35b0576a2b55dad72747e786af05c03539c2b96be236c9de39fe10d551931ac252eb68445c0cef6bbd27fa20e8c26eee5b8a9fe9c2fde6f63a03e18f8cf980d
+    commit: 4ed3bc159a8b4ac68fe253218760d035
```
penguinland, line 7: This file updated automatically at some point. I don't know what caused it, but I've put the change in a separate commit so it's easy to revert if it's not supposed to be here.

martha-johnston (approved):
as always, thanks for leaving such a detailed description, it was super helpful

src/viam/api/CMakeLists.txt
```diff
@@ -19,6 +19,7 @@ if (VIAMCPPSDK_USE_DYNAMIC_PROTOS)
 endif()
 
 set(BUF_PROTO_COMPONENTS
+  app/mltraining
```
acmorrow, line 22: Agree that this is now needed due to https://github.com/viamrobotics/api/pull/470

src/viam/api/CMakeLists.txt
```diff
@@ -178,12 +181,98 @@ if (VIAMCPPSDK_USE_DYNAMIC_PROTOS)
     ${PROTO_GEN_DIR}/component/servo/v1/servo.grpc.pb.h
     ${PROTO_GEN_DIR}/component/servo/v1/servo.pb.cc
     ${PROTO_GEN_DIR}/component/servo/v1/servo.pb.h
+    ${PROTO_GEN_DIR}/google/api/annotations.grpc.pb.cc
```
acmorrow, line 184: The existing inclusion of things from `google/api` appears only to have pulled in the `.pb` files and not the `grpc.pb` files. Can you see if things still work if you filter out the `grpc` ones?

src/viam/api/CMakeLists.txt
```diff
@@ -178,12 +181,98 @@ if (VIAMCPPSDK_USE_DYNAMIC_PROTOS)
...
+    ${PROTO_GEN_DIR}/google/api/expr/v1beta1/value.grpc.pb.h
+    ${PROTO_GEN_DIR}/google/api/expr/v1beta1/value.pb.cc
+    ${PROTO_GEN_DIR}/google/api/expr/v1beta1/value.pb.h
+    ${PROTO_GEN_DIR}/google/api/field_behavior.grpc.pb.cc
+    ${PROTO_GEN_DIR}/google/api/field_behavior.grpc.pb.h
+    ${PROTO_GEN_DIR}/google/api/field_behavior.pb.cc
+    ${PROTO_GEN_DIR}/google/api/field_behavior.pb.h
+    ${PROTO_GEN_DIR}/google/api/field_info.grpc.pb.cc
+    ${PROTO_GEN_DIR}/google/api/field_info.grpc.pb.h
+    ${PROTO_GEN_DIR}/google/api/field_info.pb.cc
```
acmorrow, line 238: I know it is a pain to do, but this is a *lot* of new code and I'm sure a lot of it is entirely irrelevant to our usage. Could you try pruning each subtree under google/api and see if there are some swaths we can kill off?

acmorrow (changes requested):
Reviewing only build stuff here, I'll leave the actual SDK change to other reviewers, though since they are almost entirely removals it basically looks fine to me.

stuqdog (commented):
Generally looks good to me! It seems we're pulling in a lot of new code though, is it possible to filter some of that down?

src/viam/api/CMakeLists.txt
```diff
@@ -353,6 +487,8 @@ set_target_properties(
   VERSION ${PROJECT_VERSION_MAJOR}.${PROJECT_VERSION_MINOR}.${PROJECT_VERSION_PATCH}
 )
 
+target_link_options(viamapi PRIVATE "LINKER:-z,defs")
```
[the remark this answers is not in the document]
penguinland: I have no clue about CMake. It's here because Drew used it to help debug linker problems, and should be removed now that we're done debugging. Sorry I can't explain more. :shrug:

src/viam/api/CMakeLists.txt
```diff
@@ -178,12 +181,98 @@ if (VIAMCPPSDK_USE_DYNAMIC_PROTOS)
...
+    ${PROTO_GEN_DIR}/google/rpc/context/attribute_context.pb.cc
+    ${PROTO_GEN_DIR}/google/rpc/context/attribute_context.pb.h
+    ${PROTO_GEN_DIR}/google/rpc/error_details.grpc.pb.cc
+    ${PROTO_GEN_DIR}/google/rpc/error_details.grpc.pb.h
+    ${PROTO_GEN_DIR}/google/rpc/error_details.pb.cc
+    ${PROTO_GEN_DIR}/google/rpc/error_details.pb.h
+    ${PROTO_GEN_DIR}/google/rpc/status.grpc.pb.cc
+    ${PROTO_GEN_DIR}/google/rpc/status.grpc.pb.h
+    ${PROTO_GEN_DIR}/google/rpc/status.pb.cc
+    ${PROTO_GEN_DIR}/google/rpc/status.pb.h
```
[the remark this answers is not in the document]
penguinland: This might be my mistake: I had assumed that all the generated grpc code needed to go in here, and this was all new since the last time the C++ SDK was synced with the API repo. I'll try pruning it down...

src/viam/examples/modules/complex/proto/buf.lock
```diff
@@ -4,5 +4,4 @@ deps:
   - remote: buf.build
     owner: googleapis
     repository: googleapis
-    commit: ee48893a270147348e3edc6c1a03de0e
-    digest: shake256:a35b0576a2b55dad72747e786af05c03539c2b96be236c9de39fe10d551931ac252eb68445c0cef6bbd27fa20e8c26eee5b8a9fe9c2fde6f63a03e18f8cf980d
+    commit: 74015a8aeb8445aa9e3e1454cb54bc35
```
penguinland, line 7: I still don't know what this file does, but it changed again. Any ideas?

penguinland (commented):
Take another look: the extra build dependencies are pruned down to just:
- app/mltraining/v1/ml_training.pb.* (dependency added in https://github.com/viamrobotics/api/pull/470), and
- google/rpc/status.pb.*mentioned in ml_training.pb.*

I'm still unsure how this PR should interact with https://github.com/viamrobotics/viam-cpp-sdk/pull/230, to get the latest .proto files over here. Any advice?

src/viam/api/CMakeLists.txt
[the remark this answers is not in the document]
penguinland: You were totally right, thanks for pushing back on this!

src/viam/api/CMakeLists.txt
penguinland: You're right: a lot of the generated protobuf files were unused and could be removed again!

src/viam/api/CMakeLists.txt
```diff
@@ -287,6 +293,7 @@ target_sources(viamapi
     ${PROTO_GEN_DIR}/google/api/annotations.pb.cc
     ${PROTO_GEN_DIR}/google/api/http.pb.cc
     ${PROTO_GEN_DIR}/google/api/httpbody.pb.cc
+    ${PROTO_GEN_DIR}/google/rpc/status.pb.cc
```
acmorrow, line 296: ⬇️ Very likely you also need to add `status.pb.h` down below line 340.

penguinland: I agree that it looks like it should be required. I was able to build without that line, but it feels more proper to add it anyway. I've added it. Things continue to build when using a local copy of the protos, and I still don't know what the plan is to update the ones in this repo itself.

To rebuild things, I've been using this: `cd ..; rm -rf build; mkdir build; cd build; cmake -DVIAMCPPSDK_OFFLINE_PROTO_GENERATION=ON -DVIAMCPPSDK_USE_DYNAMIC_PROTOS=ON -DVIAMCPPSDK_USE_LOCAL_PROTOS=ON .. -G Ninja; ninja -j 5 all` and I imagine that the final version will use fewer options on the `cmake` part.

acmorrow: You would only notice it to be missing if you tried to build a program against an installed build of the SDK.

penguinland:
Github's CI cannot build this because it requires a more recent version of the compiled .proto files from the api repo than what's currently in `src/viam/api/`. What's the procedure for updating that? I still believe the correct versions might be in #230.

penguinland:
I talked to Ethan about how to get the updated protos: he says I should rebase this branch off of #230, then change this PR to pull into that branch instead of `main`. When we're happy here, we merge this into #230, and then merge that PR into `main`. Changes will be coming soon...

penguinland:
Okay, I've rebased and changed the target branch. Let's see how this looks...

acmorrow (approved):

penguinland:
Rebased to get rid of merge conflicts. If the CI checks pass, I'm going to merge this...

src/viam/examples/modules/complex/proto/buf.lock
stuqdog: It sets the buf version for the example module but is setup to auto-update. it's safe to ignore!

PR #232, opened 2024-05-09
[RSDK-7310] Redo PR #231

penguinland:
...which was overwritten by an automated process after merging into the branch on PR 230.

Looks like this is going to have merge conflicts, which I'll try resolving in just a moment...


penguinland:
@stuqdog How does this look? I'm pulling into `workflow/update-protos` again instead of `main`.

PR #234, opened 2024-05-13
RSDK-5378 - allow setting of cpp standard

stuqdog:


acmorrow (approved):
LGTM.

Future work here would be to ensure that we 1) propagate the selected `CMAKE_CXX_STANDARD` as a public requirement on the produced libraries, and 2) continue to do that correctly if one of our dependencies (cough cough abseil) forces the matter.

PR #237, opened 2024-05-14
[RSDK-7151] Update analog reader responses to include accuracy data

penguinland:
Surprisingly, we don't need a protobuf update (from https://github.com/viamrobotics/api/pull/492) because we're already using the new protobufs and have just been ignoring the extra data they hold!

Everything compiles and the tests all pass, though I haven't tried it out more than that.

### DIFF
--- a/src/viam/sdk/components/board.cpp
+++ b/src/viam/sdk/components/board.cpp
@@ -23,14 +23,7 @@ API API::traits<Board>::api() {
 Board::status Board::from_proto(const viam::component::board::v1::Status& proto) {
     Board::status status;
     for (const auto& analog : proto.analogs()) {
-        Board::analog_value result;
-        result.value = analog.second;
-        // The status does not contain the extra data to describe the accuracy of the reader. We
-        // fill those in with 0's instead.
-        result.min_range = 0.0;
-        result.max_range = 0.0;
-        result.step_size = 0.0;
-        status.analog_reader_values.emplace(analog.first, result);
+        status.analog_reader_values.emplace(analog.first, analog.second);
     }
     for (const auto& digital : proto.digital_interrupts()) {
         status.digital_interrupt_values.emplace(digital.first, digital.second);
@@ -57,7 +50,7 @@ Board::power_mode Board::from_proto(viam::component::board::v1::PowerMode proto)
 viam::component::board::v1::Status Board::to_proto(const status& status) {
     viam::component::board::v1::Status proto;
     for (const auto& analog : status.analog_reader_values) {
-        proto.mutable_analogs()->insert({analog.first, analog.second.value});
+        proto.mutable_analogs()->insert({analog.first, analog.second});
     }
 
     for (const auto& digital : status.digital_interrupt_values) {
@@ -85,11 +78,6 @@ Board::Board(std::string name) : Component(std::move(name)){};
 bool operator==(const Board::status& lhs, const Board::status& rhs) {
     return (lhs.analog_reader_values == rhs.analog_reader_values &&
             lhs.digital_interrupt_values == rhs.digital_interrupt_values);
-}
-
-bool operator==(const Board::analog_value& lhs, const Board::analog_value& rhs) {
-    return (lhs.value == rhs.value && lhs.min_range == rhs.min_range &&
-            lhs.max_range == rhs.max_range && lhs.step_size == rhs.step_size);
 }
 
 }  // namespace sdk
--- a/src/viam/sdk/components/board.cpp
+++ b/src/viam/sdk/components/board.cpp
@@ -80,5 +80,12 @@ bool operator==(const Board::status& lhs, const Board::status& rhs) {
             lhs.digital_interrupt_values == rhs.digital_interrupt_values);
 }
 
+bool operator==(const Board::analog_value& lhs, const Board::analog_value& rhs) {
+    return (lhs.value == rhs.value &&
+            lhs.min == rhs.min &&
+            lhs.max == rhs.max &&
+            lhs.step_size == rhs.step_size);
+}
+
 }  // namespace sdk
 }  // namespace viam
--- a/src/viam/sdk/components/board.cpp
+++ b/src/viam/sdk/components/board.cpp
@@ -23,13 +23,13 @@ API API::traits<Board>::api() {
 Board::status Board::from_proto(const viam::component::board::v1::Status& proto) {
     Board::status status;
     for (const auto& analog : proto.analogs()) {
-		Board::analog_value result;
-		result.value = analog.second;
-		// The status does not contain the extra data to describe the accuracy of the reader. We
-		// fill those in with 0's instead.
-		result.min_range = 0.0;
-		result.max_range = 0.0;
-		result.step_size = 0.0;
+        Board::analog_value result;
+        result.value = analog.second;
+        // The status does not contain the extra data to describe the accuracy of the reader. We
+        // fill those in with 0's instead.
+        result.min_range = 0.0;
+        result.max_range = 0.0;
+        result.step_size = 0.0;
         status.analog_reader_values.emplace(analog.first, result);
     }
     for (const auto& digital : proto.digital_interrupts()) {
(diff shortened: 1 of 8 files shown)


src/viam/sdk/components/board.hpp
```diff
@@ -261,6 +267,7 @@ struct API::traits<Board> {
 };
 
 bool operator==(const Board::status& lhs, const Board::status& rhs);
+bool operator==(const Board::analog_value& lhs, const Board::analog_value& rhs);
```
penguinland, line 270: This line is only needed to get line 269 to work. Do we ever actually use the `Board::status` for anything? Part of me wonders if it was supposed to have been removed in https://github.com/viamrobotics/viam-cpp-sdk/pull/231, when `get_status` was removed. Maybe this is all dead code? Maybe not; maybe it needs to stay here.

src/viam/sdk/components/board.cpp
```diff
@@ -23,7 +23,14 @@ API API::traits<Board>::api() {
...
     Board::status status;
     for (const auto& analog : proto.analogs()) {
-        status.analog_reader_values.emplace(analog.first, analog.second);
+		Board::analog_value result;
+		result.value = analog.second;
+		// The status does not contain the extra data to describe the accuracy of the reader. We
+		// fill those in with 0's instead.
+		result.min_range = 0.0;
+		result.max_range = 0.0;
+		result.step_size = 0.0;
```
penguinland, line 32: Discussion question: in the [protobufs](https://github.com/viamrobotics/api/blob/main/proto/viam/component/board/v1/board.proto), this extra data is present in `ReadAnalogReaderResponse` but not in `Status`. Perhaps the `Board::status` here should change from an `unordered_map<string, analog_value>` to an `unordered_map<string, int32_t>` and just continue to hold the raw values without the accuracy information. I'm unsure.

(Part of me wonders if this entire function was supposed to be removed in #231, when board components started using the same status stuff as all other components. I'm unsure if this is dead code...)

src/viam/sdk/components/board.cpp
```diff
@@ -23,7 +23,14 @@ API API::traits<Board>::api() {
...
     for (const auto& analog : proto.analogs()) {
-        status.analog_reader_values.emplace(analog.first, analog.second);
+        Board::analog_value result;
+        result.value = analog.second;
+        // The status does not contain the extra data to describe the accuracy of the reader. We
+        // fill those in with 0's instead.
+        result.min_range = 0.0;
+        result.max_range = 0.0;
+        result.step_size = 0.0;
+        status.analog_reader_values.emplace(analog.first, result);
```
acmorrow, line 33: This can be a one liner:
```
status.analog_reader_values.emplace(analog.first, {analog.second, 0.0, 0.0, 0.0});
```

If you are willing to rely on some C++ initialization nuance, you could I _think_ make this:
```
status.analog_reader_values.emplace(analog.first, {analog.second});
```
which I believe will value initialize the float fields, meaning they become zero.

If that feels too magical, you could also give the non-`value` fields of `analog_value` non-static data member initializers explicitly setting those fields to zero.

src/viam/sdk/components/board.cpp
```diff
@@ -80,5 +87,10 @@ bool operator==(const Board::status& lhs, const Board::status& rhs) {
             lhs.digital_interrupt_values == rhs.digital_interrupt_values);
 }
 
+bool operator==(const Board::analog_value& lhs, const Board::analog_value& rhs) {
+    return (lhs.value == rhs.value && lhs.min_range == rhs.min_range &&
```
acmorrow, line 91: Consider using `std::tie`

src/viam/sdk/components/private/board_client.cpp
```diff
@@ -99,7 +99,12 @@ Board::analog_value BoardClient::read_analog(const std::string& analog_reader_na
     if (!status.ok()) {
         throw GRPCException(status);
     }
-    return response.value();
+    Board::analog_value result = {};
+    result.value = response.value();
+    result.min_range = response.min_range();
+    result.max_range = response.max_range();
+    result.step_size = response.step_size();
+    return result;
```
acmorrow, line 107: This can also be a one liner:
```
return {response.value(), response.min_range(), response.max_range(), response.step_size()};
```

src/viam/sdk/components/board.cpp
```diff
@@ -23,7 +23,14 @@ API API::traits<Board>::api() {
 Board::status Board::from_proto(const viam::component::board::v1::Status& proto) {
     Board::status status;
     for (const auto& analog : proto.analogs()) {
-        status.analog_reader_values.emplace(analog.first, analog.second);
+        Board::analog_value result;
+        result.value = analog.second;
+        // The status does not contain the extra data to describe the accuracy of the reader. We
+        // fill those in with 0's instead.
```
acmorrow, line 29: Should the accuracy component of `analog_reader` be optional?

acmorrow (commented):
Just wandering by, do with these comments what you like.

purplenicole730 (approved):
Drew's comments can clean up the code a bit, but otherwise, LGTM!

src/viam/sdk/components/board.cpp
penguinland: No, but there is some tension between the data obtained from reading an analog pin, and the data obtained from viewing a board's status. Part of this is resolvable by getting rid of a lot of the explicit status-reporting stuff, and it's still unclear to me whether this function is dead code. Perhaps the `Board::status` should contain just ints as the values in the map, rather than entire `Board::analog_value`s?

src/viam/sdk/components/board.cpp
penguinland: ...I think the way to actually resolve this tension is that the data stored in the status should not change: although we now get extra data in responses from reading analogs, we do _not_ get any such data in status messages, so should keep the old version of that.

src/viam/sdk/components/board.cpp
penguinland: TIL. Thanks!

src/viam/sdk/components/private/board_client.cpp
penguinland: I always get nervous using `{...}` syntax in C++: did I put the items it the right order? What if someone later changes the order (e.g., alphabetizes the fields in the struct)? I can't read this without also pulling up the header file to check against. but I guess here it's pretty benign: the order is the one that makes the most sense, and each value has a name that tells you what it is. Thanks for pushing me to question my assumptions!

src/viam/sdk/components/private/board_server.cpp
```diff
@@ -103,8 +103,11 @@ ::grpc::Status BoardServer::ReadAnalogReader(
         extra = struct_to_map(request->extra());
     }
 
-    const Board::analog_value result = board->read_analog(request->analog_reader_name(), extra);
-    response->set_value(result);
+    const Board::analog_response result = board->read_analog(request->analog_reader_name(), extra);
```
penguinland, line 106: I named the type `analog_response`, but named all the variables `result`. Should the type be `analog_result` instead? I'm unsure...

penguinland (commented):
I've resolved my tension between statuses holding just an int32 while the analog responses now hold 4 pieces of data: keep statuses as just an int32, and only change the return type when reading analogs! This got rid of half the things Drew suggested needed fixing, and I like the design better. Take another look!

penguinland:
The CI build is failing with `Failure: decode buf.lock: no digest specified for module buf.build/googleapis/googleapis`, and retrying the tests leads to the same error. Any idea what changed? I didn't think I had touched that. and it builds fine on my local machine (`mkdir build; cd build; cmake .. -G Ninja; ninja -j 5 all`), so I can't reproduce the problem.

stuqdog:
> The CI build is failing with `Failure: decode buf.lock: no digest specified for module buf.build/googleapis/googleapis`, and retrying the tests leads to the same error. Any idea what changed? I didn't think I had touched that. and it builds fine on my local machine (`mkdir build; cd build; cmake .. -G Ninja; ninja -j 5 all`), so I can't reproduce the problem.

@penguinland not sure what's going on with this but it's not a problem on your end; it's popping up in other PRs as well. I'll investigate.

penguinland:
Yup, I rebased off of `main` and everything works now! Thanks for looking into it.

PR #238, opened 2024-05-16
RSDK-7463 - Update Motion service in C++ SDK

nfranczak:
We are updating the python SDK to match the following changes made in API (https://github.com/viamrobotics/api/commit/503a9a41e4ef9c8e5c8399a8890c6b2ae923d96b)

We would like to also explicitly call out the changes reviewers should be aware of:

1. GeoObstacle has been renamed into GeoGeometry - we note that they store the same information and this is just a re-name.
This means that the return type of GetObstacles in the navigation service has changes from GeoObstacle into GeoGeometry, but the navigation service is not yet a thing in this SDK so nothing to worry about here :)
2. We have added an argument to the motion service method MoveOnGlobe. The argument is called bounding_regions and is a list of GeoGeometrys
3. The motion service method MoveOnGlobe also takes in a list of obstacles of type GeoObstacle. This has been updated to GeoGeometry

### DIFF
--- a/src/viam/sdk/tests/mocks/mock_motion.hpp
+++ b/src/viam/sdk/tests/mocks/mock_motion.hpp
@@ -17,8 +17,9 @@ sdk::Name fake_component_name();
 sdk::Name fake_slam_name();
 sdk::Name fake_movement_sensor_name();
 sdk::geo_point fake_geo_point();
-std::vector<sdk::geo_obstacle> fake_obstacles();
+std::vector<sdk::geo_geometry> fake_obstacles();
 std::shared_ptr<sdk::motion_configuration> fake_motion_configuration();
+std::vector<sdk::geo_geometry> fake_bounding_regions();
 
 class MockMotion : public sdk::Motion {
    public:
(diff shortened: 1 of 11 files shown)


src/viam/sdk/tests/mocks/mock_motion.hpp
```diff
@@ -89,10 +91,11 @@ class MockMotion : public sdk::Motion {
     std::string peek_destination_frame;
     double peek_heading;
     bool peek_stop_plan_called = false;
-    std::vector<sdk::geo_obstacle> peek_obstacles;
+    std::vector<sdk::geo_geometry> peek_obstacles;
     std::vector<sdk::GeometryConfig> peek_map_obstacles;
     std::shared_ptr<constraints> peek_constraints;
     std::shared_ptr<sdk::motion_configuration> peek_motion_configuration;
+    std::vector<sdk::geo_geometry> peek_bounding_regions;
```
stuqdog, line 98: We've added this as a field here but we don't seem to actually set it or peek it anywhere. We should follow the example of how `peek_obstacles` is used (i.e., set `peek_bounding_regions` in the `move_on_globe` request and then verify that it's been set in `test_motion.cpp`).

stuqdog (changes requested):
One requested change for tests, otherwise the code looks good to me.

We will need to figure out why the tests are failing before we merge this, however.

stuqdog (approved):

PR #242, opened 2024-05-31
[RSDK-7775]   Update Camera.Properties Response in C++ SDK to match API

hexbabe:
[RSDK-7775](https://viam.atlassian.net/browse/RSDK-7775)

[RSDK-7775]: https://viam.atlassian.net/browse/RSDK-7775?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

From Jira^: Right now, we’re not returning the mime type. Update the C++ SDK wrapper to match the Properties API method

In this PR, I add `mime_types` to the camera component files, extracting the data from the proto when `Camera::properties Camera::from_proto` is called, adding it to the new `mime_types` field in camera properties. I added an appropriate test to make sure this happens correctly.

### DIFF
--- a/src/viam/sdk/components/camera.cpp
+++ b/src/viam/sdk/components/camera.cpp
@@ -241,7 +241,7 @@ Camera::properties Camera::from_proto(
         proto.intrinsic_parameters();
     intrinsic_parameters = from_proto(intrinsic_parameters_proto);
 
-    mime_types.assign(proto.mime_types().begin(), proto.mime_types().end()); 
+    mime_types.assign(proto.mime_types().begin(), proto.mime_types().end());
 
     properties.distortion_parameters = distortion_parameters;
     properties.intrinsic_parameters = intrinsic_parameters;
--- a/src/viam/sdk/components/camera.cpp
+++ b/src/viam/sdk/components/camera.cpp
@@ -228,27 +228,10 @@ Camera::distortion_parameters Camera::from_proto(
 
 Camera::properties Camera::from_proto(
     const viam::component::camera::v1::GetPropertiesResponse& proto) {
-    Camera::distortion_parameters distortion_parameters;
-    Camera::intrinsic_parameters intrinsic_parameters;
-    Camera::mime_types mime_types;
-    Camera::properties properties;
-
-    const viam::component::camera::v1::DistortionParameters& distortion_parameters_proto =
-        proto.distortion_parameters();
-    distortion_parameters = from_proto(distortion_parameters_proto);
-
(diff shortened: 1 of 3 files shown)


src/viam/sdk/tests/mocks/camera_mocks.cpp
```diff
@@ -86,11 +86,18 @@ Camera::distortion_parameters fake_distortion_parameters() {
     return distortion_parameters;
 }
 
+Camera::mime_types fake_mime_types() {
+    Camera::mime_types mime_types;
+    mime_types.push_back("JPEG");
+    return mime_types;
```
acmorrow, line 92: Or just `return { "JPEG" };`

src/viam/sdk/components/camera.cpp
```diff
@@ -230,6 +230,7 @@ Camera::properties Camera::from_proto(
     const viam::component::camera::v1::GetPropertiesResponse& proto) {
     Camera::distortion_parameters distortion_parameters;
     Camera::intrinsic_parameters intrinsic_parameters;
+    Camera::mime_types mime_types;
```
acmorrow, line 233: You didn't write this function this way, so if you don't fee like changing it I'll understand, but I'd expect the entire function body could be replaced with this:
```
return {
    proto.supports_pcd(),
    from_proto(proto.intrinsic_parameters()),
    from_proto(proto.distortion_parameters()),
    { proto.mime_types().begin(), proto.mime_types().end() }
};
```

See https://godbolt.org/z/oaPs318hf

src/viam/sdk/components/camera.hpp
```diff
@@ -48,16 +48,22 @@ class Camera : public Component {
...
     };
 
+    /// @brief The supported mime types of the camera— a type alias.
+    using mime_types = std::vector<std::string>;
+
     /// @struct properties
     /// @brief The camera's supported features and settings.
     ///
-    /// `supports_pcd` indicates whether the camera has a valid implementation of `get_point_cloud`.
+    /// `supports_pcd` indicates whether the camera has a valid implementation of
```
acmorrow, line 57: Consider moving these doc comments onto each field

acmorrow (approved):
LGTM with an entirely optional improvement if you are interested, and a few nits.

acmorrow (approved):

stuqdog (approved):

PR #248, opened 2024-06-17
replace deprecated CreateMessage with Create

lia-viam:


njooma (approved):

src/viam/sdk/services/private/mlmodel_client.cpp
```diff
@@ -36,13 +46,15 @@ std::shared_ptr<MLModelService::named_tensor_views> MLModelServiceClient::infer(
...
-    auto* const req = pb::Arena::CreateMessage<mlpb::InferRequest>(arena.get());
+    auto* const req = pb::Arena::VIAM_SDK_PB_CREATE_MESSAGE<mlpb::InferRequest>(arena.get());
 
     req->set_name(this->name());
     *req->mutable_extra() = map_to_struct(extra);
-    auto* const resp = pb::Arena::CreateMessage<mlpb::InferResponse>(arena.get());
+    auto* const resp = pb::Arena::VIAM_SDK_PB_CREATE_MESSAGE<mlpb::InferResponse>(arena.get());
     ClientContext ctx;
 
+#undef VIAM_SDK_PB_CREATE_MESSAGE
```
acmorrow, line 56: Personally, I'd be OK with not bothering to undefine here in .cpp context since the macro isn't going to leak and conflicts arising later in the file seem very unlikely.

acmorrow (approved):

acmorrow:
I tested this locally on my mac where I have the newer proto via homebrew. Without this PR I reproduce the failure seen in CI when trying to release 0.0.7, but with the change here I can build again and tests pass. I think this is good to land and unblock a 0.0.7.

src/viam/sdk/services/private/mlmodel_client.cpp
lia-viam: I think you're right, just removed it

PR #249, opened 2024-06-18
RSDK-7939: Add arm component

lia-viam:
See [RSDK-7939](https://viam.atlassian.net/browse/RSDK-7939)

Maybe the only odd thing about this is the `KinematicsData` variant, a design I arrived at after discussion with @acmorrow 

It's an adaptation of https://github.com/viamrobotics/api/blob/main/proto/viam/common/v1/common.proto#L194

The variant is meant to introduce different types for different file types. In the case of unspecified data we do not store the bytes, but maybe there is a compelling reason to do so. 

Note also that this is technically part of common but I put it in the arm module code because it's not being used elsewhere, but I can move it if that's an issue

[RSDK-7939]: https://viam.atlassian.net/browse/RSDK-7939?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/viam/sdk/components/arm.hpp
+++ b/src/viam/sdk/components/arm.hpp
@@ -31,7 +31,7 @@ class Arm : public Component, public Stoppable {
     struct RawBytes {
         // Pre c++17 our derived classes aren't aggregate initializable so we need to define
         // and using declare some ctors
-        
+
         RawBytes() = default;
         RawBytes(std::vector<unsigned char> b) : bytes(std::move(b)) {}
 
--- a/src/viam/sdk/components/arm.hpp
+++ b/src/viam/sdk/components/arm.hpp
@@ -64,13 +64,14 @@ class Arm : public Component, public Stoppable {
     static KinematicsData from_proto(const viam::common::v1::GetKinematicsResponse& proto);
 
     /// @brief Get the current position of the end of the arm.
+    /// @return The `pose` representing the end position of the arm.
     inline pose get_end_position() {
         return get_end_position({});
     }
 
     /// @brief Get the current position of the end of the arm.
-    /// @param pose The destination pose for the arm.
     /// @param extra Any additional arguments to the method.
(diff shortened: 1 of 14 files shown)


src/viam/sdk/components/arm.hpp
```diff
@@ -0,0 +1,142 @@
...
+/// @class Arm arm.hpp "components/arm.hpp"
+/// @brief An `Arm` represents a physical robot arm that exists in three-dimensional space.
+/// @ingroup Arm
+///
+/// This acts as an abstract parent class to be inherited from by any drivers representing
+/// specific arm implementations. This class cannot be used on its own.
+class Arm : public Component, public Stoppable {
+    // Base class for use below in defining kinematics data strong typedefs
+    template <class Tag>
+    struct RawBytes {
```
stuqdog, line 31: (minor, style) the idiom for `struct`s in the SDK is `snake_case`

src/viam/sdk/components/arm.hpp
```diff
@@ -0,0 +1,142 @@
...
+
+    static KinematicsData from_proto(const viam::common::v1::GetKinematicsResponse& proto);
+
+    /// @brief Get the current position of the end of the arm.
+    inline pose get_end_position() {
+        return get_end_position({});
+    }
+
+    /// @brief Get the current position of the end of the arm.
+    /// @param pose The destination pose for the arm.
```
stuqdog, line 72: (minor) this should be `@return` and below the `extra` param. also probably should be included with the documentation for the `inline` version as well for consistency's sake.

stuqdog (commented):
Nice, a couple small style things otherwise looks good to me!

One other note, we'll probably want to leave certain params (ctx, some `response`s) unnamed in `arm_server` to make sure the linter test passes.

stuqdog (approved):

PR #250, opened 2024-06-20
RSDK-7642 - add simple module client for module integration testing

stuqdog:
Creates a simple module client for use in module integration tests.


lia-viam (approved):

src/viam/examples/modules/simple/CMakeLists.txt
```diff
@@ -25,3 +25,18 @@ install(
   TARGETS simple_module
   COMPONENT examples
 )
+
+add_executable(simple_module_client)
+target_sources(simple_module_client
```
lia-viam, line 30: just for my own curiosity is there a reason we do it this way (also for `add_library` calls in the sdk) instead of just `add_executable(simple_module_client client.cpp)`

acmorrow: That's my doing. I like being consistent about using the `target_xyz` calls to affect the target rather than having some of them apply at the declaration and others (e.g. `target_link_libraries`) be called for side effects. Since `target_sources` works like `target_link_libraries` I used it consistently everywhere, and left the `add_executable` and `add_library` calls are purely introducing the name with no side effects.

lia-viam: cool that makes sense!

PR #251, opened 2024-06-21
RSDK-7880 - add orientation config getters

stuqdog:


src/viam/sdk/spatialmath/orientation.hpp
lia-viam, line 1: Maybe I'm missing something, but were the return types for the non-`const` overloads supposed to have been mutable reference? As is we have `const` overloads returning a reference to `const` and mutable overloads returning a copy

stuqdog: Yeah good question! I think I didn't want to have either of these provide a mutable reference, the goal was:

* when we have a `const` `OrientationConfig`, provide a memory-efficient `const&` getter
* when we have a non-`const` `OrientationConfig`, provide a copy so that the `Config` itself won't get mutated

Was not planning on adding a setter here so a mutable reference seemed not great. Definitely open to feedback on if you think there's a better approach here! 

Also cc @hexbabe who flagged the need for getters in the first place, do you think there's a need for setters as well?

lia-viam: Ah I see what you're saying! In that case my recommendation would be to delete the non-`const` overloads, as they're not necessary to accomplish the intent you described above. Note that for
> provide a copy so that the Config itself won't get mutated

This is already accomplished by the overloads returning `const &`, and the user can opt in or out of whether they want a memory-efficient getter at callsite. For example

```cpp
const OrientationConfig cfg;

auto value_copy = cfg.get_value(); // makes a copy
const auto& value_ref = cfg.get_value();
```
Both of these would be calling `OrientationConfig::get_value() const` and there's no way to mutate the underlying config regardless. 

As for setters, no I don't think we necessarily need those if you hadn't meant to put them in! See eg https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Rh-get

lia-viam: Oh and all that said, the only minor nit I would suggest is that `get_type() const` should just return `OrientationType` by value, because it's an `enum` (ie trivial copyable integer type) so there's nothing to be gained by returning a reference

stuqdog: Awesome, thanks both!

lia-viam (approved):

PR #258, opened 2024-06-26
RSDK-8031 Pose tracker component

lia-viam:
https://viam.atlassian.net/browse/RSDK-8031


src/viam/sdk/components/private/pose_tracker_server.cpp
```diff
@@ -0,0 +1,63 @@
...
+PoseTrackerServer::PoseTrackerServer(std::shared_ptr<ResourceManager> manager)
+    : ResourceServer(std::move(manager)) {}
+
+::grpc::Status PoseTrackerServer::GetPoses(
+    ::grpc::ServerContext*,
+    const ::viam::component::posetracker::v1::GetPosesRequest* request,
+    ::viam::component::posetracker::v1::GetPosesResponse* response) noexcept {
+    return make_service_helper<PoseTracker>(
+        "PoseTrackerServer::GetPoses", this, request)([&](auto& helper, auto& pose_tracker) {
+        const std::vector<std::string> body_names(
```
stuqdog, line 24: It doesn't look like we're actually using `body_names` anywhere, I think we can get rid of this.

stuqdog (approved):
One minor thing, otherwise lgtm!

src/viam/sdk/components/private/pose_tracker_server.cpp
lia-viam: oops yes good catch!

PR #265, opened 2024-07-10
Streamable client helper

lia-viam:
Modifies `ClientHelper` to be able to handle streaming requests such as `Board::stream_ticks` or forthcoming `InputController::stream_events`. This work was factored out of the input controller branch, and the new method is able to handle both of them. 

`ClientHelper` is now templated on the stub PMF type, which is one of two introduced alias declarations, either `SyncMethodType` or `StreamingMethodType`. This replaces the `using PFn` declaration. Rather than do this, I would love to be able to write something like
```cpp
using PFn = std::conditional_t<
    is_sync_response_v<ResponseType>
    SyncMethodType<...>,
    StreamingMethodType<...>
>
```

but at a glance it does not appear possible to do through anything gRPC gives us on response types. 

Note it would be possible to forgo the `SyncMethodType` and `StreamingMethodType` alias decls entirely and just have a single `make_client_helper` which is templated on `typename MethodType` and let type deduction do the rest, but following the previous implementation I decided to opt for a more verbose but more explicit approach where the helper overloads make it clear which template parameters are accepted. 

Based on the return type of stream responses it seems that an `ErrorHandlerCallable` version is not required but I could be wrong.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,8 +1,9 @@
 # Please try to keep this file as empty as possible,
 # and document what cannot be avoided.
 
-# Give users an ignored place to build
-/build/*
+# Give users an ignored place to build, either a /build/ directory or one with 
+# config parameters such as /build-arm-Debug/
+/build*/*
 
 # For now, ignore any viam_rust_utils library in the root.
 /libviam_rust_utils*
@@ -12,3 +13,6 @@
 
 # ignore doxygen output
 /etc/docs/api/*
+
+# ignore vscode customization
+/.vscode/*
--- a/.gitignore
+++ b/.gitignore
@@ -1,9 +1,8 @@
 # Please try to keep this file as empty as possible,
 # and document what cannot be avoided.
 
-# Give users an ignored place to build, either a /build/ directory or one with 
-# config parameters such as /build-arm-Debug/
-/build*/*
+# Give users an ignored place to build
+/build/*
 
 # For now, ignore any viam_rust_utils library in the root.
 /libviam_rust_utils*
@@ -13,6 +12,3 @@
 
 # ignore doxygen output
 /etc/docs/api/*
-
-# ignore vscode customization
-/.vscode/*
--- a/src/viam/sdk/common/client_helper.hpp
+++ b/src/viam/sdk/common/client_helper.hpp
@@ -88,13 +88,20 @@ class ClientHelper {
 
         auto reader = (stub_->*pfn_)(ctx, request_);
 
-        while (reader->Read(&response_) && rhc(response_)) {
+        bool cancelled_by_handler = false;
+
+        while (reader->Read(&response_)) {
+            if (!rhc(response_)) {
+                cancelled_by_handler = true;
+                static_cast<::grpc::ClientContext*>(ctx)->TryCancel();
+                break;
+            }
         }
 
-        static_cast<::grpc::ClientContext*>(ctx)->TryCancel();
-
         const auto result = reader->Finish();
-        if (result.ok() || result.error_code() == ::grpc::StatusCode::CANCELLED) {
+
+        if (result.ok() ||
+            (cancelled_by_handler && result.error_code() == ::grpc::StatusCode::CANCELLED)) {
             return;
         }
 
--- a/src/viam/sdk/common/client_helper.hpp
+++ b/src/viam/sdk/common/client_helper.hpp
@@ -80,8 +80,9 @@ class ClientHelper {
     // A version of invoke for gRPC calls returning `(stream ResponseType)`.
     // ResponseHandlerCallable will be called for every response in the reader, and should return
     // false to indicate it is no longer interested in the stream.
-    template <typename ResponseHandlerCallable>
-    auto invoke_stream(ResponseHandlerCallable rhc) {
+    template <typename ResponseHandlerCallable,
+              typename ErrorHandlerCallable = decltype(default_ehc_)>
+    auto invoke_stream(ResponseHandlerCallable rhc, ErrorHandlerCallable&& ehc = default_ehc_) {
         *request_.mutable_name() = client_->name();
         ClientContext ctx;
 
@@ -89,6 +90,16 @@ class ClientHelper {
 
         while (reader->Read(&response_) && rhc(response_)) {
         }
+
+        static_cast<::grpc::ClientContext*>(ctx)->TryCancel();
+
+        const auto result = reader->Finish();
+        if (result.ok() || result.error_code() == ::grpc::StatusCode::CANCELLED) {
+            return;
+        }
+
+        std::forward<ErrorHandlerCallable>(ehc)(result);
+        client_helper_details::errorHandlerReturnedUnexpectedly(result);
     }
 
    private:
--- a/src/viam/sdk/common/client_helper.hpp
+++ b/src/viam/sdk/common/client_helper.hpp
@@ -14,14 +14,16 @@ namespace client_helper_details {
 [[noreturn]] void errorHandlerReturnedUnexpectedly(const ::grpc::Status&) noexcept;
 }  // namespace client_helper_details
 
-template <class StubType, class RequestType, class ResponseType>
-using StreamingMethodType = std::unique_ptr<::grpc::ClientReaderInterface<ResponseType>> (
-    StubType::*)(::grpc::ClientContext*, const RequestType&);
-
+// Method type for a gRPC call that returns a response message type.
 template <class StubType, class RequestType, class ResponseType>
 using SyncMethodType = ::grpc::Status (StubType::*)(::grpc::ClientContext*,
                                                     const RequestType&,
                                                     ResponseType*);
+
+// Method type for a gRPC call that returns a stream of response message type.
+template <class StubType, class RequestType, class ResponseType>
+using StreamingMethodType = std::unique_ptr<::grpc::ClientReaderInterface<ResponseType>> (
+    StubType::*)(::grpc::ClientContext*, const RequestType&);
 
 template <typename ClientType,
           typename StubType,
--- a/src/viam/sdk/common/client_helper.hpp
+++ b/src/viam/sdk/common/client_helper.hpp
@@ -77,6 +77,9 @@ class ClientHelper {
         client_helper_details::errorHandlerReturnedUnexpectedly(result);
     }
 
+    // A version of invoke for gRPC calls returning `(stream ResponseType)`.
+    // ResponseHandlerCallable will be called for every response in the reader, and should return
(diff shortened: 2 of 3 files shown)


.gitignore
```diff
@@ -12,3 +13,6 @@
 
 # ignore doxygen output
 /etc/docs/api/*
+
+# ignore vscode customization
```
acmorrow, line 17: Could this go in `~/.gitignore` just as well?

.gitignore
```diff
@@ -1,8 +1,9 @@
 # Please try to keep this file as empty as possible,
 # and document what cannot be avoided.
 
-# Give users an ignored place to build
-/build/*
+# Give users an ignored place to build, either a /build/ directory or one with 
+# config parameters such as /build-arm-Debug/
+/build*/*
```
acmorrow, line 6: Is there a reason that building to `build/arm-Debug` doesn't work? I'd prefer to keep this narrow. I find that without pressure to keep it absolutely minimal, `.gitignore` quickly becomes a dumping ground and impossible to clean.

src/viam/sdk/common/client_helper.hpp
```diff
@@ -21,10 +38,7 @@ class ClientHelper {
     }
 
    public:
-    using PFn = ::grpc::Status (StubType::*)(::grpc::ClientContext*,
-                                             const RequestType&,
-                                             ResponseType*);
-    explicit ClientHelper(ClientType* client, StubType* stub, PFn pfn)
+    explicit ClientHelper(ClientType* client, StubType* stub, MethodType pfn)
```
acmorrow, line 41: Is there some way to constrain `MethodType` to ensure that it matches `SyncMethodType` or `StreamingMethodType`?

src/viam/sdk/common/client_helper.hpp
```diff
@@ -63,21 +77,52 @@ class ClientHelper {
...
 
+    // A version of invoke for gRPC calls returning `(stream ResponseType)`.
+    // ResponseHandlerCallable will be called for every response in the reader, and should return
+    // false to indicate it is no longer interested in the stream.
+    template <typename ResponseHandlerCallable>
+    auto invoke_stream(ResponseHandlerCallable rhc) {
+        *request_.mutable_name() = client_->name();
+        ClientContext ctx;
+
+        auto reader = (stub_->*pfn_)(ctx, request_);
```
acmorrow, line 88: Looking more carefully at [grpc::ClientReaderInterface](https://grpc.github.io/grpc/cpp/classgrpc_1_1_client_reader_interface.html), I see that there is a `Finish` call with seems related to error handling / status of the RPC. I wonder if we were failing to interact with that properly in the existing code, and whether this helper is the opportunity to do the right thing?

src/viam/sdk/common/client_helper.hpp
```diff
@@ -63,21 +77,52 @@ class ClientHelper {
...
+
+        while (reader->Read(&response_)) {
```
acmorrow, line 90: Or even, maybe: `while (reader->Read(&response_) && rhc(response))`

src/viam/sdk/common/client_helper.hpp
acmorrow: Actually, nevermind: the helper functions sort of already do that.

src/viam/sdk/common/client_helper.hpp
```diff
@@ -12,7 +14,22 @@ namespace client_helper_details {
 [[noreturn]] void errorHandlerReturnedUnexpectedly(const ::grpc::Status&) noexcept;
 }  // namespace client_helper_details
 
-template <typename ClientType, typename StubType, typename RequestType, typename ResponseType>
+// Method type for a gRPC call that returns a response message type.
+template <class StubType, class RequestType, class ResponseType>
```
acmorrow, line 18: Any reason for `class` here rather than `typename` in the template parameter list, and below? I know they mean the same thing, basically, but I think the codebase is pretty consistent in favor of `typename` in that role.

acmorrow (changes requested):
Looks great! Mostly little things, though I do have one more general question about error handling.

src/viam/sdk/common/client_helper.hpp
lia-viam: I'm glad you asked! So yes the helper functions sort of already do it, I think we could constrain it further by defining the method typedefs as traits classes and then we could `static_assert` and/or define `ClientHelper` as a partial template specialization, but that might be getting a bit silly versus the "good enough" helper function approach.

src/viam/sdk/common/client_helper.hpp
lia-viam: ah very nice

src/viam/sdk/common/client_helper.hpp
lia-viam: oh whoops you're right, no reason just force of habit, I'll switch it to `typename`

src/viam/sdk/common/client_helper.hpp
lia-viam: I think you're right, the [docs](https://grpc.github.io/grpc/cpp/classgrpc_1_1internal_1_1_client_streaming_interface.html#a567f213cc0a5df2a03becda3e5711e25) suggest to me that this should be called after the `while` loop. If we drained the reader it will return a status immediately, and if our stream handler terminated the reading it will flush the remaining messages and then return a status.

lia-viam: This was surprisingly not forthcoming info, but it looks like `TryCancel` is the way to do it, and then a call to `Finish` should return an OK or Cancelled status, see eg 

https://groups.google.com/g/grpc-io/c/R0NTqKaHLdE?pli=1
https://github.com/grpc/grpc/issues/17292#issuecomment-460773219
https://stackoverflow.com/a/54560700

src/viam/sdk/common/client_helper.hpp
acmorrow: I think we go with "good enough" here. If you pass the wrong sort of thing it is unlikely to compile anyway. And if it *does* compile, well, then maybe that's OK too. Resolving.

src/viam/sdk/components/private/board_client.cpp
```diff
@@ -135,26 +135,17 @@ Board::digital_value BoardClient::read_digital_interrupt(const std::string& digi
...
-        }
-    };
+    return make_client_helper(this, *stub_, &StubType::StreamTicks)
+        .with(extra,
+              [&](auto& request) {
+                  for (const auto& name : digital_interrupt_names) {
+                      request.add_pin_names(name);
+                  }
+              })
+        .invoke_stream([&](auto& response) {
```
acmorrow, line 145: In light of the findings around `TryCancel` and `Finish`, I think there is now an identified correctness issue with any other streaming client call that isn't making use of those. Can you have a look around to try to see if there are others and just pull them into using the new helper?

src/viam/sdk/common/client_helper.hpp
acmorrow: I'm learning this here with you, so I have a few more questions:
- Is it possible for (stub->*pfn) to return an empty `unique_ptr` such that `reader` is not valid? If that happens, is there an error to be obtained from the `ClientContext` in some way?
- I wonder if (and sorry, this will mean undoing an earlier suggestion of mine) the case of `reader->Read(&response)` returning false and `rhc` returning `false` now need to be handled differently. In particular, if `reader->Read` returns false, I believe it means no more messages are coming, so a `Cancel` may not be needed (but an error check is!), whereas if `rhc` returns `false` it means the users callback no longer wants more messages, so a `TryCancel` is probably needed.

lia-viam: To the first point, this was again not forthcoming, but it looks like no: the stub calls its own `StreamEventsRaw` method, which then calls `internal::ClientReaderFactory<R>::Create`, and [that call](https://github.com/grpc/grpc/blob/master/include/grpcpp/support/sync_stream.h#L170) is always a `new` construction

Second point, I mulled that over for a bit and found someone asking in more detail but not really getting a helpful reply, because imo this ticket points out an ambiguity/absence in docs:
https://github.com/grpc/grpc/issues/22839
Anyway I'll spare you all the permutations I was considering but yes maybe it's better to revert to separate condition checking. I *suspect* that it's fine to call `TryCancel` when the stream already finished, but we may as well only do it if the reader is dry anyway.

src/viam/sdk/components/private/board_client.cpp
lia-viam: I think the only ones so far are 
- `Board`, changed in this PR
- `InputController`, forthcoming but based off this PR as well
- `AudioInput`, not yet started
there's also the gizmo example, should I change that too?

I checked for other streaming protos, I think none other are in robot components so they're not covered:
https://github.com/search?q=repo%3Aviamrobotics%2Fapi+path%3A%2F%5Eproto%5C%2F%2F+stream&type=code

src/viam/sdk/common/client_helper.hpp
```diff
@@ -63,21 +77,67 @@ class ClientHelper {
...
+        ClientContext ctx;
+
+        auto reader = (stub_->*pfn_)(ctx, request_);
+
+        bool cancelled_by_handler = false;
+
+        while (reader->Read(&response_)) {
+            if (!rhc(response_)) {
+                cancelled_by_handler = true;
+                static_cast<::grpc::ClientContext*>(ctx)->TryCancel();
```
acmorrow, line 96: Why do you need the static_cast here?

src/viam/sdk/components/private/board_client.cpp
acmorrow: Great, thanks for checking, I wasn't sure.

src/viam/sdk/common/client_helper.hpp
acmorrow: OK, I think the logic looks pretty solid to me now. Resolving.

acmorrow (approved):
LGTM mod a question about the static cast.

src/viam/sdk/common/client_helper.hpp
lia-viam: It's so I can call `grpc::ClientContext::TryCancel`, because `ClientContext ctx` above is `viam::sdk::ClientContext` which wraps the grpc context as a private member, and whose only public interface is a user-defined implicit conversion to `grpc::ClientContext*`, presumably so the stub method can be called without any explicit casts or operators

acmorrow: Got it. Thanks for explaining. I knew that once but forgot. I may have even suggested it be that way...

PR #272, opened 2024-07-26
RSDK-6636: Refactored AttributeMap with type-erased ProtoType

lia-viam:
This is a proof of concept/RFC refactor of the ProtoType class using C++ type erasure.

See eg 
https://davekilian.com/cpp-type-erasure.html
https://learn.microsoft.com/en-us/shows/goingnative-2013/inheritance-base-class-of-evil (deeper dive)

This is a pretty straightforward implementation which borrows heavily from Sean Parent's sample code. 

The code is insulated from the rest of the codebase for now, although there are now unit tests which exercise more or less a superset of the existing tests for `ProtoType`. 

I have omitted `struct_to_map` and `map_to_struct` but they could easily be added. 

If we like this approach, an obvious next step would be to rewrite the interfaces to insulate proto from our API/ABI. This would be pretty easy, I think we could do something like

```cpp
namespace google {
namespace protobuf {
class Value;
}
}

class ProtoT {
public:
    // ...
    friend void to_proto_value(const ProtoT& t, google::protobuf::Value* v);
    // ...
};

template<class U = google::protobuf::Value>
U to_proto_value(const ProtoT& t) {
    U u;
    to_proto_value(t, &u);
    return u;
}
```

Which could then be called as `google::protobuf::Value v = to_proto_value(ProtoT(5))` in any class which includes the requisite header. 

A less obvious next step would be optimizations like SBO storage so as not to incur a heap allocation eg for objects where `sizeof(T) <= sizeof(double)`.


stuqdog (approved):
LGTM though I confess I'm a bit out of my depth with this one, would definitely be interested in seeing @acmorrow's thoughts.

lia-viam:
Per discussion with @acmorrow , updated the PR to have this class insulate proto from our API/ABI. I've added tests for copy and move construction, plus a helper class to make moves and default construction better behaved.

I also added a casting API similar to LLVM's, and added tests for that as well. 
https://llvm.org/docs/ProgrammersManual.html#the-isa-cast-and-dyn-cast-templates

src/viam/sdk/common/proto_t.hpp
```diff
@@ -0,0 +1,293 @@
...
+}  // namespace protobuf
+}  // namespace google
+
+namespace viam {
+namespace sdk {
+
+// Type erased value for storing protobuf Value types.
+// The "simple" value types are empty, bool, int, double, and string.
+// Moreover, a ProtoT can be a vector or string-map of ProtoT.
+class ProtoT {
```
acmorrow, line 26: I know the existing type is called ProtoType, but in google terminology this stuff is known as Struct and Value. Could we find names that are more aligned? Like `ProtoT` is `ProtoValue` and the map typename is `ProtoStruct`?

src/viam/sdk/common/proto_t.hpp
```diff
@@ -0,0 +1,293 @@
...
+    // Checking cast to T, returns non-owning non-null pointer if argument is_a<T>()
+    template <typename T>
+    friend T* dyn_cast(ProtoT&);
+
+    // Checking cast to T, returns non-owning non-null pointer if argument is_a<T>()
+    template <typename T>
+    friend T const* dyn_cast(const ProtoT&);
+
+   private:
+    struct vtable_t {
```
acmorrow, line 104: While reviewing, I'm getting a little tangled up between `vtable_t`, `model_t`, and `storage_t` and `kind_t` and how they interrelate / what purposes they serve. Maybe a more elaborated class comment for `ProtoT` (or whatever it becomes named) would be useful.

acmorrow, line 104: Do the trailing `_t`'s actually serve a purpose, or could the names be simplified to `vtable`, `storage`, etc.

src/viam/sdk/common/proto_t.hpp
```diff
@@ -0,0 +1,293 @@
...
+        void (*dtor)(void*) noexcept;
+        void (*copy)(void const*, void*);
+        void (*move)(void*, void*) noexcept;
+        void (*to_proto_value)(void const*, google::protobuf::Value*);
+        int (*kind)() noexcept;
+        bool (*equal_to)(void const*, void const*, const vtable_t&);
+    };
+
+    template <typename T>
```
acmorrow, line 113: The universe of `T`'s for which `model_t` is instantiated is closed, I think? And it is only used internally by ProtoT. It makes me wonder if the definition could be moved to .cpp scope and whether explicit instantiations could be made ahead of time.

src/viam/sdk/common/proto_t.hpp
```diff
@@ -0,0 +1,293 @@
...
+    struct model_t {
+        model_t(T t) : data(std::move(t)) {}
+
+        static void dtor(void* self) noexcept {
+            static_cast<model_t*>(self)->~model_t();
+        }
+
+        static void copy(void const* self, void* dest) {
+            new (dest) model_t(*static_cast<model_t const*>(self));
```
acmorrow, line 122: I'm having vague thoughts about std::launder here due to seeing placement new, but I need to remind myself about what that does and how it works.

src/viam/sdk/common/proto_t.hpp
```diff
@@ -0,0 +1,293 @@
...
+
+    ProtoT(const google::protobuf::Value* value);
+
+    vtable_t vtable_;
+    storage_t self_;
+};
+
+using AttrMap = std::unordered_map<std::string, ProtoT>;
+
+void to_proto_value(std::nullptr_t, google::protobuf::Value* v);
```
acmorrow, line 208: Should these just be canonically called `to_proto` with the idea that that's the same name we would use elsewhere?

acmorrow (commented):
Focusing mostly on the header here, I'm still trying to fully understand the machinery, but these are the comments I have right now.

src/viam/sdk/common/proto_t.hpp
lia-viam: They can be simplified!

src/viam/sdk/common/proto_t.hpp
lia-viam: Good call, we can definitely do this with explicit instantiations.

src/viam/sdk/common/proto_t.hpp
lia-viam: I like that, and then the named constructor can be called `from_proto`

src/viam/sdk/common/proto_t.hpp
```diff
@@ -0,0 +1,293 @@
...
+// The "simple" value types are empty, bool, int, double, and string.
+// Moreover, a ProtoT can be a vector or string-map of ProtoT.
+class ProtoT {
+   public:
+    // Construct a null object.
+    ProtoT() noexcept : ProtoT(nullptr) {}
+
+    // Construct a nonempty object.
+    template <typename T>
+    ProtoT(T t) : vtable_{model_t<T>::vtable}, self_{std::move(t)} {}
```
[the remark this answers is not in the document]
lia-viam: I think so yes!

src/viam/sdk/common/proto_t.hpp
```diff
@@ -0,0 +1,293 @@
...
+    template <typename T>
+    friend T* dyn_cast(ProtoT&);
+
+    // Checking cast to T, returns non-owning non-null pointer if argument is_a<T>()
+    template <typename T>
+    friend T const* dyn_cast(const ProtoT&);
+
+   private:
+    struct vtable_t {
+        void (*dtor)(void*) noexcept;
```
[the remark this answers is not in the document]
lia-viam: Ah yup good catch

src/viam/sdk/common/proto_t.hpp
```diff
@@ -0,0 +1,293 @@
...
+    // Construct a nonempty object.
+    template <typename T>
+    ProtoT(T t) : vtable_{model_t<T>::vtable}, self_{std::move(t)} {}
+
+    // Deduction helper constructor for string from string literal
+    ProtoT(const char* str) : ProtoT(std::string(str)) {}
+
+    // Move construction this from other, leaving other in the "unspecified but valid state" of its
+    // moved-from type
+    ProtoT(ProtoT&& other) noexcept
```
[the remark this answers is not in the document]
lia-viam: You're right, I think the unordered_map screws up `noexcept`ness for us

src/viam/sdk/common/proto_t.hpp
```diff
@@ -0,0 +1,293 @@
...
+        : vtable_(std::move(other.vtable_)), self_(std::move(other.self_), vtable_) {}
+
+    ProtoT(const ProtoT& other) : vtable_(other.vtable_), self_(other.self_, other.vtable_) {}
```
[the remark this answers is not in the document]
lia-viam: We can do that!

src/viam/sdk/common/proto_t.hpp
lia-viam: Ah right so probably a helpful thing to link in the comments would be this Sean Parent blog post or the diff from where a similar implementation is used in Adobe's code:

https://stlab.cc/tip/2017/12/23/small-object-optimizations.html
https://github.com/stlab/libraries/pull/130/files#diff-f3b7b621d95fa28949c56df8f3376872440c4761b4107c381eadf052e55bcc02

And also Louis Dionne's local storage class, on which this storage class is based:
https://github.com/ldionne/dyno/blob/master/include/dyno/storage.hpp#L389

The idea is we need to BYO vtable because the kind of object storage/access we're doing is UB for polymorphic types

As for the code itself yes I will definitely document these further

src/viam/sdk/common/proto_t.hpp
```diff
@@ -0,0 +1,293 @@
...
+            return *static_cast<T const*>(self) == *static_cast<T const*>(other);
+        }
+
+        static constexpr vtable_t vtable{dtor, copy, move, to_proto_value, kind, equal_to};
+        T data;
+    };
+
+    struct storage_t {
+        static constexpr std::size_t small_size =
+            sizeof(std::unordered_map<std::string, std::string>);
```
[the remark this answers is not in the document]
lia-viam: A better name for it might be `local_storage_size`, it serves as the buffer size for the storage class. It's the size of a string-string map because the `unordered_map<string, ProtoT>` class is the largest class in our universe of types, but `ProtoT` is an incomplete type at time of declaration of the `storage` class, and `sizeof(unordered_map<K, V>)` doesn't depend on `K` or `V`. That said this could do with a comment and probably a `static_assert` to make sure what I've said holds since I think it's not guaranteed by the standard.

src/viam/sdk/common/proto_t.hpp
```diff
@@ -0,0 +1,293 @@
...
+        storage_t(storage_t&& other, const vtable_t& vtable) {
+            vtable.move(other.get(), this->get());
+        }
+
+        void swap(const vtable_t& this_vtable, storage_t& other, const vtable_t& other_vtable) {
+            if (this == &other) {
+                return;
+            }
+
+            unsigned char tmp[small_size];
```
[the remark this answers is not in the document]
lia-viam: I think you're right, was confusing myself a bit reading that `aligned_storage` has recently been deprecated, but realized that they mean the issue is people writing `aligned_storage<N> my_storage` rather than `aligned_storage_t<N> my_storage` etc

src/viam/sdk/common/proto_t.hpp
lia-viam: I was wondering about that myself, especially since std::launder didn't exist pre-c++17, though people were implementing variant and optional types all the while. This discussion from the boost mailing list suggests we're probably fine:
https://groups.google.com/g/boost-developers-archive/c/-ivUay0RxQw

src/viam/sdk/common/proto_t.hpp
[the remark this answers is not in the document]
lia-viam: Update: it looks like `vector` and `unordered_map` are nothrow move constructible at least on Apple Clang, so I've done a little type trait fiddling to make it so we can condition the `noexcept`ness of this depending on whether all the possible stored types are nothrow movable

src/viam/sdk/common/proto_t.hpp
lia-viam: That makes sense to me!

src/viam/sdk/common/proto_value.hpp
```diff
@@ -0,0 +1,324 @@
...
+
+class Value;
+class Struct;
+}  // namespace protobuf
+}  // namespace google
+
+namespace viam {
+namespace sdk {
+
+namespace impl {
```
acmorrow, line 25: I think we are keeping `impl` for the private namespace, maybe this should be `details`, which I think we use elsewhere for publicly visible implementation details.

src/viam/sdk/common/proto_value.hpp
```diff
@@ -0,0 +1,324 @@
...
+
+struct move_may_throw {
+    move_may_throw() noexcept(false) = default;
+    move_may_throw& operator=(move_may_throw&&) noexcept(false) = default;
```
acmorrow, line 29: Do you need move construction in here too?

src/viam/sdk/common/proto_value.hpp
```diff
@@ -0,0 +1,324 @@
...
+                                 std::is_nothrow_move_constructible<
+                                     std::unordered_map<std::string, move_may_throw>>{}> {};
+
+}  // namespace impl
+
+/// @brief Type-erased value for storing google::protobuf::Value types.
+/// A ProtoValue can be nullptr, bool, int, double, std::string, or, recursively, a vector or
+/// string-map of ProtoValue.
+/// This type is used at API/ABI boundaries for interfacing with grpc/proto code.
+class ProtoValue {
```
acmorrow, line 50: I think it is great how much of this was able to exist out of the header.

src/viam/sdk/common/proto_value.hpp
```diff
@@ -0,0 +1,324 @@
...
+    friend bool operator==(const ProtoValue& lhs, const ProtoValue& rhs);
+
+    void swap(ProtoValue& other) noexcept(impl::all_proto_moves_noexcept{});
+
+    /// @brief Construct from proto value
+    /// @note This method is trivially templated to insulate google::protobuf::Value from our
+    /// API/ABI. It is meant to be called with no template parameters in a translation unit which
+    /// includes <google/protobuf/struct.pb.h>
+    template <typename Value = google::protobuf::Value>
+    static ProtoValue from_proto(const Value& v);  // NOLINT(misc-no-recursion)
```
acmorrow, line 88: Do you still need this `NOLINT`?

src/viam/sdk/common/proto_value.hpp
```diff
@@ -0,0 +1,324 @@
...
+        bool (*equal_to)(void const*, void const*, const vtable&);
+    };
+
+    // Class template for a concrete model of the concept described by vtable.
+    // All construction and equality operations are implemented using defaults of the stored T.
+    template <typename T>
+    struct model {
+        model(T t) noexcept(std::is_nothrow_move_constructible<T>{});
+
+        static void dtor(void* self) noexcept;
```
acmorrow, line 135: Do you want a static assertion that T is nothrow destructible?

src/viam/sdk/common/proto_value.hpp
```diff
@@ -0,0 +1,324 @@
...
+        // stand-ins for std::vector<ProtoValue> and ProtoStruct which are not available until end
+        // of class definition.
+        using BufType = std::aligned_union_t<0,
+                                             std::nullptr_t,
+                                             bool,
+                                             int,
+                                             double,
+                                             std::string,
+                                             std::vector<void*>,
+                                             std::unordered_map<std::string, void*>>;
```
acmorrow, line 168: I see why you need to use `void*` as a standin. If there isn't one already, could there be a static assertion later that the sizes are the same whether or not it is `void*` or `ProtoStruct`?

src/viam/sdk/common/proto_value.hpp
```diff
@@ -0,0 +1,324 @@
...
+    template <typename T>
+    bool is_a() const;
+
+    /// @brief Checking cast to T, returns non-owning non-null pointer if argument is_a<T>()
+    template <typename T>
+    friend T* dyn_cast(ProtoValue&);
+
+    /// @brief Checking cast to T, returns non-owning non-null pointer if argument is_a<T>()
+    template <typename T>
+    friend T const* dyn_cast(const ProtoValue&);
```
acmorrow, line 108: Are these, effectively, the main accessors? Why make them friend casts rather than just regular `get` members?

acmorrow (changes requested):
This is lookin a lot cleaner and I like how well the implementation was able to be moved to the .cpp file and out of the header. Some comments and questions.

src/viam/sdk/common/proto_value.hpp
acmorrow: Actually, even better, add some static assertions at .cpp scope once you have access to the full `ProtoValue` type.

src/viam/sdk/common/proto_value.hpp
```diff
@@ -0,0 +1,324 @@
...
+
+    ///@}
```
acmorrow, line 110: Do you want to think about supporting a visitation API?

src/viam/sdk/common/proto_value.hpp
```diff
@@ -0,0 +1,324 @@
...
+/// @note This method is trivially templated to insulate google::protobuf::Struct from our
+/// API/ABI. It is meant to be called with no template parameters in a translation unit which
+/// includes <google/protobuf/struct.pb.h>
+template <typename Struct = google::protobuf::Struct>
+Struct map_to_struct(const ProtoStruct& m) {
+    Struct s;
+    map_to_struct(m, &s);
+
+    return s;
+}
```
acmorrow, line 268: Can we use this to work out what the general shape for all "SDK type" to "proto type" interconversions will look like? I suspect we want some sort of trait that binds the two together as well.

acmorrow (commented):
A few more thoughts that I had overnight...

src/viam/sdk/common/proto_value.hpp
lia-viam: Oops yes that's a typo, first one should have been move construction

src/viam/sdk/common/proto_value.hpp
lia-viam: Same! And only one round of explicit insantiations needed, a bunch of other ones happen implicitly.

I was wondering if *all* of it could be implicit instantiations due to the `ProtoValue(google::protobuf::Value*)` constructor which calls all the other ones, but the linker didn't like that

src/viam/sdk/common/proto_value.hpp
lia-viam: Unfortunately yes https://github.com/viamrobotics/viam-cpp-sdk/actions/runs/10321482772/job/28574485369

I think this may come down to a clang-tidy version thing, running it locally complained about the `NOLINT` in the cpp file at the definition rather than wanting it at the declaration.

src/viam/sdk/common/proto_value.hpp
lia-viam: Not sure if it's necessary? In our closed universe of types that's always true, and also throwing destructors are UB anyway right

src/viam/sdk/common/proto_value.hpp
lia-viam: There is the following in the cpp file, maybe this is adequate?
https://github.com/viamrobotics/viam-cpp-sdk/blob/b659c767e9d3d250692bf3a2a50598e78266d252/src/viam/sdk/common/proto_value.cpp#L146-L148

I think that more accurately names the object of concern here, ie, 
- is the storage large enough for any type being constructed?
- is there any type being constructed with alignment too strict for the storage alignment?

We could assert on the `sizeof(ProtoStruct)` too though, what do you think?

src/viam/sdk/common/proto_value.hpp
lia-viam: They are, yes! It's kind of semantic and I'm not opposed to calling it `get` instead, I suppose `get` is a better fit for the mental model of "this is kind of like a JSON value type" rather than casting which evokes polymorphism.

src/viam/sdk/common/proto_value.hpp
lia-viam: Not sure if I follow, could you say more?

src/viam/sdk/common/proto_value.hpp
lia-viam: That had occurred to me, I'm open to it! Should be pretty straightforward to write

src/viam/sdk/common/proto_value.hpp
acmorrow: I think it is worth explicitly validating that `alignof` and `sizeof` for `std::vector<void*>` and `std::unordered_map<std::string, void*>` match the `alignof` and `sizeof` for those some types with `ProtoStruct` as the mapped type.

src/viam/sdk/common/proto_value.hpp
acmorrow: For instance, in `board.hpp` we have these:

```
    /// @brief Creates a `power_mode` enum from its proto representation.
    static power_mode from_proto(viam::component::board::v1::PowerMode proto);

    /// @brief Converts a `power_mode` enum to its proto representation.
    static viam::component::board::v1::PowerMode to_proto(power_mode power_mode);
```

But I think we agree that we want to eliminate the "by size" usage of the proto types in our API, and move towards something more like what is being done in this review, where they are all defined in terms of opaque pointers:

```
static power_mode from_proto(viam::component::board::v1::PowerMode const* proto);
static to_proto(power_mode power_mode, viam::component::board::v1::PowerMode* proto);
```

But then we also want some templates to wrap this up more nicely so that if you have opted into having the `v1::PowerMode` type available you can use it with a better API:

```
v1::PowerMode to_proto(const power_mode& pm) {
    v1::PowerMode proto;
    to_proto(pm, &proto);
    return proto;
}

power_mode from_proto(const v1::PowerMode& proto) {
    return from_proto(&proto);
}
```

And I have some hand-wavy idea about whether there should be some sort of traits class which ties the two together:

```
template<typename ProtoType>
struct associated_sdk_type;

template<>
struct associated_sdk_type<v1::PowerMode> {
    using type = power_mode;
}

template<typename SdkType>
struct associated_proto_type;

template<>
struct associated_proto_type<power_mode> {
    using type = v1::PowerMode;
}
```

Because then maybe all the `to_proto` and `from_proto` things become one template? Or maybe you only need the trait for one direction because overloading handles the other?

But, generally, I'm interested in whether we can use this review to get the protocol sorted out that will be used globally in the SDK.

src/viam/sdk/common/proto_value.hpp
acmorrow: Well, the `static_assert` doesn't cost much. It is possible to have `noexcept(false)` dtors. We don't have any of those right now I believe, and I think it is unlikely that we would. But I think I'd prefer to know if somehow one materialized unexpectedly in the future.

src/viam/sdk/common/proto_value.hpp
```diff
@@ -0,0 +1,324 @@
...
+ProtoStruct struct_to_map(const Struct& s) {
+    return struct_to_map(&s);
+}
+
+/// @brief Convert a ProtoStruct to a google::protobuf::Struct.
+/// @note This method is trivially templated to insulate google::protobuf::Struct from our
+/// API/ABI. It is meant to be called with no template parameters in a translation unit which
+/// includes <google/protobuf/struct.pb.h>
+template <typename Struct = google::protobuf::Struct>
+Struct map_to_struct(const ProtoStruct& m) {
```
acmorrow, line 263: These `map_to_struct` and `struct_to_map` names are the same as those for the existing `ProtoType`. Is there going to be a conflict/ambiguity if they are both in scope? Or are they intended to be drop-in replacements as part of a strategy of rolling over to this implementation? I was never much of a fan of the names using `map` anyway.

acmorrow (commented):
I think this is looking pretty solid and can probably be merged soon. I would though like to think about what the plan would be to roll the SDK over to using the new types. Should it be flag guarded at build time? Could the old type become a wrapper over the new and everything just works? Basically, what's the plan for making this the real version?

src/viam/sdk/common/proto_value.hpp
lia-viam: Answering this and the question above about roll-out strategy:

For a couple reasons I think this may need to be a one-fell-swoop approach:
- Old `ProtoType` and `AttributeMap` code is littered with `std::make_shared` and lots of pointer dereferencing, null checks, etc. I think this would make a wrapper approach more trouble than it's worth.
- A lot of use of `AttributeMap` originates from the root of an inheritance hierarchy with a lot of leaves, e.g., `do_command` in all the component classes, which then is coupled with the corresponding server implementations.

That said, adding `map_to_struct` and `struct_to_map` was motivated by an audit of the existing code, and in those cases the replacement might be a drop-in, modulo a pointer deref in some cases

src/viam/sdk/common/proto_value.hpp
lia-viam: Ahh I see, let me experiment a bit and I'll report back!

acmorrow (approved):
This LGTM as a starting point and the work on it can continue in other PRs. Maybe take it out of the CMakeLists before merging so it isn't live in the codebase until later.

PR #279, opened 2024-08-14
RSDK-8541: ProtoValue get_unchecked and visit API

lia-viam:
Adds an `unchecked_get` API to get the stored value in a `ProtoValue`, and a `visit` API in the spirit of `std::visit` or similar. 

Tests both `unchecked_get` and `visit`. Note that the `ProtoValue` tests have been moved into their own source file now. In this diff all that's new in `test_proto_value.cpp` is the `get_unchecked` tests. The `visit` tests are in their own test source file.


src/viam/sdk/common/proto_value.hpp
```diff
@@ -273,33 +297,56 @@ Struct map_to_struct(const ProtoStruct& m) {
...
 
 template <>
-struct kind_t<bool> : std::integral_constant<int, 1> {};
+struct kind<bool> {
+    using type = std::integral_constant<int, 1>;
+};
 
 template <>
-struct kind_t<int> : std::integral_constant<int, 2> {};
+struct kind<int> {
```
acmorrow, line 316: It just occurred to me that proto's struct doesn't actually have an integer type. Just double. Do we want to be in the business of defining that conversion, or should we push that to users, since we cannot always know their intent.

src/viam/sdk/common/proto_value.hpp
```diff
@@ -273,33 +297,56 @@ Struct map_to_struct(const ProtoStruct& m) {
...
+}  // namespace proto_value_details
+
+/// @brief ProtoValue RTTI type trait.
+/// This type trait is used to implement the ProtoValue::kind method which provides the type
+/// discriminator constant that is used in the value access API.
+/// @remark A ProtoValue can only be constructed from types for which this trait is well formed.
+/// This type trait can be used to induce substitution failure on non-ProtoValue constructible
+/// types.
+template <typename T>
+using kind_t = typename proto_value_details::kind<T>::type;
```
acmorrow, line 349: Do you need the `_t`?

src/viam/sdk/common/proto_value_visit.hpp
```diff
@@ -0,0 +1,95 @@
+#pragma once
+
+#include <viam/sdk/common/proto_value.hpp>
+
+#include <type_traits>
```
acmorrow, line 5: This one should come earlier than `proto_value.hpp`

src/viam/sdk/common/proto_value_visit.hpp
```diff
@@ -0,0 +1,95 @@
...
+            return std::forward<Visitor>(visitor)(value.get_unchecked<int>());
+        case kind_t<double>{}:
+            return std::forward<Visitor>(visitor)(value.get_unchecked<double>());
+        case kind_t<std::string>{}:
+            return std::forward<Visitor>(visitor)(value.get_unchecked<std::string>());
+        case kind_t<std::vector<ProtoValue>>{}:
+            return std::forward<Visitor>(visitor)(value.get_unchecked<std::vector<ProtoValue>>());
+        case kind_t<ProtoStruct>{}:
+            return std::forward<Visitor>(visitor)(value.get_unchecked<ProtoStruct>());
+        default:
```
acmorrow, line 37: Is a `default` needed at all? Should it result in calling the nullptr case for visitor? I feel like it shouldn't be possible or should be treated as some sort of fatal error.

src/viam/sdk/tests/test_proto_value.cpp
```diff
@@ -0,0 +1,366 @@
...
+#include <unordered_map>
+
+#include <google/protobuf/struct.pb.h>
+
+#include <boost/mp11/algorithm.hpp>
+#include <boost/mp11/integer_sequence.hpp>
+#include <boost/mp11/tuple.hpp>
+#include <boost/test/included/unit_test.hpp>
+
+#include <viam/sdk/common/proto_value.hpp>
```
acmorrow, line 13: It is a good habit to make the header under test the first include in the associated test, since it demonstrates that the header compiles in isolation. Usually we follow that practice in the `.cpp` for something as well, so it matters less for, say `proto_value.hpp` than it does for `proto_value_visit.hpp`, but generally just worth doing in my opinion.

src/viam/sdk/tests/test_proto_value_visit.cpp
```diff
@@ -0,0 +1,166 @@
...
+#include <tuple>
+#include <type_traits>
+
+#include <boost/mp11/algorithm.hpp>
+#include <boost/mp11/list.hpp>
+#include <boost/mp11/tuple.hpp>
+#include <boost/test/included/unit_test.hpp>
+
+#include <viam/sdk/common/proto_value.hpp>
+#include <viam/sdk/common/proto_value_visit.hpp>
```
acmorrow, line 12: Let's make this the very first header, in its own group.

src/viam/sdk/common/proto_value.cpp
```diff
@@ -104,6 +104,56 @@ bool ProtoValue::is_null() const {
...
+    return *(this->self_.template get<T>());
+}
+
+template bool& ProtoValue::get_unchecked<bool>();
+template int& ProtoValue::get_unchecked<int>();
+template double& ProtoValue::get_unchecked<double>();
+
+template bool ProtoValue::get_unchecked<bool>() const;
+template int ProtoValue::get_unchecked<int>() const;
+template double ProtoValue::get_unchecked<double>() const;
```
acmorrow, line 125: I think these should be declared in the header as `extern template` so consumers of the header know that they can avoid instantiating them.

src/viam/sdk/common/proto_value.cpp
```diff
@@ -104,6 +104,56 @@ bool ProtoValue::is_null() const {
...
+template std::vector<ProtoValue>& ProtoValue::get_unchecked<std::vector<ProtoValue>>() &;
+template ProtoStruct& ProtoValue::get_unchecked<ProtoStruct>() &;
+
+template std::string const& ProtoValue::get_unchecked<std::string>() const&;
+template std::vector<ProtoValue> const& ProtoValue::get_unchecked<std::vector<ProtoValue>>() const&;
+template ProtoStruct const& ProtoValue::get_unchecked<ProtoStruct>() const&;
+
+template std::string&& ProtoValue::get_unchecked<std::string>() &&;
+template std::vector<ProtoValue>&& ProtoValue::get_unchecked<std::vector<ProtoValue>>() &&;
+template ProtoStruct&& ProtoValue::get_unchecked<ProtoStruct>() &&;
```
acmorrow, line 155: Ditto

acmorrow (approved):
LGTM, but please address the things I raised here before merging. If you want me to take another look before merging just re-request my review, otherwise, I don't think another round of review is required unless there are substantive changes required to address any of my comments. I don't think there will be.

src/viam/sdk/common/proto_value.hpp
lia-viam: I've been wondering about this too--proto just has a `number` type which promotes everything to double. There are some gotchas in the unit tests about this:
https://github.com/lia-viam/viam-cpp-sdk/blob/feature/proto-value-visit/src/viam/sdk/tests/test_proto_value.cpp#L43

src/viam/sdk/common/proto_value.hpp
lia-viam: I suppose not, more just convention

src/viam/sdk/common/proto_value_visit.hpp
lia-viam: It's "needed" just to not make the compiler complain, but in theory it's not possible by the invariants we maintain on the proto value objects. We could do a `BOOST_UNREACHABLE_RETURN` or similar.

src/viam/sdk/common/proto_value.hpp
lia-viam: Actually just tried doing this change, this made the compiler complain because of the ambiguity with the `kind()` method, I think we might be better off keeping the `_t`, or possibly using capital `Kind`, actually not super clear what our convention is here as I think it varies a bit throughout the SDK

lia-viam:
@acmorrow just to sum up for re-review, addressed most of the comments, further decision probably needed on the "number" type and on if we want to put `unreachable` in our `switch` statements, which also affects existing code.

src/viam/sdk/common/proto_value_visit.hpp
acmorrow: Would having `kind` return an enum over the available types placate the compiler about not having a default?

acmorrow (changes requested):
This looks great. I'm really happy with how little implementation lives in `proto_value.hpp` now, and the new visitation functions look great.

I think the only things left to do are:
- See if it is possible to eliminate the default case
- Decide about how to handle int (personal preference to disallow it and make the caller figure it out).

src/viam/sdk/common/proto_value.hpp
lia-viam: I'm in favor of getting rid of the separate `int` and `double` types and replacing them with a `number` type which is just a `double` underneath, as this would match what's in proto
https://protobuf.dev/reference/protobuf/google.protobuf/#value

Deleting the conversion from `int` entirely is tempting, but maybe this will cause us headaches when migrating from `ProtoType`? The `variant` underlying that class also has distinct `int` and `double`, which is arguably a mistake for the same reason, but there will definitely be instances of `std::make_shared<ProtoType>(5)`

src/viam/sdk/common/proto_value.hpp
lia-viam: Good point!

src/viam/sdk/common/proto_value_visit.hpp
lia-viam: I think that would work actually, I'll give it a shot

src/viam/sdk/common/proto_value.hpp
acmorrow: OK, so several thoughts have come from this discussion:

- Does `ProtoValue` really need a `template` constructor, or should it just have constructors for each type. As written, you will just get a funky probably-link-time error because it can't instantiate it if you try to, say, construct a `ProtoValue` of some unrelated type.

- Should (some of the) `ProtoValue` constructors (templated or not) be `explicit`?

- I suspect that if we eliminate the ability to hold integers, then the one numeric constructor should take `double`, but allow implicit conversions (so, not explicit). That'd let `std::make_shared<ProtoType>(5)` still work, I think. What I want to avoid is confusing situations where by accepting `int` as a constructor argument, we allow things with conversion to `int` to be passed, but then those things become doubles internally.

lia-viam: All good questions. My thoughts
- I may as well just spell out a bunch of non-`template` constructors for each type, although I think I would still want to have them defer to a private `template` constructor. Whatever amount of duplication is present it would still be fewer lines than all the explicit instantiations I'm writing
- I think yeah ideally we would want to follow the example of JSON libraries, where you have sort of easy ergonomic interfaces for building up objects like [this](https://www.boost.org/doc/libs/1_86_0/libs/json/doc/html/json/quick_look.html), which you get because the [constructors](https://www.boost.org/doc/libs/1_85_0/libs/json/doc/html/json/ref/boost__json__value/value.html) are not `explicit`.
- Yeah I think this makes sense and we will just be deferring to standard C++ language rules about type promotion/argument overloading

All that said, maybe I should make a separate PR to do
- have only `number` type
- define non-`template` public constructors

and then rebase this one off that one?

acmorrow: Yes, let's go forward with what you have now where `int` is a thing, and then defer the whole `int` removal and any associated constructor fussing into a followup PR. I don't think you need to do the rebase part.

lia-viam: OK cool so just to confirm should I still do the `kind_t` removal and changing to an `enum Kind` for this PR too?

acmorrow: Yes, I think the `kind_t` removal is pretty trivial, and I'd really like to see the `default` case go away if possible, so I think the enum thing makes more sense in this review than in a new one.

src/viam/sdk/common/proto_value.hpp
```diff
@@ -47,8 +47,14 @@ struct all_moves_noexcept
 /// A ProtoValue can be nullptr, bool, int, double, std::string, or, recursively, a vector or
 /// string-map of ProtoValue.
 /// This type is used at API/ABI boundaries for interfacing with grpc/proto code.
+/// @remark The (special) member function templates below all operate on a closed subset of types,
+/// so we provide explicit instantiations for all valid template types. See below the class
+/// definition.
 class ProtoValue {
    public:
+    /// @brief Type discriminator constants for possible values stored in a ProtoValue.
+    enum Kind { null, bool_, int_, double_, string, list, struct_ };
```
acmorrow, line 56: - We usually prefix constants with `k_` (google style thing, but one I like), so `k_null`, `k_bool` will all work I think.
- I think maybe you should give these explicit numeric values since they form part of the interface.
- Maybe make the order the same order as the struct proto does: https://github.com/protocolbuffers/protobuf/blob/main/src/google/protobuf/struct.proto#L66-L76. Throw `int` at the end since we expect to remove it.

acmorrow (commented):
This LGTM mod some suggested / minor tweaks to the new enum. No need to do another round of review for those changes unless you want to.

I like the new typedef for `ListValue` too.

acmorrow (approved):

src/viam/sdk/common/proto_value.hpp
lia-viam: Sounds good, going to do all these except keep the same order we have (but with `int` thrown at the end still). The reason for the proto sequencing escapes me, and I think it's easier mnemonically to write switch statements and sequence the methods if it goes from least to most data-rich types which is also what you see, eg, in JSON library APIs usually

PR #282, opened 2024-08-19
Fix aarch build

lia-viam:
This should fix failures seen it the latest draft release, see eg 
https://github.com/viamrobotics/viam-cpp-sdk/actions/runs/10457244069/job/28956432798#step:7:326
https://gcc.gnu.org/bugzilla/show_bug.cgi?id=54483#c13

For whatever reason this was only failing on the aarch c++14 linux builds, although it seems they were correct to fail it.

### DIFF
--- a/src/viam/sdk/common/proto_value.hpp
+++ b/src/viam/sdk/common/proto_value.hpp
@@ -223,6 +223,7 @@ class ProtoValue {
     storage self_;
 };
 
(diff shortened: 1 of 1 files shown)


lia-viam:
See https://github.com/viamrobotics/viam-cpp-sdk/actions/runs/10457957698 for that this fixes the issue

stuqdog (approved):

PR #284, opened 2024-08-29
APP-6070 Update apis.json to use new fields

DTCurrie:
Updates the `apis.json` file to use the new fields we use for code sample generation. See similar changes for the [Python SDK](https://github.com/viamrobotics/viam-python-sdk/pull/635), [RDK](https://github.com/viamrobotics/rdk/pull/4180), and [TS-SDK](https://github.com/viamrobotics/viam-typescript-sdk/pull/343).

It's important to note the TS-SDK and CPP-SDK had a departure from other SDKs where they declared a `client` field. I have updated that to use `importName` to match other code sample implementations.


purplenicole730 (approved):

njooma:
@DTCurrie checking in -- can we merge?

DTCurrie:
> @DTCurrie checking in -- can we merge?

I was going to make another pass to clean up some of the streamable vs. not-streamable returns. So I’d prefer to keep it open, but if it’s annoying we can merge it, we just need to wait until after the deploy today so I have a chance to get the app side changes reviewed.

PR #287, opened 2024-09-16
RSDK-4735: Allow selective building and installing of unit tests and examples

lia-viam:
Lets the user control building of tests/examples. Some features worth remark:
- these examples are never exposed for subproject builds
- examples and tests are dependent, so it is not possible, for example, to specify building the examples but not the tests. 

These options could be decoupled but that is probably out of scope for this ticket

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -57,7 +57,6 @@ option(BUILD_SHARED_LIBS "If enabled, build shared libraries" ON)
 
 # Configure project-global options:
 
-
 # - `VIAMCPPSDK_ENFORCE_COMPILER_MINIMA`
 #
 # The user can elect to disable enforcement of compiler minima (at
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -55,8 +55,8 @@ project(viam-cpp-sdk
 #
 option(BUILD_SHARED_LIBS "If enabled, build shared libraries" ON)
 
-
 # Configure project-global options:
+
 
 # - `VIAMCPPSDK_ENFORCE_COMPILER_MINIMA`
 #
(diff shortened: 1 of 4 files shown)


CMakeLists.txt
```diff
@@ -127,6 +126,18 @@ option(VIAMCPPSDK_SANITIZED_BUILD "Build with address and UB sanitizers (less pe
...
+# The following options are only defined if this project is not being included as a subproject
+if (CMAKE_CURRENT_SOURCE_DIR STREQUAL CMAKE_SOURCE_DIR)
+  include(CMakeDependentOption)
+  option(VIAMCPPSDK_BUILD_EXAMPLES "Build the example executables (requires building tests too)" ON)
+
+  # Note that the complex module example requires the testing library and adds its tests to the unit
+  # test suite. Thus you can only disable tests if examples are also disabled.
+  # The call below says don't give the user the option to disable tests unless examples are already
+  # disabled, and default to building the tests in either case.
+  cmake_dependent_option(VIAMCPPSDK_BUILD_TESTS "Build the unit test suite" ON "NOT VIAMCPPSDK_BUILD_EXAMPLES" ON)
```
stuqdog, line 138: (q) This syntax looks unfamiliar to me, in particular the `"NOT VIAMCPPSDK_BUILD_EXAMPLES"` portion. I suspect I understand from context what this is meant to do but can you maybe explain briefly what `NOT` is doing here and why the variable name is in a string?

stuqdog (approved):
nice, lgtm!

CMakeLists.txt
lia-viam: Yeah this is more insane CMake syntax lol, here is the documentation:
https://cmake.org/cmake/help/latest/module/CMakeDependentOption.html
So CMakeDependentOption is like the `option` syntax but for situations where you don't always want the option to be exposed to the user. `"NOT VIAMCPPSDK_BUILD_EXAMPLES"` is the `<depends>` option in the docs above, it's logical test for exposing the option to the user. So `VIAMCPPSDK_BUILD_TESTS` will only be exposed to the user if `NOT VIAMCPPSDK_BUILD_EXAMPLES` is true, and as for why it's in quotes...classic travesty of everything in CMake being string-ly typed

stuqdog: ooh okay, that makes sense. thanks for clarifying!

acmorrow:
@lia-viam - I really wish CMake provided a better way to do this, but after a modest amount of research, I haven't found one. So, I think, especially given that this is already merged as a side effect of the conan work, it probably just makes sense to close this now.

PR #288, opened 2024-09-19
use boost program_options in example

abe-winter:
## What changed
- Use boost::program_options to pass in host + credentials in one of the examples
## Why
Incidental change while using this example. Not in love with this part of boost, p-ranav/argparse seems to be the modern equivalent, and cmake friendly? (But haven't tried it).

Not sure this is necessary / useful here, feel free to discard if not appropriate for sdk examples.


lia-viam:
Doing this had occurred to me too but I ended up deciding against it, kind of for the reasons you describe that people largely don't like using Boost program options, and the SDK examples are supposed to be usable as kind of cookie cutter copy-paste projects that someone would then build their own code on top of, in which case they probably would just use another argument-parsing library.

acmorrow:
There is prior art for using `boost::program_options` in the C++ SDK examples, so I'm pretty good with it. And I'd prefer that we used it rather than introducing a new dependency. It may not be a perfect library but it does just fine.

abe-winter:
I don't feel strongly about this and don't have context in this repo, one of you needs to make the decision unilaterally

acmorrow:
I think it is @lia-viam's call either way.

lia-viam:
Hahah let's leave this one aside then but thanks for the PR anyway!

lia-viam:
Oh wait just saw @acmorrow 's comment saying we already do this elsewhere! I think we can merge this in that case

lia-viam:
@abe-winter or anyone else, feel free to merge this when the build passes, I just merged `main` since it was blocking

abe-winter:
can you approve? repo is locked to sdk team I think

PR #289, opened 2024-09-19
Add frame_rate to cpp sdk

jckras:


src/viam/sdk/components/private/camera_server.cpp
```diff
@@ -114,6 +114,7 @@ ::grpc::Status CameraServer::GetProperties(
         *response->mutable_intrinsic_parameters() =
             Camera::to_proto(properties.intrinsic_parameters);
         response->set_supports_pcd(properties.supports_pcd);
+        response->set_frame_rate(properties.frame_rate);
```
jckras, line 117: is this the correct function "set_frame_rate()" because it comes from Property from camera and not getPropertiesResponse

src/viam/sdk/components/camera.hpp
```diff
@@ -65,6 +65,9 @@ class Camera : public Component {
 
         /// @brief Contains the mime types the camera supports.
         Camera::mime_types mime_types;
+
+        /// @brief Contains the camera's frame rate.
+        float frame_rate;
```
lia-viam, line 70: I think we probably want to use `double` here instead of `float`

src/viam/sdk/components/private/camera_server.cpp
lia-viam: Yep! In this method we are populating a `GetPropertiesResponse` using data from the camera's `Properties`

lia-viam (changes requested):

src/viam/sdk/components/camera.hpp
jckras: I used a float32 in the api changes so that is why I put in float here, but good with me! just for my own understanding why double instead of float here?

lia-viam: Actually sorry that's my mistake, thought this was a situation where a data format was using `float` to mean "this number has decimals" but we should translate it to `double` in C++, but I'm now seeing there are other components which do in fact use `float`

jckras: got it - in general is it better to use doubles instead of floats in C++?

lia-viam: You can think of it as the default "thing with a decimal" yeah, in some sense this is what the C++ language does too because a number with a decimal is deduced as double by default:
```cpp
template<class T>
void f(T t);
void g() { 
    f(5.0); // calls f<double>
}

auto x = 5.0; // x has type double
```
As for which is better, I think from a clarity/usability perspective it's safe to go with `double` by default, but if you're in an application or environment doing serious number crunching or with memory/bandwidth constraints you can decide if `float` might make more sense

src/viam/sdk/components/camera.cpp
```diff
@@ -254,7 +255,7 @@ viam::component::camera::v1::DistortionParameters Camera::to_proto(
     return proto;
 }
 
-Camera::Camera(std::string name) : Component(std::move(name)){};
+Camera::Camera(std::string name) : Component(std::move(name)) {};
```
lia-viam, line 258: This drives me crazy but sometimes vscode "format on save" will put a space here which then makes the linter complain, possibly due to mismatched versions of `clang-format`, you may just have to temporarily disable format on save or delete the space with another editor to get the build to pass

src/viam/sdk/components/camera.hpp
jckras: I see. thank you for explaining!

PR #290, opened 2024-09-20
RSDK-8343 - create public version_metadata file

stuqdog:


src/viam/sdk/tests/test_common.cpp
```diff
@@ -115,6 +116,16 @@ BOOST_AUTO_TEST_CASE(test_from_dm_from_extra) {
     BOOST_CHECK_EQUAL(from_dm_from_extra(map), false);
 }
 
+BOOST_AUTO_TEST_CASE(test_version_metadata) {
+    // we don't want to check the specific values because they're going to be changing,
```
lia-viam, line 120: How about we get the `sdk_version()` string here, and then check that it matches 
```
std::to_string(sdk_major_version()) + "." + std::to_string(sdk_minor_version()) + "." + std::to_string(sdk_patch_version());
```

src/viam/sdk/common/version_metadata.cpp
```diff
@@ -0,0 +1,41 @@
...
+
+#include <viam/sdk/common/private/version_metadata.hpp>
+
+namespace viam {
+namespace sdk {
+
+std::string sdk_version() {
+    std::string version_metadata(impl::k_version);
+    version_metadata.erase(0, version_metadata.find(';') + 1);
+    return version_metadata.substr(0, version_metadata.find(';'));
```
lia-viam, line 13: Maybe have the function body as
```cpp
static std::string result = []{
    std::string version_metadata(impl::k_version);
    version_metadata.erase(0, version_metadata.find(';') + 1);
    return version_metadata.substr(0, version_metadata.find(';'));
}();

return result;
```

That way we only ever do this computation once. 

This might seem like a micro optimization, but if this stuff later gets plugged into, eg, the response type of a grpc request that throws out the version, it could easily be hit over and over

src/viam/sdk/common/version_metadata.cpp
```diff
@@ -0,0 +1,41 @@
...
+}
+
+int get_sub_version(int which) {
+    auto version = sdk_version();
```
lia-viam, line 17: Similarly here, we could do something like

```cpp
static const std::array<int, 3> components =  [] {
    std::array<int, 3> result;
    
    // copy your function here but let the `for` loop parse the `int`s into the array

    return result;
}();
```

and then your function just returns `components[i]`;

lia-viam (changes requested):

src/viam/sdk/common/version_metadata.cpp
stuqdog: ooh good thinking! thanks :)

lia-viam (approved):

PR #296, opened 2024-09-24
RSDK-8826: Conan builds in CI 

lia-viam:
CI Job for building conan across the CI matrix.

Currently scheduled to run after hours on weekdays if there is a new commit (the `prepare` block) or with a manual trigger. Since this is kind of a heavy job we don't want it to run on every PR all the time.

This was largely copy-pasted from the Release job yaml


acmorrow (approved):
LGTM, but someone with more experience with GH workflows should also probably sign off, as I honestly don't have that much experience writing them.

stuqdog (approved):

PR #301, opened 2024-10-02
RSDK-8892: Remove ProtoType

lia-viam:
This PR finally does the migration for getting rid of ProtoType and AttributeMap in favor of ProtoValue et al.

Sorry in advance this diff is a monster. Generally speaking the changes in header files are just search and replace `const AttributeMap& extra` for `const ProtoStruct& extra`, any non-trivial stuff is for the most part happening in implementation files. 

There are a lot of simplifications happening in the unit test code and the examples, which were basically the only places where we are actually constructing and using `AttributeMap`. Even fewer actual uses of `ProtoType`, mostly in the example code.


src/viam/examples/modules/complex/gizmo/impl.cpp
```diff
@@ -15,20 +15,21 @@ using namespace viam::sdk;
...
-    const auto* const arg1_string = arg1->second->get<std::string>();
-    if (!arg1_string || arg1_string->empty()) {
-        std::ostringstream buffer;
-        buffer << gizmo_name << ": Required non-empty string parameter `arg1`"
-               << "` is either not a string or is an empty string";
-        throw std::invalid_argument(buffer.str());
+
+    const ProtoValue& arg1_val = arg1->second;
+    if (arg1_val.is_a<std::string>() && !arg1_val.get_unchecked<std::string>().empty()) {
+        return arg1_val.get_unchecked<std::string>();
```
acmorrow, line 27: Is aw this patern in `base/impl.cpp` above too, and sort of shrugged it off, but, it now that I see it twice it feels a little chatty. Is `ProtoValue` missing some affordance for this sort of pattern? Would `std::optional` (or a polyfill for it) simplify things? Should `ProtoValue` have some monadic APIs (and_then, transform, or_else), etc?

This, btw, is what motivated my slack ping about polyfills / C++17 since I feel like `optional` probably plays a role in such an API.

src/viam/examples/motor/example_motor.cpp
```diff
@@ -109,9 +109,7 @@ int main() {
 
         // Explicitly stop the motor
         cout << "Stopping motor" << endl;
-        vs::AttributeMap stop_map =
-            std::make_shared<std::unordered_map<std::string, std::shared_ptr<vs::ProtoType>>>();
-        motor->stop(stop_map);
+        motor->stop({});
```
acmorrow, line 112: Nice

src/viam/sdk/common/proto_value.hpp
```diff
@@ -328,17 +328,17 @@ void to_proto(const ProtoList& vec, google::protobuf::Value* v);
...
 void map_to_struct(const ProtoStruct& m, google::protobuf::Struct* s);
 
 /// @brief Convert a type to a google::protobuf::Value.
 /// @note This method is trivially templated to insulate google::protobuf::Value from our
 /// API/ABI. It is meant to be called with no template parameters in a translation unit which
 /// includes <google/protobuf/struct.pb.h>
-template <typename T, typename Value = google::protobuf::Value>
-Value to_proto(T&& t) {
+template <typename Value = google::protobuf::Value>
+Value to_proto(const ProtoValue& pv) {
```
acmorrow, line 339: For me, `pv` reads like `pointer to v`. Maybe just call it `proto_value`.

src/viam/sdk/common/proto_value.hpp
```diff
@@ -328,17 +328,17 @@ void to_proto(const ProtoList& vec, google::protobuf::Value* v);
 void to_proto(const ProtoStruct& m, google::protobuf::Value* v);
 void to_proto(const ProtoValue& t, google::protobuf::Value* v);
 
-ProtoStruct struct_to_map(google::protobuf::Struct const* s);
+void struct_to_map(google::protobuf::Struct const* s, ProtoStruct& m);
```
acmorrow, line 331: What motivated changing this API?

acmorrow (approved):
This looks good to me.

My only thoughts are:

- It looks like the access patterns are a little chatty and I'd like to have an idea how that can be improved. It doesn't need to be in this review, could be a ticket.
- Was unclear why the signature of `struct_to_map` (needed to be?) changed.

Just let me know your thoughts on those two items and unless there are subtantive changes that result from that (I doubt it) please merge this!

acmorrow:
Oh, one more thought: how do you plan to communicate this change to module authors or others who are likely to be affected and need to make changes in their downstream code?

src/viam/sdk/common/proto_value.hpp
lia-viam: The old way was erroring out on compilation with an infinite loop of template insantiations since we had both

```
ProtoStruct struct_to_map(protobuf::Struct const* s);

template<typename Struct = protobuf::Struct>
ProtoStruct struct_to_map(const Struct& s);
```
absent any constraints, such as `enable_if_t<!std::is_pointer<std::remove_reference_t<T>>{}>`, I believe they ended up calling each other over and over.

This fixed that ambiguity without `enable_if`, and it sort of matches the implied semantic convention that the overloads taking pointer are not really for normal use

src/viam/examples/modules/complex/gizmo/impl.cpp
lia-viam: Yeah it was bugging me as well that this kept coming up, being that it's the one scenario where we end up more verbose than the previous code. 

I personally am a big fan of optionals and monadic operations but in this case we might be better off with a pointer API. For example Boost.JSON has [these](https://www.boost.org/doc/libs/1_85_0/libs/json/doc/html/json/ref/boost__json__value/if_double.html), and we could spell them as `try_get<T>` or similar, since `if_bool` strikes me as a bit idiosyncratic 

In any case definitely in agreement that this would be nice to simplify, will create a ticket.

PR #303, opened 2024-10-04
Use proto value get in example code

lia-viam:
In #301 I overlooked the existence of the `get<T>` API which resulted in some really awkwardly phrased inspection code, see https://github.com/viamrobotics/viam-cpp-sdk/pull/301#discussion_r1788130096

Should have read my own code/docs! This PR simplifies the accessors in the example code to look mostly like they did before, now with slightly fewer pointer derefs.


acmorrow (approved):
LGTM!

PR #306, opened 2024-10-15
RSDK-8822: Whole archive linking

lia-viam:
This PR fixes static builds and example build/running for the SDK, following the observation in RSDK-8822 that the API needs to be linked to the SDK with whole archive linking.

This is in some sense a "good enough" solution because we are capitulating to the kind of annoying way that CMake handles this where whole archive linking is specified by the SDK rather than being a property of the API. That said, another way of looking at it is we're responsible for the CMake install and the Conan install, and Conan still (with some effort) models the correct behavior, ie, that whole archive is part of the link interface of the API. So in the interest of being able to say that static builds and examples work this is probably the best solution for now.

There are changes made to the conan test package and the examples too, so that now we are verifying the proper operation of an example module linked against the SDK. 

The one possibly controversial thing I would submit is the use of `--push-state` and `--pop-state` to do whole archive linking. This is more correct than doing `--whole-archive` and `--no-whole-archive` in the sense that it would not clobber a prior setting of whole archive linking, but somewhat less general in that it does not work on linkers with no push/pop state. See here for the full logic that CMake uses:
https://github.com/Kitware/CMake/blob/ddf1d2944fe53b0fb0be79621c53d2d235fce07b/Modules/Platform/Linker/GNU.cmake#L35

Welcome feedback from anyone (possibly cc @njooma) on platforms we want to support vis-a-vis whether or not push/pop state will be a thing--I believe this landed in the gcc that ships in Ubuntu 16.04, and I'll have to check for lld as well 

I suspect that with the platforms we support we can assume that the linker has these flags, but maybe CI and/or user feedback will suggest otherwise


stuqdog (approved):
lgtm though I confess lacking some context, particularly on the push/pop state question.

lia-viam:
Removing the draft status here--going to do Conan/Release matrix CI support in another PR

stuqdog (approved):

acmorrow (approved):
This seems fine. Have you tested whether you are able to make a static build of the tflite module that works on macOS with this approach?

lia-viam:
@acmorrow happy to report I've been able to do a static tflite build on macOS and a bookworm container as well!

PR #309, opened 2024-10-22
Fix conan shared builds and conan CI

lia-viam:
It turns out shared library builds have been broken all along with conan and I did not realize because the test_package was not meaningfully testing the library, but then I further did not realize because the cron + new commits dispatch for the conan job was broken. This PR
- fixes shared builds
- changes the CI trigger for the conan job


.github/workflows/conan.yml
```diff
@@ -18,17 +18,6 @@ jobs:
...
-        id: new_commit_count
-        shell: bash
-        run: echo "commit_count=$(git log --oneline --since '24 hours ago' | wc -l)" >> $GITHUB_OUTPUT
-
-      - name: Cancelling scheduled build with no new commits
-        uses: andymckay/cancel-action@0.2
-        if: |
-          steps.new_commit_count.outputs.commit_count == '0' &&
-          github.event_name == 'schedule'
-
```
acmorrow, line 31: This seems unrelated to the linking changes

conanfile.py
```diff
@@ -89,7 +84,7 @@ def package_info(self):
         for component in ["viamsdk", "viamapi"]:
            self.cpp_info.components[component].set_property("cmake_target_name", "viam-cpp-sdk::{}".format(component))
            self.cpp_info.components[component].set_property("pkg_config_name", "viam-cpp-sdk-lib{}".format(component))
-           self.cpp_info.components[component].requires = ["grpc::grpc++", "protobuf::libprotobuf"]
+           self.cpp_info.components[component].requires = ["grpc::grpc++", "grpc::grpc++_reflection", "protobuf::libprotobuf"]
```
acmorrow, line 87: It feels a little wrong to me that we need to reiterate library level dependencies known to cmake up here in the conan stuff. I'd expect that it would operate more at the package level and let CMake handle the library details. Can you elaborate on what's happening here and why?

conanfile.py
```diff
@@ -117,8 +112,7 @@ def package_info(self):
             "xtensor::xtensor",
 
             "viam_rust_utils",
-            "abseil::absl_strings",
-            "grpc::grpc++_reflection"
+            "abseil::absl_strings"
```
acmorrow, line 115: Similar feelings on this list. Why are the internal linkage dependencies of `viamsdk` getting reiterated here?

acmorrow (dismissed):

conanfile.py
```diff
@@ -32,25 +32,20 @@ def set_version(self):
...
         # The SDK supports older grpc and protobuf, but these are the oldest
         # maintained conan packages.
-        self.requires('grpc/[>=1.48.4]', transitive_headers=True)
-        self.requires('protobuf/[>=3.17.1]', transitive_headers=True)
+        self.requires('grpc/[>=1.48.4]', transitive_headers=True, transitive_libs=True)
+        self.requires('protobuf/[>=3.17.1]', transitive_headers=True, transitive_libs=True)
 
         self.requires('xtensor/[>=0.24.3]', transitive_headers=True)
-        self.requires('abseil/[>=20230125.3]')
+        self.requires('abseil/[>=20230125.3]', transitive_libs=True)
```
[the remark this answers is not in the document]
lia-viam: Hmm in fact I might be able to remove it from `xtensor` because it is a `header-library`, see https://docs.conan.io/2/reference/conanfile/methods/requirements.html#package-type-trait-inferring

Let me see if `abseil` needs to be there too or not, I added it because comparing the conan-generated build files to the ones generated from our CMake install, ours were pulling in all abseil into the link line

jckras (commented):
Just trying to gather context for my own understanding:

- How are shared libraries being used in cpp-sdk? does the shared-ness refer to whether libraries are being compiled as shared or static?
- what do you mean by the "shared-ness of grpc and protobuf needs to match our shared-ness"... the way we compile gRPC and protobuf libraries must be the same as the way we compile our project?

lia-viam:
@jckras 

>     * How are shared libraries being used in cpp-sdk? does the shared-ness refer to whether libraries are being compiled as shared or static?

Exactly!
 
>     * what do you mean by the "shared-ness of grpc and protobuf needs to match our shared-ness"... the way we compile gRPC and protobuf libraries must be the same as the way we compile our project?

That part is a bit vague and Drew suggested elaborating on it as well but yes, we need a shared grpc/proto for a shared SDK build and vice versa, going to update that comment and also elaborate on it

conanfile.py
```diff
@@ -32,25 +32,20 @@ def set_version(self):
...
-        # stuff will probably break.
-        # If you want your shared build to also build the world as shared, you
-        # can invoke conan with -o "&:shared=False" -o "*:shared=False",
-        # possibly with --build=missing or --build=cascade as desired,
-        # but this is probably not necessary.
-        if not self.options.shared:
-            self.options["*"].shared = False
+        # The shared-ness of grpc and protobuf needs to match our shared-ness or stuff will break
+        self.options["grpc"].shared = self.options.shared
+        self.options["protobuf"].shared = self.options.shared
```
jckras, line 37: are grpc and protobuf the only libraries we care about? and if so how do we know that (just for my own understanding)

.github/workflows/conan.yml
```diff
@@ -18,17 +18,6 @@ jobs:
       - name: Checkout Code
         uses: actions/checkout@v4
 
-      - name: Get new commit count
-        id: new_commit_count
```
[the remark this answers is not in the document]
lia-viam: I think it's just a bit redundant/unnecessary, I would like to get the new commit count thing working correctly but I think the correct logic might actually be something like "did the last build fail, or have there been any new commits since the last successful build" which escapes me a bit as to how to express that in github actions

conanfile.py
lia-viam: Yeah so beforehand we were taking kind of a firehose approach of forcing the shared-ness of every single dependency which I think was not necessary. I'll elaborate on this in the doc comment but I think the thing that can potentially break is related to static/global data variable initializers being run by API which is calling into stuff from grpc, but maybe @acmorrow can confirm

conanfile.py
acmorrow: Thanks for the links and context, it makes more sense to me now.

conanfile.py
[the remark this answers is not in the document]
jckras: just for my own understanding, why do you set some properties like `transitive_libs` or  `transitive_headers` for some libraries and not others? and how do you know which ones to set?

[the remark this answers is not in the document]
lia-viam: That actually took me a while to figure out and I had to ask on github, but there's a bit more in the page I linked above outside of the anchor point: https://docs.conan.io/2/reference/conanfile/methods/requirements.html

A `requires` in and of itself just means that you need this thing to build our project, and the `transitive_meow` properties then determine whether those requirements are visible downstream. With `boost` for example the SDK uses Boost, but moreover there are places where Boost is part of our public API and appears in public header files, which means that a consumer of the SDK must also have the include paths for the Boost libraries to work properly.

Transitive libs is similar but for linking libraries instead of include paths to headers, I'm still trying to get my head around it 100% though

.github/workflows/conan.yml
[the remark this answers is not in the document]
jckras: not sure how helpful this is but I found this online: https://github.com/marketplace/actions/last-successful-commit 

I am not super familiar with github actions but maybe you could use this to get the SHA of the last successful commit, then get the SHA of the latest commit in the branch and compare them. If they match then there have been no new commits since the last successful build, otherwise there have been new commits?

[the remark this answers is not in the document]
lia-viam: I will give that a shot!

conanfile.py
jckras: so are you saying there are potential issue when the API calls functions from gRPC, and those gRPC functions rely on static or global variables? and previously shared-ness was being forced on everything so this wasn't a problem?

test_package/conanfile.py
```diff
@@ -41,7 +40,5 @@ def test(self):
                 out = proc.communicate()[0]
                 pass
 
-            ConanOutput("test package").writeln(out)
```
jckras, line 44: if i understand correctly, this just prints the output from the test command `subprocess.Popen(cmd, shell=True, stdout=subprocess.PIPE, text=True)`, why do no longer want to see the output?

.github/workflows/conan.yml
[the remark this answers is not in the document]
lia-viam: @jckras ended up giving it a shot with `last-successful-commit`, I _think_ this is how it should work but probably won't really know until a scheduled build tries to run

lia-viam:
Just sharing this for reference/discussion:
https://github.com/conan-io/conan-center-index/issues/23421

njooma (approved):

PR #310, opened 2024-10-25
RSDK-9077: Docker containers for module development

lia-viam:
This implements a minimal proof of concept version of docker containers for module development. The old images have been moved to a `base-images` directory, and now the `Dockerfile.sdk-build` can be used off an existing build of a base image. 

This is validated by the changes to the `test.yml` workflow, removing the duplicated dockerfile we used to use.


.github/workflows/test.yml
```diff
@@ -22,5 +22,15 @@ jobs:
       ###########################################
       - name: build-docker-test
         run: |
-          docker build -t cpp . -f etc/docker/tests/Dockerfile.debian.bullseye
-          docker run -e BUILD_SHARED=${{ matrix.BUILD_SHARED }} --rm -i -w /tmp cpp /bin/bash
+          docker build -t cpp . -f etc/docker/base-images/Dockerfile.debian.bullseye
```
jckras, line 25: is this creating a base image if you don't already have one? otherwise what is this `docker build `doing compared to the one below it/why do we need two?

etc/docker/Dockerfile.sdk-build
```diff
@@ -0,0 +1,47 @@
+# This Dockerfile is meant to be built on top of an existing image from the base-images directory.
+# Suggested usage, from the SDK root:
```
jckras, line 2: I'm not sure if I understand correctly but "Suggested commands to build and run from the SDK root" might be more clear here?

etc/docker/Dockerfile.sdk-build
```diff
@@ -0,0 +1,47 @@
...
+
+RUN mkdir build && \
+    cd build && \
+    cmake .. -G Ninja \
+    -DCMAKE_BUILD_TYPE=${BUILD_TYPE} \
+    -DBUILD_SHARED_LIBS=${BUILD_SHARED} \
+    -DVIAMCPPSDK_OFFLINE_PROTO_GENERATION=ON \
+    -DVIAMCPPSDK_BUILD_TESTS=${BUILD_TESTS} \
+    -DVIAMCPPSDK_BUILD_EXAMPLES=${BUILD_EXAMPLES} \
+    ${EXTRA_CMAKE_ARGS}
```
jckras, line 43: why are we setting this build flags here instead of the run.sh file like before?

etc/docker/tests/run_test.sh
```diff
@@ -1,17 +1,8 @@
...
-cmake .. -G Ninja -DVIAMCPPSDK_USE_DYNAMIC_PROTOS=ON \
-		-DVIAMCPPSDK_OFFLINE_PROTO_GENERATION=ON \
-		-DVIAMCPPSDK_SANITIZED_BUILD=$BUILD_SHARED \
-		-DVIAMCPPSDK_CLANG_TIDY=ON \
-		-DBUILD_SHARED_LIBS=$BUILD_SHARED
-
-ninja all
-ninja install
 INSTALL_DIR="$(pwd)/install"
+BUILD_SHARED=$(grep 'BUILD_SHARED_LIBS' CMakeCache.txt | cut -d '=' -f 2)
```
jckras, line 5: are we adding this line because of the other shared library changes in #309 that are happening?

jckras (approved):
I am still learning about docker and development environments so I left a few questions for my own understanding. Otherwise LGTM

.github/workflows/test.yml
lia-viam: yeah this is creating the base image which the step below then uses to build the SDK. The idea is to have a modular, many-to-one type scenario where we can have a bunch of base images (the ones in `base-images`) which, if built, can then be used as input to a single "build the sdk on top of a base" image. from what I found researching this requires having stuff in separate Dockerfiles which are in turn built with separate invocations of `docker build`

etc/docker/Dockerfile.sdk-build
lia-viam: so it used to be that CI ran tests by using a separate Dockerfile (the deleted one) which copy-pasted everything from the `Dockerfile.debian.bullseye` but with the addition of an `ENTRYPOINT` to `run_test.sh`, but really what that script was doing was like `build_the_sdk_and_run_test`. So with this PR we have a Dockerfile which now takes care of the `build_the_sdk` part, and it can use the existing `Dockerfile.debian.bullseye` without needing the duplicated one

etc/docker/tests/run_test.sh
lia-viam: no not super related except to the extent that this is part of the ongoing saga of having shared/static builds properly supported :) but in particular this script is building with CMake rather than Conan so it already works properly

njooma (approved):
I think this is a great starting point. Eventually, we might want to release our base tags online so that the end user doesn't have to build it themselves ahead of time.

PR #312, opened 2024-10-28
RSDK-5986 - module logging propagation and resource-level loggers

stuqdog:
Before and after (server side):
Screenshot 2024-10-28 at 13 47 28
Screenshot 2024-10-28 at 13 47 54

Before and after (client side):
Screenshot 2024-10-28 at 13 48 16
Screenshot 2024-10-28 at 13 48 27


stuqdog:
@acmorrow fyi, this is still marked as draft because the abort trap issue isn't fully resolved, but I do think it's ready for you to look at whenever you're able. I've managed to significantly reduce (though not eliminate) the occurrence of abort traps, and isolate the failure points when we _do_ get an abort trap. These places are highlighted in comments in the code. Let me know if you want to discuss or have any questions or thoughts!

src/viam/sdk/rpc/dial.cpp
```diff
@@ -35,10 +35,16 @@ void ViamChannel::close() {
         return;
     }
     closed_ = true;
+    // CR erodkin: every now and again this call causes a failure. From adding log statements
+    // in rust-utils, I have confirmed that the failure happens within the rust code.
```
stuqdog, line 39: Flagging this, for some reason `free_string` very occasionally fails in the rust call, specifically [here](https://github.com/viamrobotics/rust-utils/blob/6a7f896403d99d2f26e6464a3d86fa42eeb4615e/src/ffi/dial_ffi.rs#L285). Because the rust code is so minimal I haven't tried much on that side beyond explicitly dropping the `CString`, which didn't seem to make a difference.

src/viam/sdk/common/utils.hpp
```diff
@@ -84,16 +90,6 @@ ProtoStruct with_debug_entry(ProtoStruct&& map, std::string debug_key);
...
 
-/// @brief Set the boost trivial logger's severity depending on args.
-/// @param argc The number of args.
-/// @param argv The commandline arguments to parse.
-///
-/// Sets the boost trivial logger's severity to debug if "--log-level=debug" is the
-/// the third argument. Sets severity to info otherwise. Useful for module
-/// implementations. See LogLevel documentation in the RDK for more information on
-/// how to start modules with a "log-level" commandline argument.
-void set_logger_severity_from_args(int argc, char** argv);
```
stuqdog, line 95: This is no longer necessary because we support per-logger severity levels. Additionally, it was causing us to fail to send `debug` logs to RDK.

lia-viam:
Closing in favor of #383

PR #322, opened 2024-11-13
build docker base images in CI

abe-winter:
## What changed
- add CI yaml for uploading the docker base images to ghcr
## Why
Allow people to use these without having to build them on-laptop
## Note
Github CI requires a workflow to be on the main branch before it can be tested; this hasn't been tested and may require another PR. (Once the workflow has ever been on the main branch, it can then be run from other branches).

### DIFF
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -24,5 +24,6 @@ jobs:
       with:
         context: .
         push: true
-        tags: ${{ inputs.image }}:${{ inputs.tag }}
+        tags: "${{ inputs.image }}:${{ inputs.tag }}"
+        file: etc/docker/base-images/${{ inputs.dockerfile }}
         labels: ${{ steps.meta.output.labels }}
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -1,12 +1,25 @@
 on:
   workflow_dispatch:
     inputs:
-      image:
-        default: ghcr.io/viamrobotics/cpp-base
+      image-prefix:
+        description: "gets suffixed with 'base' and 'sdk' to create actual image name"
+        default: ghcr.io/viamrobotics/cpp-
       dockerfile:
         default: Dockerfile.debian.bullseye
       tag:
         default: bullseye-amd64
+      build-base:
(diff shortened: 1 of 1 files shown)


.github/workflows/docker.yml
```diff
@@ -0,0 +1,61 @@
...
+      build-base:
+        description: "whether to build the base image. the base images change less often and may not be necessary to rebuild."
+        type: boolean
+        default: false
+      build-sdk:
+        description: "whether to build the SDK image. if this is true and no corresponding base image exists, the job will fail."
+        type: boolean
+        default: false
+      push:
+        description: "whether to push the images after building them"
```
jckras, line 20: just for my own understanding, In what cases would we run this workflow without pushing images to GHCR? why do we have the option to disable pushing `inputs.push` here?

.github/workflows/docker.yml
```diff
@@ -0,0 +1,61 @@
...
+        images: ${{ inputs.image-prefix }}base
+    - uses: docker/build-push-action@v5
+      if: inputs.build-base
+      with:
+        push: ${{ inputs.push }}
+        tags: "${{ inputs.image-prefix }}base:${{ inputs.tag }}"
+        file: etc/docker/base-images/${{ inputs.dockerfile }}
+        labels: ${{ steps.base-meta.output.labels }}
+
+    # build sdk (if inputs.build-sdk)
```
jckras, line 47: Can you explain the purpose of separating `build-base` and `build-sdk` into two separate steps? are there times where only one would be built?

jckras (approved):
LGTM. Left a few questions for my own understanding

.github/workflows/docker.yml
abe-winter: I did this because the instructions in the readme were already two-step

but in general the purpose of 'base images' is to save time on the constant part of the build

here specifically, could imagine us rebuilding the base images weekly or monthly, and rebuilding the sdk image on every release or on main branch update

.github/workflows/docker.yml
abe-winter: push=false for testing only; for commands with a big side effect (in this case overwriting a public image), I like to have a 'dry run' option

stuqdog (approved):

PR #323, opened 2024-11-14
RSDK-3589 add wrapper for navigation service

abe-winter:
## What changed
- add abstract Navigation class, navigation server and client, register them
- incidental: utils for repeatedPtrToVec and vecToRepeatedPtr
## Why
So people can write nav modules in C++.
## Follow-ups
- someone with deeper understanding of the interface should rewrite docstrings in navigation.hpp
- maybe add example
- refactor `repeatedPtrToVec` and friends

### DIFF
--- a/src/viam/sdk/common/private/proto_utils.hpp
+++ b/src/viam/sdk/common/private/proto_utils.hpp
@@ -6,7 +6,9 @@
 
 #include <viam/api/common/v1/common.pb.h>
 
-namespace viam::sdk::impl {
+namespace viam {
+namespace sdk {
+namespace impl {
 
 /// @brief Copies elements from a protobuf repeated pointer array into a std::vector. Src type
 /// must have a `to_proto` method.
@@ -56,4 +58,6 @@ void repeatedPtrToVec(const google::protobuf::RepeatedPtrField<Src>& src,
     }
 }
 
-}  // namespace viam::sdk::impl
+}  // namespace impl
+}  // namespace sdk
+}  // namespace viam
--- a/src/viam/sdk/common/private/proto_utils.hpp
+++ b/src/viam/sdk/common/private/proto_utils.hpp
@@ -6,8 +6,7 @@
 
 #include <viam/api/common/v1/common.pb.h>
 
-namespace viam {
-namespace sdk {
+namespace viam::sdk::impl {
 
 /// @brief Copies elements from a protobuf repeated pointer array into a std::vector. Src type
 /// must have a `to_proto` method.
@@ -57,5 +56,4 @@ void repeatedPtrToVec(const google::protobuf::RepeatedPtrField<Src>& src,
     }
 }
 
-}  // namespace sdk
-}  // namespace viam
+}  // namespace viam::sdk::impl
--- a/src/viam/sdk/common/proto_utils.hpp
+++ b/src/viam/sdk/common/proto_utils.hpp
@@ -10,7 +10,7 @@ namespace viam {
 namespace sdk {
 
 /// @brief Copies elements from a protobuf repeated pointer array into a std::vector. Src type
-/// must be implicitly convertible to Dst (probably via operator on Src).
+/// must have a `to_proto` method.
 template <typename Src, typename Dst>
 void vecToRepeatedPtr(const std::vector<Src>& vec, google::protobuf::RepeatedPtrField<Dst>& dest) {
     dest.Clear();
@@ -21,7 +21,7 @@ void vecToRepeatedPtr(const std::vector<Src>& vec, google::protobuf::RepeatedPtr
 }
 
 /// @brief Non-member to_proto() version. (necessary for moving generated types out of wrapper
-/// headers). Template param F is a function that converts from Src to Dst.
+/// headers). Takes explicit `to_proto`.
 template <typename Src, typename Dst>
 void vecToRepeatedPtr(const std::vector<Src>& vec,
                       google::protobuf::RepeatedPtrField<Dst>& dest,
@@ -33,8 +33,8 @@ void vecToRepeatedPtr(const std::vector<Src>& vec,
     }
 }
 
-/// @brief Copies elements from a std::vector into a protobuf repeated pointer array. Src type
-/// must be implicitly convertible to Dst (probably via constructor on Dst).
+/// @brief Copies elements from a std::vector into a protobuf repeated pointer array. Dst type
+/// must have a `from_proto` static method.
 template <typename Src, typename Dst>
 void repeatedPtrToVec(const google::protobuf::RepeatedPtrField<Src>& src, std::vector<Dst>& vec) {
     vec.clear();
@@ -44,8 +44,8 @@ void repeatedPtrToVec(const google::protobuf::RepeatedPtrField<Src>& src, std::v
     }
 }
 
-/// @brief Non-member to_proto() version. (necessary for moving generated types out of wrapper
-/// headers). Template param F is a function that converts from Src to Dst.
+/// @brief Non-member from_proto() version. (necessary for moving generated types out of wrapper
+/// headers). Takes explicit `from_proto`.
 template <typename Src, typename Dst>
 void repeatedPtrToVec(const google::protobuf::RepeatedPtrField<Src>& src,
                       std::vector<Dst>& vec,
--- a/src/viam/sdk/common/proto_utils.hpp
+++ b/src/viam/sdk/common/proto_utils.hpp
@@ -1,0 +1,53 @@
+/// @file common/proto_utils.hpp
+///
+/// @brief Utils that require generated proto includes. These should be #included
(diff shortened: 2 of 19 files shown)


src/viam/sdk/services/navigation.hpp
```diff
@@ -0,0 +1,101 @@
...
+            return ret;
+        }
+
+        std::string destination_waypoint_id;
+        std::vector<geo_point> geopoints;
+    };
+
+    API api() const override;
+
+    virtual Mode get_mode(const std::string name, const ProtoStruct& extra) = 0;
```
lia-viam, line 73: this header looks good overall, maybe you were already going to do this but we'll want to have the overrides which don't take the `extra` parameter similar to how it looks in the motion service for example

src/viam/sdk/services/navigation.hpp
```diff
@@ -0,0 +1,101 @@
...
+        k_gps,
+    };
+
+    struct LocationResponse {
+        geo_point location;
+        double compass_heading;
+    };
+
+    struct Waypoint {
+        Waypoint(const viam::service::navigation::v1::Waypoint& proto)
```
lia-viam, line 40: similar to the other comment about keeping API stuff out of headers, @stuqdog is currently working on a PR to take these kinds of conversion functions out of public headers, so maybe you guys can talk so that we have a similar approach in this PR as well

lia-viam (commented):
overall this looks really solid, just added some suggestions about SDK coding standards changes which are currently in progress!

src/viam/sdk/services/navigation.hpp
lia-viam: but one option for example is, since I think these are only being used in `navigation_client.cpp` to just inline them there

src/viam/sdk/common/utils.cpp
```diff
@@ -184,5 +184,21 @@ bool from_dm_from_extra(const ProtoStruct& extra) {
     return false;
 }
 
+template <typename Src, typename Dst>
+void vecToRepeatedPtr(const std::vector<Src>& vec, google::protobuf::RepeatedPtrField<Dst>& dest) {
```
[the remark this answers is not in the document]
abe-winter: good call -- clearing the destination in d3345726 to protect against adding to non-empty collection

I think `Assign(begin, end)` doesn't work because the types are different -- tried to get it working with constructor/operator implicit conversion approach, couldn't, guessing even harder now that the constructor/operators have been moved out of common headers

src/viam/sdk/services/navigation.hpp
abe-winter: added overrides without `extra` in 3f7c6c7f

src/viam/sdk/services/navigation.hpp
abe-winter: okay I removed all *.pb.h includes from the wrapper header (navigation.hpp)

how:
- I moved the implicit conversions for `Waypoint` and `Path` to `auto_from_proto` / `auto_to_proto` template specializations in a cpp file
- then did the same for the std::vector <-> RepeatedPointer conversions

now the invocation looks like:
```c++
// when destination type has a `to_proto` method, use it:
repeatedPtrToVec(response.obstacles(), *ret);

// otherwise, pass in the conversion function:
repeatedPtrToVec(response.waypoints(), *ret, auto_to_proto);
```

can do this more cleanly w/ c++20 concepts I think? (because the template system can detect the presence of the to_proto method).

may have overthought this

src/viam/sdk/common/proto_utils.hpp
```diff
@@ -0,0 +1,61 @@
...
+    dest.Clear();
+    dest.Reserve(vec.size());
+    for (auto& x : vec) {
+        *dest.Add() = x.to_proto();
+    }
+}
+
+/// @brief Non-member to_proto() version. (necessary for moving generated types out of wrapper
+/// headers). Template param F is a function that converts from Src to Dst.
+template <typename Src, typename Dst>
```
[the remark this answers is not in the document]
abe-winter: explain this one -- this is better because it would select this specialization and return a static assert error, rather than giving a long 'no matching template' error?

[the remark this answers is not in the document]
acmorrow: I think, as written, yours will only accept function pointers. I could be wrong, I'd need to test it out on godbolt.

src/viam/sdk/common/private/proto_utils.hpp
```diff
@@ -6,8 +6,7 @@
 
 #include <viam/api/common/v1/common.pb.h>
 
-namespace viam {
-namespace sdk {
+namespace viam::sdk::impl {
```
acmorrow, line 9: Oops, sorry, I misled you with my comment. This syntax is c++17 only. You still need to do it with multiple lines. I was just trying to be clear that the `impl` namespace should be the innermost:

```
namespace viam {
namespace sdk {
namespace impl {
```

abe-winter: oops -- switched back in 3def54d5

src/viam/sdk/common/proto_utils.hpp
```diff
@@ -0,0 +1,61 @@
...
+
+#include <viam/api/common/v1/common.pb.h>
+
+namespace viam {
+namespace sdk {
+
+/// @brief Copies elements from a protobuf repeated pointer array into a std::vector. Src type
+/// must be implicitly convertible to Dst (probably via operator on Src).
+template <typename Src, typename Dst>
+void vecToRepeatedPtr(const std::vector<Src>& vec, google::protobuf::RepeatedPtrField<Dst>& dest) {
```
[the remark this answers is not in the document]
lia-viam: So generally speaking I'm very much against output parameters being passed as pointers, it's one of the things I dislike the most about google's C++ coding guidelines. I think with your counterexample above the point is that, in C++ and as high level as this code is, we generally aren't passing around pointer types in the first place, and the point of references is that they must point to a valid object or else it's UB. So as for

> oops, what if pdest is nullptr

That's on the caller for dereferencing a null pointer in the first place, and is equally true of if `my_vec` was `*p_my_vec` for some reason, and this is kind of the point of C++ having reference types. 

Callsite legibility I'm somewhat sympathetic to but would be better done with something like `std::ref(arg)` or `gsl::output_parameter`, because then you don't lose the non-null guarantee. 

.....All that said, I think this could be a pointer parameter just for ABI insulation reasons

src/viam/sdk/services/private/navigation_client.cpp
```diff
@@ -0,0 +1,128 @@
...
+#include <viam/sdk/common/utils.hpp>
+#include <viam/sdk/services/navigation.hpp>
+
+namespace viam {
+namespace sdk {
+namespace impl {
+
+using namespace viam::service::navigation::v1;
+
+Navigation::Waypoint waypoint_from_proto(const viam::service::navigation::v1::Waypoint& proto) {
```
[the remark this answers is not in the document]
abe-winter: renamed to bare from/to proto in 4e9f32c

src/viam/sdk/tests/test_navigation.cpp
```diff
@@ -0,0 +1,136 @@
...
+        auto loc = client.get_location();
+        BOOST_CHECK_EQUAL(loc, Navigation::LocationResponse{});
+        mock->loc.compass_heading = 1;
+        loc = client.get_location();
+        BOOST_CHECK_CLOSE(loc.compass_heading, 1., 1e-4);
+    });
+}
+
+// weak backport of std::transform.
+template <typename T, typename U>
```
lia-viam, line 45: I think just template this as `template <typename T, typename U, typename Func>` and have the second argument be `Func f`

abe-winter: that made it sad

could specify template param I guess? but can I leave as-is?

> /host/src/viam/sdk/tests/test_navigation.cpp: In lambda function:
> /host/src/viam/sdk/tests/test_navigation.cpp:78:14: error: no matching function for call to 'mapOver(std::vector<viam::sdk::Navigation::Waypoint>&, std::__cxx11::basic_string<char> (*)(const viam::sdk::Navigation::Waypoint&))'
>    78 |             });
>       |              ^
> /host/src/viam/sdk/tests/test_navigation.cpp:46:16: note: candidate: 'template<class T, class U, class Func> std::vector<ValueType> viam::sdktests::navigation::mapOver(const std::vector<U>&, Func)'
>    46 | std::vector<T> mapOver(const std::vector<U>& source, Func f) {
>       |                ^~~~~~~
> /host/src/viam/sdk/tests/test_navigation.cpp:46:16: note:   template argument deduction/substitution failed:
> /host/src/viam/sdk/tests/test_navigation.cpp:78:14: note:   couldn't deduce template parameter 'T'
>    78 |             });

lia-viam (approved):

PR #324, opened 2024-11-14
Add MoveThroughJointPositions to Arm component

raybjork:
I took a first pass at adding this new function to the sdk but don't completely understand everything that is required in doing this, so I'm hoping this can be a launching point to learn more about coding practices in this environment

### DIFF
--- a/src/viam/sdk/components/arm.hpp
+++ b/src/viam/sdk/components/arm.hpp
@@ -103,6 +103,20 @@ class Arm : public Component, public Stoppable {
     virtual void move_to_joint_positions(const std::vector<double>& positions,
                                          const ProtoStruct& extra) = 0;
 
+    /// @brief Move each joint on the arm through the positions specified in @param positions
+    /// @param options optional specifications to be obeyed during the motion.
+    inline void move_through_joint_positions(const std::vector<std::vector<double>>& positions, 
+                                             const component::arm::v1::MoveOptions& options) {
+        return move_through_joint_positions(positions, options, {});
+    }
+
+    /// @brief Move each joint on the arm through the positions specified in @param positions
+    /// @param options optional specifications to be obeyed during the motion.
+    /// @param extra Any additional arguments to the method.
+    virtual void move_through_joint_positions(const std::vector<std::vector<double>>& positions,
+                                              const component::arm::v1::MoveOptions& options, 
+                                              const ProtoStruct& extra) = 0;
+
     /// @brief Reports if the arm is in motion.
     virtual bool is_moving() = 0;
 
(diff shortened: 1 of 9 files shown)


src/viam/sdk/components/private/arm_client.cpp
```diff
@@ -46,6 +46,31 @@ void ArmClient::move_to_joint_positions(const std::vector<double>& positions,
...
+                      request.mutable_options()->set_max_vel_degs_per_sec(
+                          *options.max_vel_degs_per_sec);
+                  }
+
+                  if (options.max_acc_degs_per_sec2) {
+                      request.mutable_options()->set_max_acc_degs_per_sec2(
+                          *options.max_acc_degs_per_sec2);
+                  }
+
+                  for (const std::vector<double>& pos : positions) {
```
acmorrow, line 65: You can say `const auto& pos` here, if you like.

src/viam/sdk/components/private/arm_server.cpp
```diff
@@ -53,6 +53,32 @@ ::grpc::Status ArmServer::MoveToJointPositions(
...
+    ::grpc::ServerContext*,
+    const ::viam::component::arm::v1::MoveThroughJointPositionsRequest* request,
+    ::viam::component::arm::v1::MoveThroughJointPositionsResponse*) noexcept {
+    return make_service_helper<Arm>(
+        "ArmServer::MoveThroughJointPositions", this, request)([&](auto& helper, auto& arm) {
+        std::vector<std::vector<double>> positions;
+
+        positions.reserve(request->positions_size());
+        for (const auto& values : request->positions()) {
+            positions.push_back({values.values().begin(), values.values().end()});
```
acmorrow, line 66: This can possibly be written `positions.emplace_back(values.values().begin(), values.values().end)` which will avoid a move and construct the vector in place.

acmorrow (commented):
I was just taking a look out of general interest

src/viam/sdk/components/arm.hpp
```diff
@@ -103,6 +109,20 @@ class Arm : public Component, public Stoppable {
     virtual void move_to_joint_positions(const std::vector<double>& positions,
                                          const ProtoStruct& extra) = 0;
 
+    /// @brief Move each joint on the arm through the positions specified in @param positions
+    /// @param options optional specifications to be obeyed during the motion.
+    inline void move_through_joint_positions(const std::vector<std::vector<double>>& positions,
```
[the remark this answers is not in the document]
raybjork: I did this because I'm new to C++ and thought this would be an easy solution.  Tensors sound much better I think

lia-viam (approved):

PR #325, opened 2024-11-15
RSDK6734 - Remove viam/api imports from public headers

stuqdog:
Moves all(1) proto conversion out of public files and into either the relevant client or server (when specific to a particular resource), or into the new `private/proto_conversions.xpp` (when used in multiple places). The one exception is `private/encoder.xpp`, which contains methods only used for encoders, but used by both client and server code.

A lot of this can be skimmed over/ignored, as it's just mechanical moves of methods with straightforward changes (e.g., taking an object of type `T` as an arg instead of being a method in class `T`).

Notable changes are as follows (I have tried to add comments indicating where these occur as well):
* add `private/client_helper.hpp`. all it does is `#include <google/protobuf/struct.pb.h>` so that the public client helper doesn't have to.
* remove `id_` field from `LinkConfig` type. This was never used and was impossible to access. Technically breaking since the shape of a `LinkConfig` has changed.
* A new constructors (see `ResourceConfig`) and getters (see `WorldState`). Shouldn't be anything controversial, they getters return `const&` and the constructors just take the data members.
* Changed `get_resource_name` in a resource to 1) no longer take a `name` argument (we were always just using the resource name so we might as well not ask for it), and 2) return a `Name` instead of a protobuf `ResourceName`. This is technically breaking though I doubt anyone was relying on this method to return a `ResourceName` before.

Also a couple flybys here and there (mostly removing unused imports, removed some proto conversion methods that were unused, removing an unused `stub` created in the `dial` method). 

 (1) technically `viam/api` imports still exist for `robot_server.hpp` and `module_service.hpp`. Given that these are both explicitly defining an implementation of a proto type, and that we want these to be public, I question whether removing the `viam/api` imports from these is possible or desirable.


stuqdog:
Cool, thanks for the comprehensive reply! I think it might be good to talk non-async on this since there's probably going to be a fair amount of back and forth, but one quick high level clarification that I think will help drive that conversation:

> I'm not sure that one giant proto conversions header makes the most sense from a user experience point of view since it's kind of giant and might be harder to find what you're looking for, and also involves pulling in a lot of potentially not relevant stuff, eg if I want to get some conversions for X component, I'm also pulling in everything to do with module service, spatial math, etc

So as I understand it, a major goal here is that from a user experience point of view, proto no longer exists. The proto conversions header is intentionally private private specifically because a user should never have to do proto conversions with any of our internal types. So I'm not sure we should be worrying about how a user finds this, or including any of this in `some_component.xpp` at all.

lia-viam:
The related work is almost done and I think I'm done cherry-picking from this PR too, going to close

PR #327, opened 2024-11-18
use docker base image in CI

abe-winter:
## What changed
- consume prebuilt base image in test.yml rather than building every time
- split test.yml build + test into separate steps (so the location of errors is clearer)
- incidental: add clang-format to base image
## Follow-up work
- scheduled rebuilds of base image
## Sample run
https://github.com/viamrobotics/viam-cpp-sdk/actions/runs/11903253787

### DIFF
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -30,6 +30,12 @@ jobs:
     steps:
     - uses: actions/checkout@v4
 
+    - uses: docker/login-action@v3
+      with:
+        registry: ghcr.io
+        username: ${{ github.actor }}
+        password: ${{ secrets.GITHUB_TOKEN }}
+
     # build base (if inputs.build-base)
     - uses: docker/metadata-action@v5
       id: base-meta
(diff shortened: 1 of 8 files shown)


lia-viam:
for my benefit and posterity can you elaborate/eli5 a bit on consuming the pre-built base image? like where is that image coming from, is it manually uploaded rn or could be scheduled in the future, etc

abe-winter:
> where is that image coming from, is it manually uploaded rn or could be scheduled in the future, etc

the image is created by this job https://github.com/viamrobotics/viam-cpp-sdk/actions/workflows/docker.yml

and it gets uploaded to here https://github.com/viamrobotics/viam-cpp-sdk/pkgs/container/cpp-base/307921886?tag=bullseye-amd64

(if you scroll down on that page, note it embeds the git sha so we have a fairly good idea of where this came from; not sure if the sha is attested by the runner, but they're giving us a blue check on the right side of the page; I think some attestation is happening)

could easily add a cron trigger to docker.yml if we wanted

abe-winter:
these docker images are also listed on the side of the repo main page if you need to find them in a hurry

image

lia-viam (approved):

PR #331, opened 2024-11-25
remove instability notice

abe-winter:
## What changed
- Remove the 'instability' banner from the top of the readme
## Why
Guessing we wrote this before we were doing tagged releases; now that we have version numbers, breaking changes are less destructive.


purplenicole730 (approved):
Thanks for the reasoning! The versions making this notice no longer effective makes sense to me

lia-viam:
let's merge this after I finish the stuff I'm doing with getting API/proto imports out of headers as that's already set to include some breaking changes

stuqdog (approved):

PR #335, opened 2024-12-03
RSDK-6574: Fix use of `time_point`

lia-viam:
Fixes the incorrect `std::chrono::time_point` type used throughout the SDK. I have opted to call this `viam::sdk::time_pt` so as not to create name clashes with `time_point` for anyone doing `using namespace std::chrono`.

### DIFF
--- a/src/viam/sdk/common/utils.cpp
+++ b/src/viam/sdk/common/utils.cpp
@@ -22,8 +22,6 @@
 namespace viam {
 namespace sdk {
 
-using time_point = std::chrono::time_point<long long, std::chrono::nanoseconds>;
-
 std::vector<unsigned char> string_to_bytes(const std::string& s) {
     std::vector<unsigned char> bytes(s.begin(), s.end());
     return bytes;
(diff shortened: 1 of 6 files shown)


src/viam/sdk/common/utils.hpp
```diff
@@ -18,30 +18,28 @@ const std::string kService = "service";
 const std::string kRDK = "rdk";
 const std::string kBuiltin = "builtin";
 
+using time_pt = std::chrono::time_point<std::chrono::system_clock, std::chrono::nanoseconds>;
```
acmorrow, line 21: What motivated the switch from calling it `time_point` to `time_pt`?

src/viam/sdk/common/utils.cpp
```diff
@@ -34,21 +32,23 @@ std::string bytes_to_string(const std::vector<unsigned char>& b) {
     return img_string;
 };
 
-time_point timestamp_to_time_pt(const google::protobuf::Timestamp& timestamp) {
-    const std::chrono::seconds seconds(timestamp.seconds());
-    const std::chrono::nanoseconds nanos(timestamp.nanos());
-    return time_point(std::chrono::duration_cast<std::chrono::system_clock::duration>(seconds) +
-                      nanos);
+time_pt timestamp_to_time_pt(const google::protobuf::Timestamp& timestamp) {
```
acmorrow, line 35: Would this later be unified with #332 ?

acmorrow (commented):
Just interested. I'm happy to see this is being fixed!

src/viam/sdk/common/utils.hpp
```diff
@@ -18,30 +18,28 @@ const std::string kService = "service";
...
-    const std::chrono::time_point<long long, std::chrono::nanoseconds>& time_pt);
+/// @brief convert a time_point to a google::protobuf::Timestamp.
+google::protobuf::Timestamp time_pt_to_timestamp(time_pt);
 
 std::vector<unsigned char> string_to_bytes(std::string const& s);
 std::string bytes_to_string(std::vector<unsigned char> const& b);
 
 std::chrono::microseconds from_proto(const google::protobuf::Duration& proto);
-google::protobuf::Duration to_proto(const std::chrono::microseconds& duration);
+google::protobuf::Duration to_proto(std::chrono::microseconds duration);
```
stuqdog, line 42: (q) curious why we removed the `const&`ness of the `duration` arg?

src/viam/sdk/common/utils.hpp
```diff
@@ -18,30 +18,28 @@ const std::string kService = "service";
...
     static viam::common::v1::ResponseMetadata to_proto(const response_metadata& metadata);
 };
 
 bool operator==(const response_metadata& lhs, const response_metadata& rhs);
 
-/// @brief convert a google::protobuf::Timestamp to
-/// std::chrono::time_point<long long, std::chrono::nanoseconds>
-std::chrono::time_point<long long, std::chrono::nanoseconds> timestamp_to_time_pt(
-    const google::protobuf::Timestamp& timestamp);
+/// @brief convert a google::protobuf::Timestamp to time_point
```
stuqdog, line 32: (nit) `s/time_point/time_pt`

src/viam/sdk/common/utils.hpp
```diff
@@ -18,30 +18,28 @@ const std::string kService = "service";
...
+time_pt timestamp_to_time_pt(const google::protobuf::Timestamp& timestamp);
 
-/// @brief convert a std::chrono::time_point<long long, std::chrono::nanoseconds> to
-/// a google::protobuf::Timestamp.
-google::protobuf::Timestamp time_pt_to_timestamp(
-    const std::chrono::time_point<long long, std::chrono::nanoseconds>& time_pt);
+/// @brief convert a time_point to a google::protobuf::Timestamp.
```
stuqdog, line 35: (nit) `s/time_point/time_pt`

stuqdog (approved):
a question and a nit, otherwise lgtm!

src/viam/sdk/common/utils.hpp
lia-viam: I find people do 'using namespace std::chrono' pretty often so I think this would be a name clash with our `time_point` and `std::chrono::time_point<Clock, Duration>`

src/viam/sdk/common/utils.cpp
lia-viam: yes I look forward to doing so haha

src/viam/sdk/common/utils.hpp
lia-viam: See eg https://stackoverflow.com/a/3009568, which applies here because a `std::chrono::microseconds` is just a single integer type under the hood

stuqdog: ahh that makes sense, thanks!

PR #347, opened 2025-01-03
remove api imports from examples

lia-viam:
Cherry-picked from #325

### DIFF
--- a/src/viam/examples/dial/example_dial.cpp
+++ b/src/viam/examples/dial/example_dial.cpp
@@ -13,10 +13,6 @@
 #include <grpcpp/grpcpp.h>
 #include <grpcpp/support/status.h>
 
-#include <viam/api/common/v1/common.pb.h>
-#include <viam/api/robot/v1/robot.grpc.pb.h>
-#include <viam/api/robot/v1/robot.pb.h>
-
 #include <viam/sdk/components/generic.hpp>
 #include <viam/sdk/robot/client.hpp>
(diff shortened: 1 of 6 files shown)


src/viam/examples/dial/example_dial.cpp
```diff
@@ -13,10 +13,6 @@
 #include <grpcpp/grpcpp.h>
 #include <grpcpp/support/status.h>
 
-#include <viam/api/common/v1/common.pb.h>
```
acmorrow, line 16: ⬆️ - Are the `grpcpp` headers still required in this file and others?

lia-viam: In this particular file no! But while we're on the subject `gizmo/api.hpp` and `summation/api.hpp` both include `grpcpp/channel.h`. This could be replaced with the `grpc_fwd.hpp` file but I'm not sure if it's worth the effort, thoughts?

lia-viam: Or rather, not that it takes much effort but that it's a little awkward and perhaps it's a non-goal for users to not have google imports in headers

acmorrow: I wouldn't bother in the ones that are doing custom APIs, since in that case I think the user is up to something complex, but for cases where it is just using the standard C++ SDK API I'd drop out the grpc headers if possible I think. Otherwise they will get copied in endlessly.

PR #349, opened 2025-01-06
RSDK-9549: Make RpcSubtype model api ResourceRPCSubtype

lia-viam:
Changes the implementation of an SDK type to more closely model its corresponding API type. Note that handler_map.cpp contains what is more or less the to_proto and from_proto of RPCSubtype, so we remove the ServiceDescriptor member.

The comparison operators in this and some other classes have been optimized to not construct a concatenated string every time they're called

Also includes some drive-by changes for domino-effect missing includes, minor cleanups, etc 

Per @stuqdog we're not sure why RPCSubtype originally had a ServiceDescriptor member--it seems its only use was in using the DebugString to implement `operator<` to make RPCSubtype usable as a `map` key. I'm not sure what tests if any will exercise that this is still behaving as expected or if we know why this might have been implemented that way in the first place, cc @acmorrow for any ideas or sugestions

### DIFF
--- a/src/viam/examples/modules/simple/client.cpp
+++ b/src/viam/examples/modules/simple/client.cpp
@@ -1,3 +1,4 @@
+#include <iostream>
 #include <memory>
 #include <string>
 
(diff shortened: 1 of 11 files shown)


stuqdog (approved):
lgtm though per offline conversation, before we merge I would love to see confirmation that we can actually run this and use resources (both built-ins and new modular types) after removing the `ServiceDescriptor`.

lia-viam:
Validated with a test robot on app dot viam that we were still able to get components/modules and exercise methods on them

PR #350, opened 2025-01-07
RSDK-9672 Insulate grpc Status

lia-viam:
Takes grpc status includes out of public headers

Note that service_helper has been moved to private for the time being--if there is demand we can re-publish it but it is not currently used anywhere and provides some awkward headaches for google header insulation


src/viam/sdk/common/client_helper.hpp
```diff
@@ -138,13 +135,13 @@ class ClientHelper {
 
         const auto result = reader->Finish();
 
-        if (result.ok() ||
-            (cancelled_by_handler && result.error_code() == ::grpc::StatusCode::CANCELLED)) {
+        if (result.ok() || (cancelled_by_handler &&
+                            client_helper_details::isStatusCancelled(result.error_code()))) {
```
stuqdog, line 139: (nit) delete blank newline here?

stuqdog (approved):

PR #354, opened 2025-01-13
RSDK-9731 Remove v2 namespace prefix for proto conversion

lia-viam:
Now that we're done the migration the v2 namespace is no longer necessary.

In the process of doing this it turned out the v2 namespace was having more than cosmetic effects--I have changed `to_proto` and `from_proto` to regular functions rather than callable objects, and introduced using declarations for them in various private `client.cpp` and `server.cpp` files so that in cases where there is a private `impl::to_proto`, overload resolution still finds the omni-conversion functions in the parent namespace.


stuqdog (approved):
nice, lgtm! 🚀

PR #356, opened 2025-01-15
Remove DiscoverComponents

randhid:
We're fully deprecating this API on March 10th 2025, but since the SDK team is gearing up to remove the instability notice and release v0.1.0, I thought perhaps we could remove this API from the C++ from the get-go.

This is my first contribution to the cpp-sdk, apologies for any obvious mistakes!


lia-viam:
This is great thanks for the PR! Had to rebase against main since the last merge but will merge when the build passes.

lia-viam (approved):

PR #357, opened 2025-01-15
RSDK-9740: Make google libraries a private dependency of the SDK 

lia-viam:


src/viam/examples/modules/simple/CMakeLists.txt
```diff
@@ -19,6 +19,7 @@ add_executable(simple_module
 target_link_libraries(simple_module
   PRIVATE Threads::Threads
   viam-cpp-sdk::viamsdk
+  ${VIAMCPPSDK_GRPCXX_LIBRARIES}
```
acmorrow, line 22: I'm a little surprised that simple_module has a direct link dependency on grpc?

src/viam/examples/modules/tflite/CMakeLists.txt
```diff
@@ -44,6 +44,7 @@ if (VIAMCPPSDK_BUILD_TFLITE_EXAMPLE_MODULE)
   target_link_libraries(tflite_module
     PRIVATE Threads::Threads
     PRIVATE viam-cpp-sdk::viamsdk
+    PRIVATE ${VIAMCPPSDK_GRPCXX_LIBRARIES}
```
acmorrow, line 47: Ditto tflite module dependency on grpc

src/viam/examples/project/cmake/CMakeLists.txt
```diff
@@ -72,4 +74,5 @@ target_link_libraries(example_dial
 
 target_link_libraries(example_module
   viam-cpp-sdk::viamsdk
+  gRPC::grpc++
```
acmorrow, line 77: Same question

src/viam/sdk/CMakeLists.txt
```diff
@@ -236,13 +236,13 @@ target_link_directories(viamsdk
...
 
 target_link_libraries(viamsdk
-  PUBLIC ${VIAMCPPSDK_GRPCXX_LIBRARIES}
-  PUBLIC ${VIAMCPPSDK_GRPC_LIBRARIES}
-  PUBLIC ${VIAMCPPSDK_PROTOBUF_LIBRARIES}
   PUBLIC Boost::headers
   PUBLIC Boost::log
   PUBLIC xtensor
   PRIVATE ${VIAMCPPSDK_GRPCXX_REFLECTION_LIBRARIES}
+  PRIVATE ${VIAMCPPSDK_GRPCXX_LIBRARIES}
```
acmorrow, line 243: 🎉

src/viam/sdk/CMakeLists.txt
```diff
@@ -236,13 +236,13 @@ target_link_directories(viamsdk
...
+  PRIVATE ${VIAMCPPSDK_GRPC_LIBRARIES}
+  PRIVATE ${VIAMCPPSDK_PROTOBUF_LIBRARIES}
```
acmorrow, line 245: ⬇️  - I have a suspicion there may no longer be any actual direct usage of abseil in viamsdk. Can you please check?

src/viam/examples/dial_api_key/CMakeLists.txt
```diff
@@ -19,6 +19,7 @@ add_executable(example_dial_api_key
 target_link_libraries(example_dial_api_key
   Boost::program_options
   viam-cpp-sdk::viamsdk
+  ${VIAMCPPSDK_GRPCXX_LIBRARIES}
```
acmorrow, line 22: I don't see any actual use of grpc in this examples sources.

acmorrow (changes requested):
This is really great. I think several of the places where grpc is being tagged as a dependency are perhaps vestigial?

src/viam/sdk/CMakeLists.txt
lia-viam: happily you're right!

src/viam/examples/project/cmake/CMakeLists.txt
```diff
@@ -45,6 +45,8 @@ find_package(Threads REQUIRED)
 #
 find_package(viam-cpp-sdk CONFIG REQUIRED viamsdk)
 
+find_package(gRPC CONFIG REQUIRED)
```
acmorrow, line 48: Still needed?

BUILDING.md
```diff
@@ -27,15 +27,15 @@ guide, before continuing further.
...
+[xtensor](https://github.com/xtensor-stack/xtensor) >= 0.24.3
+and on the transitive dependencies of
 those projects.
 
 You will need to install these required dependencies before building
 the Viam C++ SDK, preferably by way of your system package manager:
 
-- Debian: `apt-get install cmake build-essential libabsl-dev libboost-all-dev libgrpc++-dev libprotobuf-dev libxtensor-dev`
-- MacOS with Homebrew: `brew install abseil cmake boost grpc protobuf xtensor`
+- Debian: `apt-get install cmake build-essential libboost-all-dev libgrpc++-dev libprotobuf-dev libxtensor-dev`
```
acmorrow, line 37: I think there are some references in the top level CMakeLists about absl as well.

acmorrow (changes requested):

src/viam/examples/project/cmake/CMakeLists.txt
lia-viam: Oops good catch thank you

conanfile.py
```diff
@@ -47,11 +47,11 @@ def requirements(self):
...
         # The SDK supports older grpc and protobuf, but these are the oldest
         # maintained conan packages.
-        self.requires('grpc/[>=1.48.4]', transitive_headers=True)
-        self.requires('protobuf/[>=3.17.1]', transitive_headers=True)
+        self.requires('grpc/[>=1.48.4]')
+        self.requires('protobuf/[>=3.17.1]')
 
         self.requires('xtensor/[>=0.24.3]')
-        self.requires('abseil/[>=20230125.3]', transitive_libs=True)
+        self.requires('abseil/[>=20230125.3]')
```
acmorrow, line 54: Is this actually still needed? I'd expect absl to be a transitive dep of proto/grpc so not needing to be mentioned here?

conanfile.py
```diff
@@ -94,6 +94,8 @@ def package_info(self):
            if self.settings.os in ["Linux", "FreeBSD"]:
             self.cpp_info.components[component].system_libs = ["pthread"]
 
+        self.cpp_info.components["viamapi"].requires.append("abseil::absl_strings")
```
acmorrow, line 97: Guessing this is related to above inclusion.

acmorrow (approved):
LGTM, just a question.

conanfile.py
lia-viam: I was surprised by this too but yes--my working theory is that it's because the API gencode does include some abseil headers/need some of their symbols, this was making the build fail with undefined symbol errors when I took it out

conanfile.py
lia-viam: Exactly!

PR #359, opened 2025-01-21
RSDK-9719: Simple module sensor

lia-viam:
Replaces the prior "simple" module example with one that just implements a Sensor component, in line with eg Python SDK https://github.com/viamrobotics/viam-python-sdk/tree/main/examples/simple_module

### DIFF
--- a/src/viam/examples/modules/simple/client.cpp
+++ b/src/viam/examples/modules/simple/client.cpp
@@ -3,9 +3,9 @@
 #include <string>
 
 #include <viam/sdk/common/proto_value.hpp>
+#include <viam/sdk/components/sensor.hpp>
 #include <viam/sdk/robot/client.hpp>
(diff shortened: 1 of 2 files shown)


src/viam/examples/modules/simple/client.cpp
```diff
@@ -33,20 +33,36 @@ int main() {
...
+        std::cerr << "could not get 'mysensor' resource from robot\n";
         return EXIT_FAILURE;
     }
 
     ProtoStruct command{{"hello", "world"}};
-    ProtoStruct resp = printer->do_command(command);
+    ProtoStruct resp = sensor->do_command(command);
 
     if (command != resp) {
         std::cerr << "Got unexpected result from 'printer1'\n";
```
stuqdog, line 47: `s/printer1/mysensor`

PR #374, opened 2025-02-25
RSDK-9298: Instance class 

lia-viam:
Introduces a one-and-only `Instance` class for the SDK. For the time being this only has a `Registry`, but it will pave the way for inclusion of better logging and resource-level logging in particular. 

Some of the machinery in `instance.cpp` for singleton management was adapted from https://github.com/mongodb/mongo-cxx-driver/blob/02ad73bc306eb9044eb691de96ad41c13107738d/src/mongocxx/lib/mongocxx/v_noabi/mongocxx/instance.cpp

A few classes were updated to now get a pointer to a `Registry` rather than calling the no-longer-`static` `Registry` methods. The existence of `Instance::current()` means technically these could just get the `Registry` from the current instance, or possibly that `Registry` could have a `static Registry& current()` method...open to feedback here.

Every example has been updated to define an instance object, with the same boilerplate comment put in front explaining its usage


src/viam/sdk/module/service.cpp
```diff
@@ -207,31 +208,36 @@ Dependencies ModuleService::get_dependencies_(
 
 std::shared_ptr<Resource> ModuleService::get_parent_resource_(const Name& name) {
     if (!parent_) {
-        parent_ = RobotClient::at_local_socket(parent_addr_, {0, boost::none});
+        parent_ = RobotClient::at_local_socket(parent_addr_, {0, boost::none}, registry_);
```
lia-viam, line 211: Question for reviewers--I am wondering if `RobotClient` should be part of the `Instance` as well, it seems like we always want it? Looking through the examples I believe the only scenario where `RobotClient` is not constructed explicitly is one where we just construct a `ModuleService` but that ends up constructing a `RobotClient` in this method anyway.

Looking ahead to the logger PR, the logger will need a `RobotClient` for the gRPC Log call, so we may want to avoid a situation where `RobotClient` is getting delayed instantiation and just construct one on application startup.

stuqdog: I think I'd prefer to keep `RobotClient` distinct. 1) I think of the `Instance` as being a bit of a black box; the user knows they need one and that they need to keep it around, but they don't need to know anything about what's inside. By contrast, a `RobotClient` is something that users may interact with directly and frequently. 2) There should only ever be one live instance, whereas one could imagine a user writing client code that interacts with multiple robots and so therefore has multiple robot clients.

lia-viam: Those points make sense. Follow-ups
- if there are multiple `RobotClient` should log messages be getting sent to the `Log` call of each of them? 
- given the comment about `Instance` insulation, how do we feel about now having to do `inst.registry()` for some of the constructors?

stuqdog: Hmm good questions!

> if there are multiple RobotClient should log messages be getting sent to the Log call of each of them?

I think this is a question we can largely ignore, thankfully. If there are multiple `RobotClients` then the user is interacting with purely client code and so the logs should be showing up in the shell, so we dont have to call a `RobotClient::log` at all.

If on the other hand, we _do_ have to send logs via the `Log` method then we're dealing with a module. But, each module is started up as its own process for a single Robot, so we can always only send logs via the `Log` call to the module's parent.

> given the comment about Instance insulation, how do we feel about now having to do inst.registry() for some of the constructors?

Caveat: I will readily confess to not being particularly hip to C++ conventions. If what I'm saying flies in the face of idiomatic C++ design then obviously we should prefer remaining idiomatic.

I personally prefer to not do that. Per our offline conversation yesterday and to reiterate some of that: I'd love to hide as much as possible from users about the instance, and ideally not really expose any methods at all to the user about it. If it's needed in particular cases, I'd prefer to just pass the `instance` as an argument and let the `RobotClient` constructor, e.g., deal with getting the `Registry` from a private method via some friend class setup.

I don't think this is always clean and easy to do, as when registering custom resources, but I imagine we could add a `register_resource` method to a `RobotClient` that under the hood interacts with the registry. I think this is a bit cleaner and a bit more intuitive (when you want a robot client to deal with a custom resource, you register that resource with the client!). A possible downside here is this pattern is slightly misleading: we're saying "register it with the robot client" but in fact we're registering with the singleton registry, which means another robot client would _also_ be able to work with that resource type. But I don't think this is particularly harmful, and we can work around it if it is.

src/viam/sdk/common/instance.cpp
```diff
@@ -0,0 +1,53 @@
...
+    if (!current_instance.load()) {
+        // This variable declaration calls the default ctor, storing a current instance.
+        static Instance inst;  // NOLINT (misc-const-correctness)
+    }
+
+    Instance* current = current_instance.load();
+
+    if (current == reinterpret_cast<Instance*>(&sentinel)) {
+        throw Exception("instance was destroyed");
+    }
```
stuqdog, line 47: (q) pardon my ignorance, I don't think I fully understand this. Under what conditions will this exception throw?

stuqdog (commented):
Nice, this is super exciting! Thanks so much for putting this together, it's a complicated one but a big win.

I've done a first pass; there's a lot to think about here so I think there will be more back and forth but here are some initial thoughts.

src/viam/sdk/common/instance.hpp
```diff
@@ -0,0 +1,28 @@
...
+
+#include <memory>
+
+namespace viam {
+namespace sdk {
+
+/// @brief Instance management for Viam C++ SDK applications.
+/// This is a single instance class which is responsible for global setup and teardown related to
+/// the SDK. An Instance must be constructed before doing anything else in a program, and it must
+/// remain alive in a valid state for the duration of the program. Creating multiple overlapping
```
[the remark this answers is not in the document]
lia-viam: There's no difficulty per se, we can just tell people "don't ever do this, it's UB" 🙂

src/viam/sdk/common/instance.cpp
lia-viam: Oh wait actually this gets to your comment above about making it an error to have multiple instances even if non-overlapping, so it already is an error to have multiple of them

src/viam/sdk/robot/client.hpp
```diff
@@ -147,6 +147,7 @@ class RobotClient {
     status get_machine_status() const;
 
    private:
+    Registry* registry_;
```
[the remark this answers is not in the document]
lia-viam: Maybe cc @acmorrow when he gets a sec, but wondering if `Registry` can just get `friend` access to instance and be able to return a reference to the registry member rather than having to access it through `Instance::current` with the `atomic::load`s. My though is that since we say it's UB to use any Viam types without an `Instance` already constructed we can skip the whole "create if not exists" logic of `Instance::current` 

I agree that the performance penalty can probably be ignored here, but if so then we could do 

```
static Registry& Registry::get()
{
    static Registry& result = Instance::current().impl_->registry;
    return result;
}
```
in which case the `static` would only be initialized one-and-done style

src/viam/sdk/common/instance.hpp
```diff
@@ -0,0 +1,28 @@
...
+/// This is a single instance class which is responsible for global setup and teardown related to
+/// the SDK. An Instance must be constructed before doing anything else in a program, and it must
+/// remain alive in a valid state for the duration of the program. Creating multiple Instance
+/// objects in the same program is an error.
+class Instance {
+   public:
+    Instance();
+    ~Instance();
+
+    static Instance& current();
```
acmorrow, line 18: Do you want to doc-comment this, particularly if it is mostly intended for testing.

src/viam/sdk/registry/registry.cpp
```diff
@@ -90,6 +91,12 @@ const Model& ModelRegistration::model() const {
     return model_;
 };
 
+Registry& Registry::get() {
+    static Registry& result = Instance::current().impl_->registry;
```
acmorrow, line 95: - This will bring `Instance` into existence even if one wasn't created, I think. Those maybe aren't the right semantics, since then a user of the library could get away with not creating an `Instance` themselves.
- Why the static& rather than just calling `Instance::current()` each time.

src/viam/sdk/registry/registry.cpp
```diff
@@ -173,61 +182,48 @@ const google::protobuf::ServiceDescriptor* Registry::get_service_descriptor_(
...
-        BOOST_LOG_TRIVIAL(warning)
-            << "Attempted to initialize the Registry but it was already initialized.";
-        return;
-    }
-    initialized_ = true;
     register_resources();
     grpc::reflection::InitProtoReflectionServerBuilderPlugin();
 }
 
-std::unordered_map<std::string, std::shared_ptr<const ModelRegistration>> Registry::resources_;
```
acmorrow, line 226: 👍

src/viam/sdk/registry/registry.hpp
```diff
@@ -103,26 +103,28 @@ class ModelRegistration {
 /// @brief A registry of known resources.
 class Registry {
    public:
+    /// @brief Get the application-wide instance of Registry.
+    static Registry& get();
```
acmorrow, line 107: Or `instance`? I also wonder a bit why to expose this on `Registry` rather than on `Instance`.

src/viam/sdk/registry/registry.cpp
lia-viam: - I think you're right that those aren't the semantics we want--maybe it should be something like `Instance::current(bool create_if_not_exists = false)`
- This was the result of a discussion me and @stuqdog were going back and forth on, and we thought that `Instance` should be a total black box so that it's in some sense an implementation detail that `Instance` has-a `Registry` or, later, `Logging-Core`. In either case though, I think that the divergences from our use cases and Mongo's mean that `Instance::current` does not end up being unit tests only

src/viam/sdk/registry/registry.hpp
lia-viam: Fair question--elaborated in another comment

src/viam/sdk/common/instance.cpp
```diff
@@ -0,0 +1,57 @@
+#include <viam/sdk/common/instance.hpp>
+
+#include <viam/sdk/common/exception.hpp>
+#include <viam/sdk/common/private/instance.hpp>
+#include <viam/sdk/registry/registry.hpp>
+
+#include <atomic>
```
stuqdog, line 7: (nit) can we put this `include` block above the `viam/sdk` include block?

stuqdog (approved):
nice, a couple nits but otherwise lgtm!

PR #376, opened 2025-02-28
RSDK-6158 - initial connection options

stuqdog:


src/viam/sdk/rpc/dial.cpp
```diff
@@ -60,10 +60,18 @@ void DialOptions::set_entity(boost::optional<std::string> entity) {
...
+void DialOptions::set_initial_connection_attempts(int attempts) {
+    initial_connection_attempts_ = attempts;
+}
+
 void DialOptions::set_timeout(std::chrono::duration<float> timeout) {
     timeout_ = std::move(timeout);
 }
 
+void DialOptions::set_initial_connection_attempt_timeout(std::chrono::duration<float> timeout) {
+    initial_connection_attempt_timeout_ = std::move(timeout);
```
lia-viam, line 72: minor but you can remove the `std::move` here, a `chrono::duration<T>` is a `T` under the hood, no need to `std::move` a `float`.

I am seeing that this was also how it was done in `set_timeout` above, so maybe that can be a drive-by

src/viam/sdk/rpc/dial.hpp
```diff
@@ -49,11 +56,15 @@ class DialOptions {
     const boost::optional<std::string>& entity() const;
     bool allows_insecure_downgrade() const;
     const std::chrono::duration<float>& timeout() const;
+    int initial_connection_attempts() const;
+    const std::chrono::duration<float>& initial_connection_attempt_timeout() const;
```
lia-viam, line 60: Let's return this by value instead of ref to `const`; see other comment

src/viam/sdk/rpc/dial.cpp
```diff
@@ -84,6 +100,34 @@ bool DialOptions::allows_insecure_downgrade() const {
...
+        attempts_remaining = -1;
+    }
+    opts.set_timeout(opts.initial_connection_attempt_timeout());
+
+    while (attempts_remaining > 0) {
+        try {
+            auto connection = dial(uri, opts);
+            opts.set_timeout(timeout);
+            return connection;
+        } catch (...) {
```
lia-viam, line 123: Let's just catch `const std::exception&` here, a `catch (...)` is generally inadvisable except at application boundaries.

src/viam/sdk/rpc/dial.cpp
```diff
@@ -84,6 +100,34 @@ bool DialOptions::allows_insecure_downgrade() const {
     return allow_insecure_downgrade_;
 }
 
+std::shared_ptr<ViamChannel> ViamChannel::dial(const char* uri,
+                                               const boost::optional<DialOptions>& options,
+                                               bool initial_attempt) {
+    if (!initial_attempt) {
+        return dial(std::move(uri), options);
```
lia-viam, line 107:
```suggestion
        return dial(uri, options);
```

lia-viam (changes requested):

src/viam/sdk/rpc/dial.hpp
```diff
@@ -49,11 +56,15 @@ class DialOptions {
...
 
     void set_entity(boost::optional<std::string> entity);
     void set_credentials(boost::optional<Credentials> creds);
     void set_allow_insecure_downgrade(bool allow);
     void set_timeout(std::chrono::duration<float> timeout);
+    void set_initial_connection_attempts(int attempts);
```
acmorrow, line 66: A small proposal: these and the other `set_x` methods should return `DialOptions&` so you can chain them:

```
DialOptions options{};
options.set_entity(...).set_timeout(...)....;
```

src/viam/sdk/rpc/dial.hpp
```diff
@@ -16,9 +16,16 @@ class ViamChannel {
    public:
     void close();
     ViamChannel(std::shared_ptr<GrpcChannel> channel, const char* path, void* runtime);
+
     static std::shared_ptr<ViamChannel> dial(const char* uri,
                                              const boost::optional<DialOptions>& options);
 
+    // allows for specifying that this dial is an initial connection attempt, indicating that
+    // initial connection options should be used.
+    static std::shared_ptr<ViamChannel> dial(const char* uri,
```
acmorrow, line 25: I have something of a personal dislike for `bool` parameters like this, and it makes me wonder why this `initial` aspect isn't either just a setting on the `DialOptions`, or a differently named method.

My distaste arrises because at the call site, you end up like this:

```
ViamChannel::dial(my_uri, my_options, false);
```

What does `false` mean?

Whereas either of these makes the intention clear at the call site without needing to chase down the declaration / documentation for `dial`:

```
ViamChannel::dial(my_uri, my_options.as_initial());
```

or

```
ViamChannel::dial_initial(my_uri, my_options)
```

src/viam/sdk/rpc/dial.cpp
acmorrow: I wonder if `clang-tidy` will think of it as a missed move? Personally, I always `move` things like these, since otherwise it feels like I'm making assumptions about implementation.

acmorrow (commented):
Just taking a look, random thoughts, feel free to do with them as you see fit.

src/viam/sdk/rpc/dial.cpp
lia-viam: I guess that's fair in general but in this case also worth noting that `duration` does not have move constructors

stuqdog: I generally err on the side of `move` unless `clang` explicitly tells me not to do so just because it's simpler and makes me less likely to miss a `move` that _is_ relevant but I don't feel particularly strongly. I'm happy to go with whatever makes sense to y'all!

stuqdog: Ended up going with @lia-viam's suggestion just because it was the one that came in first. But again, I'm not picky and my slight inclination if anything is to keep `move`ing things even if it's unnecessary. So if you two land on the other approach I'm happy to change it back.

lia-viam (approved):

src/viam/sdk/rpc/dial.hpp
```diff
@@ -16,9 +16,15 @@ class ViamChannel {
    public:
     void close();
     ViamChannel(std::shared_ptr<GrpcChannel> channel, const char* path, void* runtime);
+
     static std::shared_ptr<ViamChannel> dial(const char* uri,
                                              const boost::optional<DialOptions>& options);
 
+    // Specifies that this dial is an initial connection attempt and that initial connection options
+    // should be used.
+    static std::shared_ptr<ViamChannel> dial_initial(const char* uri,
```
acmorrow, line 25: I'm a little curious how one is to use `dial` and `dial_initial` now. If I'm writing a loop to connect, does it look like this:

```
bool first = false;
while true {
    if (first) {
        first = false;
        if (auto chan = dial_initial(...); chan)
            return chan;
    } else {
        if (auto chan = dial(...); chan)
            return chan;
    }
}
```

It feels a little awkward for the user.

src/viam/sdk/rpc/dial.cpp
```diff
@@ -72,18 +91,51 @@ const boost::optional<Credentials>& DialOptions::credentials() const {
...
+        try {
+            auto connection = dial(uri, opts);
+            opts.set_timeout(timeout);
+            return connection;
+        } catch (const std::exception&) {
+            attempts_remaining -= 1;
+        }
+    }
+
+    throw Exception(ErrorCondition::k_connection, "Unable to establish connection");
```
acmorrow, line 136: Should this maybe instead throw whatever the last exception was, so that the information about the cause of the failure isn't lost?

src/viam/sdk/rpc/dial.hpp
```diff
@@ -16,9 +16,15 @@ class ViamChannel {
    public:
     void close();
     ViamChannel(std::shared_ptr<GrpcChannel> channel, const char* path, void* runtime);
+
     static std::shared_ptr<ViamChannel> dial(const char* uri,
                                              const boost::optional<DialOptions>& options);
 
+    // Specifies that this dial is an initial connection attempt and that initial connection options
```
acmorrow, line 23: This seems as good a time as any to doc comment this function, as well as `dial` above.

acmorrow (changes requested):

src/viam/sdk/rpc/dial.hpp
stuqdog: I think in most cases users shouldn't use `dial_initial` at all. The preferred entrypoint for creating a robot client is `RobotClient::at_address(...)` which calls `dial` itself, so the user doesn't have to know about either.

If a user _does_ need to create a channel themselves for whatever reason, then they would just call `dial_initial(...)` if they want to specify an initial timeout at that time (say they're trying to dial to a robot as it's starting up so they want to retry), or `dial(...)` otherwise. The `while true` loop above is redundant; `dial_initial(...)` is a wrapper around calls to `dial` already.

src/viam/sdk/rpc/dial.hpp
stuqdog: Done; added a clause as well clarifying that use of `dial` is discouraged and that `RobotClient::at_address(...)` is preferred.

src/viam/sdk/rpc/dial.hpp
stuqdog: Just for explicitness: if you want something akin to this loop where you try to connect forever until it succeeds, you would do the following:

```
DialOptions opts;
opts.set_initial_connection_attempts(0); // 0 is a sentinel value that says "try forever"
return ViamChannel::dial_initial(uri, opts);
```

acmorrow: Got it, thanks for clarifying.

src/viam/sdk/rpc/dial.cpp
```diff
@@ -72,18 +91,53 @@ const boost::optional<Credentials>& DialOptions::credentials() const {
...
+
+    std::exception exception;
+    while (attempts_remaining > 0) {
+        try {
+            auto connection = dial(uri, opts);
+            opts.set_timeout(timeout);
+            return connection;
+        } catch (const std::exception& e) {
+            attempts_remaining -= 1;
+            exception = e;
```
acmorrow, line 134: This is going to slice the exception, I think (i.e. only the `std::exception` part will be copied, but any subtype will be lost). I think you can just look at `attempts_remaining` inside the `catch` and `throw` if there are non left.

acmorrow (approved):
LGTM

PR #378, opened 2025-03-07
RSDK-9901 - run tests in update proto workflow

stuqdog:
Sets a new token for checkout step to cause tests to run during proto update workflow.

Unlike https://github.com/viamrobotics/viam-python-sdk/pull/864, we use `GIT_ACCESS_TOKEN` instead of `REPO_READ_TOKEN`. `REPO_READ_TOKEN` does not exist in this repo's secrets, and per @abe-winter the `GIT_ACCESS_TOKEN` _may_ work as well. If it does not, we can revisit.

### DIFF
--- a/.github/workflows/update_protos.yml
+++ b/.github/workflows/update_protos.yml
@@ -14,6 +14,8 @@ jobs:
       image: ghcr.io/viamrobotics/canon:amd64
     steps:
       - uses: actions/checkout@v4
+        with:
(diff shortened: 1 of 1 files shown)


lia-viam:
This seems straightforward and I'm provisionally ready to approve, I guess we won't know for sure if it worked until it's merged and a new proto update comes in though?

njooma (approved):

PR #379, opened 2025-03-10
Fix conan build failures

lia-viam:
Conan build has been broken as of https://github.com/viamrobotics/viam-cpp-sdk/actions/runs/13652553282

It seems that the release of grpc 1.69 conan package is causing a version conflict with abseil. However, as of #357 there is no longer any direct consumption of abseil in SDK/API code, so it should be fine to remove the dependency from conan entirely.


stuqdog (commented):
Abstractly this looks reasonable to me, though I'm curious as to why the tests are now failing?

stuqdog (approved):
per offline convo, lgtm as a "let's lock in what gains we can" fix. if/when it becomes relevant, we can revisit why a couple tests are still failing.

PR #387, opened 2025-03-21
RSDK-10305 resources should support close

stuqdog:
* adds `close` method to resources
* calls `close` on a resource during a module reconfigure call if the method doesn't have `reconfigure` implemented

### DIFF
--- a/src/viam/sdk/module/service.cpp
+++ b/src/viam/sdk/module/service.cpp
@@ -115,6 +115,7 @@ struct ModuleService::ServiceImpl : viam::module::v1::ModuleService::Service {
             Registry::get().lookup_model(cfg.name());
         if (reg) {
             try {
+                res->close();
                 const std::shared_ptr<Resource> res = reg->construct_resource(deps, cfg);
                 manager->replace_one(cfg.resource_name(), res);
             } catch (const std::exception& exc) {
--- a/src/viam/sdk/resource/resource.cpp
+++ b/src/viam/sdk/resource/resource.cpp
@@ -26,5 +26,8 @@ Name Resource::get_resource_name() const {
     return get_resource_name(kResource);
 }
 
+// default behavior is to just return
+void Resource::close() {}
+
 }  // namespace sdk
 }  // namespace viam
--- a/src/viam/sdk/resource/resource.hpp
+++ b/src/viam/sdk/resource/resource.hpp
@@ -25,6 +25,9 @@ class Resource {
     /// @brief Return the resource's name.
     virtual std::string name() const;
 
+    /// @brief Closes a resource
+    virtual void close();
+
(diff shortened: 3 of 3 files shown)


src/viam/sdk/module/service.cpp
```diff
@@ -115,6 +115,7 @@ struct ModuleService::ServiceImpl : viam::module::v1::ModuleService::Service {
             Registry::get().lookup_model(cfg.name());
         if (reg) {
             try {
+                res->close();
```
stuqdog, line 118: @randhid per what you were discussing in slack, if we're trying to reconfigure a modular resource that doesn't implement reconfigure we will now 1) stop it if it's stoppable, 2) call its `close` method always, and then 3) rebuild it

src/viam/sdk/resource/resource.cpp
```diff
@@ -26,5 +26,8 @@ Name Resource::get_resource_name() const {
     return get_resource_name(kResource);
 }
 
+// default behavior is to just return
+void Resource::close() {}
```
stuqdog, line 30: @randhid this was the implementation detail I wanted to run by you. looking at golang, the default behavior for `close` is a no-op. is that desired here as well?

randhid: Is it true that that is the default behaviour? Mind pointing me to where you are  seeing that?

My experience is needing to opt-into the no-op behaviour by embedding a `resource.TriviallyCloseable` in my resource class, or opting into always rebuild behaviour by including `resource.AlwaysRebuild`, my understanding hwne having to implement go modules is to have to implement close using those two or a Close implementation, or it just won't build.

stuqdog: Ahh no, I'm mistaken. The default is a no-op for the `TriviallyCloseable` specifically, you are correct!

> by embedding a `resource.TriviallyCloseable` in my resource class, or opting into always rebuild behaviour by including `resource.AlwaysRebuild`

I think I'm a little confused by this, how does `AlwaysRebuild` affect the `Close` behavior? since a `Resource` needs to define `Close` and `Close` could certainly be called outside the context of a `Reconfigure`, doesn't including `AlwaysRebuild` still require us to define `Close`?

randhid: I'm sorry I should have been clearer. I was trying to explain that there are two embeddings available from the resource package: `TriviallyCloseable,` which provides a no-op closer, and `AlwaysRebuild,` which forces a close and rebuild flow. You are correct - `AlwaysRebuild` requires you to implement a Close.

lia-viam:
this is getting closed as won't do, with the recommendation that such logic should probably go in a destructor implementation

PR #391, opened 2025-03-24
Draft: RSDK-10311: Switch component

lia-viam:
Claude-assisted switch component

Currently this does not compile because the buf (or protoc? unclear) gencode sometimes creates `namespace switch_` and other times `namespace switch`, and the latter is not a legal namespace. Will have to figure out a workaround.


acmorrow (approved):
The CMake parts of this look good to me. Thanks for doing this the right way with CMake, rather than dropping down to unix tools which would have broken my in-progress windows work.

PR #393, opened 2025-03-25
conformant ref qualified overloads

lia-viam:
MSVC revealed that the way I was previously overloading member functions was not legal. In particular you can't do

```cpp
struct X {
    void f(); // no ref qualifier
    void f() const& // ref qualifier
};
```


acmorrow (approved):
LGTM

PR #394, opened 2025-03-26
[RSDK-10284]: MLModelService: don't check for name if there is only 1 expected tensor, return error if expected tensors not present

bhaney:
In creating the C++ tflite mlmodel module, there was a difference in expected behavior from the Go tflite mlmodel module. 

When using the Go module, if the ml model expected an input tensor called, let's say, "serving_default_images:0", but the user served an input tensor called "image", there is an assumption in the Go implementation that if there is only one input tensor, and the model expects only one tensor, then the input should be fed in regardless of the name.

In the C++ code, the C++ SDK itself implements a stricter criteria, that if the user-fed input tensors do not have a named tensor that matches the size and type of what an ml model expects, that tensor will not even make it to the inputs of the MLModelService implementation. It will be filtered out of the inputs. 

This change makes the C++ SDK less strict. If the metadata expects one tensor, and the input provides one tensor, then it won't do a name check, but still check the type and size. It will also return an error if a required input tensor is missing (in the case of inputs > 1)


src/viam/sdk/services/private/mlmodel_server.cpp
```diff
@@ -36,29 +36,10 @@ ::grpc::Status MLModelServiceServer::Infer(
...
-            if (tensor_type != input.data_type) {
-                std::ostringstream message;
-                using ut = std::underlying_type<MLModelService::tensor_info::data_types>::type;
-                message << "Tensor input `" << input.name << "` was the wrong type; expected type "
-                        << static_cast<ut>(input.data_type) << " but got type "
-                        << static_cast<ut>(tensor_type);
-                return helper.fail(::grpc::INVALID_ARGUMENT, message.str().c_str());
-            }
-            inputs.emplace(std::move(input.name), std::move(tensor));
+	for (const auto& [tensor_name, api_tensor] : request->input_tensors().tensors()) {
```
acmorrow, line 40: If we do this, then we are relying on every mlmodelservice implementation to do its own checking and error handling. Maybe they already do, maybe they don't. Given that they were allowed to assume, previously, that `MLModelServiceServer` was defending them from mismatched / mistyped inputs, it seems quite possible that they are not.

I'm wondering how narrowly we could scope the exception. Would it still work if we:
- Only applied the exception if metadata says there is only one input and we only received one input tensor, and kept the original logic otherwise?
- Continued to check the types match in the 1 vs 1 case?

Meanwhile, I think a meaningful error should be returned when validating against the metadata: sorry for missing this TODO and making tensors disappear! Could you please add an error case there?

acmorrow (changes requested):

bhaney:
only `mlmodel_server.cpp` changed, I just ran linting on the whole directory

bhaney:
only `mlmodel_server.cpp` changed, but I ran the linter on the whole repo

src/viam/sdk/services/private/mlmodel_server.cpp
bhaney: Did your first suggestion, only apply exception if metadata says there is one input, and there is one input.

bhaney: and also added the error message

src/viam/sdk/services/private/mlmodel_server.cpp
```diff
@@ -12,12 +12,11 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-#include <viam/sdk/services/private/mlmodel_server.hpp>
-
 #include <viam/sdk/common/private/service_helper.hpp>
 #include <viam/sdk/rpc/server.hpp>
 #include <viam/sdk/services/mlmodel.hpp>
 #include <viam/sdk/services/private/mlmodel.hpp>
+#include <viam/sdk/services/private/mlmodel_server.hpp>
```
acmorrow, line 19: It looks like maybe your editor got a little aggressive in its formatting opinions, here and in other files. Can you put back the header ordering? It actually is intentional the way it is, with the associated header coming first in its own group (helps ensure the header is standalone).

src/viam/sdk/services/private/mlmodel_server.cpp
```diff
@@ -58,7 +53,35 @@ ::grpc::Status MLModelServiceServer::Infer(
...
                 return helper.fail(::grpc::INVALID_ARGUMENT, message.str().c_str());
             }
-            inputs.emplace(std::move(input.name), std::move(tensor));
+            inputs.emplace(tensor_pair.first, std::move(tensor));
+        } else {
+            // Normal case: multiple tensors, do metadata checks
+            for (const auto& input : md.inputs) {
+                const auto where = request->input_tensors().tensors().find(input.name);
+                if (where == request->input_tensors().tensors().end()) {
+                    // Ignore any inputs for which we don't have metadata, since
```
acmorrow, line 62: Comment is stale.

acmorrow (approved):
LGTM

src/viam/sdk/services/private/mlmodel_server.cpp
bhaney: undid the bad linting

acmorrow: This one should go back up as well.

acmorrow:
LGTM mod fixing the (now only one) moved header inclusion and fixing up the now-stale comment a bit.

src/viam/sdk/services/private/mlmodel_server.cpp
```diff
@@ -58,7 +53,35 @@ ::grpc::Status MLModelServiceServer::Infer(
                         << static_cast<ut>(tensor_type);
                 return helper.fail(::grpc::INVALID_ARGUMENT, message.str().c_str());
             }
-            inputs.emplace(std::move(input.name), std::move(tensor));
+            inputs.emplace(tensor_pair.first, std::move(tensor));
+        } else {
+            // Normal case: multiple tensors, do metadata checks
+	    // If there are extra tensors in the inputs that not found in the metadata,
+	    // they will not be passed on to the implementation.
```
bhaney, line 60: Added a comment to say that if the user provides more inputs than are necessary, the extra inputs will be skipped over

src/viam/sdk/services/private/mlmodel_server.cpp
bhaney: done

PR #399, opened 2025-04-01
pin cmake for conan jammy

lia-viam:
cmake 4.0 is causing a similar build failure as in the release build


lia-viam:
folding this into #390

PR #401, opened 2025-04-03
basic windows and macos build compatibility improvements

acmorrow:
Small fixes needed for building the C++ SDK on Windows. This is far from complete - this is just the stuff that seems reasonable to land as is. I'll leave some comments to explain a few things.


CMakeLists.txt
```diff
@@ -150,6 +150,10 @@ if (VIAMCPPSDK_ENFORCE_COMPILER_MINIMA)
     if(CMAKE_CXX_COMPILER_VERSION VERSION_LESS 10.0)
       message(FATAL_ERROR "Insufficient Apple clang version: XCode 10.0+ required")
     endif()
+  elseif (CMAKE_CXX_COMPILER_ID STREQUAL "MSVC")
```
acmorrow, line 153: This establishes a compiler minimum for Windows. Since there is no history here, the newest Visual Studio seems like fair game. We can always try to downgrade it later if we find a need.

src/viam/examples/camera/example_camera.cpp
```diff
@@ -1,6 +1,5 @@
 #include <fstream>
 #include <string>
-#include <unistd.h>
```
acmorrow, line 3: I don't know why all these files included `<unistd.h>` and at least on my mac they appear to compile without it. We will see what CI says.

src/viam/sdk/CMakeLists.txt
```diff
@@ -34,8 +34,18 @@ configure_file(common/grpc_fwd.hpp.in common/grpc_fwd.hpp)
 
 # Set compile and link options based on arguments
 if (VIAMCPPSDK_USE_WALL_WERROR)
-  target_compile_options(viamsdk PRIVATE -Wall -Werror)
+  if (CMAKE_CXX_COMPILER_ID STREQUAL "Clang")
```
acmorrow, line 37: This enables warnings as errors flags for MSVC, and makes it an error if you ask for it and we can't do it on the current toolchain (or the toolchain is unknown).

src/viam/sdk/components/camera.hpp
```diff
@@ -9,7 +9,12 @@
 #include <vector>
 
 #include <boost/endian/conversion.hpp>
+
```
acmorrow, line 12: Well, this happened: https://github.com/xtensor-stack/xtensor/pull/2829#issuecomment-2770772047

I could think of no more graceful way to handle this. I don't love using `__has_include` but here we are. I think it argues in favor of maybe moving to C++17.

Another way to deal with this would be via configure checks in CMake, but I don't want to spend time working through that.

I'm open to suggestions on this.

src/viam/sdk/module/service.cpp
```diff
@@ -4,8 +4,6 @@
 #include <memory>
 #include <sstream>
 #include <string>
-#include <sys/socket.h>
```
acmorrow, line 7: I think these were only needed for the creation of the socket below, which we are no longer doing.

src/viam/sdk/module/service.cpp
```diff
@@ -115,8 +113,8 @@ struct ModuleService::ServiceImpl : viam::module::v1::ModuleService::Service {
             Registry::get().lookup_model(cfg.name());
         if (reg) {
             try {
-                const std::shared_ptr<Resource> res = reg->construct_resource(deps, cfg);
-                manager->replace_one(cfg.resource_name(), res);
+                const std::shared_ptr<Resource> resource = reg->construct_resource(deps, cfg);
```
acmorrow, line 116: Windows throws a shadowing warning here which gets promoted to an error because we have `res` above.

src/viam/sdk/robot/client.cpp
```diff
@@ -216,7 +215,7 @@ void RobotClient::refresh_every() {
             std::this_thread::sleep_for(std::chrono::seconds(refresh_interval_));
             refresh();
 
-        } catch (std::exception& exc) {
+        } catch (std::exception&) {
```
acmorrow, line 218: Windows warning that this was unused.

src/viam/sdk/module/service.cpp
```diff
@@ -252,13 +250,8 @@ ModuleService::~ModuleService() {
 }
 
 void ModuleService::serve() {
-    const mode_t old_mask = umask(0077);
```
acmorrow, line 255: I forgot to mention this in my batch of comments. This code, which has been here forever, has always done nothing. It sure _looked_ like it was making the socket that we were going to use for communication with RDK, but look carefully and you can see that we never actually bind the socket anywhere, or invoke it in any way. The actual socket gets made by gRPC.

src/viam/sdk/components/camera.hpp
lia-viam: it occurs to me that `__has_include` is a C++17 feature and also the breaking xtensor PR included came hot on the heels of a C++17 migration (https://github.com/xtensor-stack/xtl/pull/283) so if the compiler doesn't have `__has_include` maybe we can default to the old include path

lia-viam (approved):
this looks good overall, `__has_include` comment would be nice to address but non-blocking

src/viam/sdk/components/camera.hpp
acmorrow: That's a good observation. I think though, that with how I've written it, that's the practical effect. If the platform is C++17 then `__has_include` _will_ be `define`'d. If `__has_include` isn't defined, then we aren't C++17, so we couldn't be using the newer `xtl`, so the old header is the right one to pick.

Or are you suggesting just testing for C++17 directly and choosing the header based on that, without using `__has_include` at all.

PR #402, opened 2025-04-07
[RSDK-10385] Windows build system improvements and rust_utils workarounds

acmorrow:
- Ensure `__cplusplus` is set right on Windows builds.
- Don't download or link to `viam_rust_utils` on Windows.
- Redirect the `viam_rust_utils` entry points to `abort` on Windows.
- Adjust `viamapi` library name on Windows and apply `/WHOLEARCHIVE`
- Use `std::invoke_result_t` instead of `result_of` in C++17, where the latter is deprecated
- Add windows headers where needed.

### DIFF
--- a/src/viam/sdk/CMakeLists.txt
+++ b/src/viam/sdk/CMakeLists.txt
@@ -132,6 +132,7 @@ target_sources(viamsdk
     rpc/dial.cpp
     rpc/server.cpp
     rpc/private/viam_grpc_channel.cpp
+    rpc/private/viam_rust_utils_stubs.cpp
     services/discovery.cpp
     services/generic.cpp
     services/mlmodel.cpp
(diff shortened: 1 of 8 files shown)


src/viam/sdk/module/service.cpp
```diff
@@ -1,3 +1,9 @@
+#ifdef _WIN32
+#define NOMINMAX
+#include <winsock2.h>
```
acmorrow, line 3: I don't love this. I'm surprised the gRPC headers don't handle this for themselves, but without this this file doesn't compile.

The right way to do this would be with a global config precompiled header that got pulled into every `.cpp` we have, but I don't want to spend time on that right now, so I will file a follow-up ticket.

src/viam/sdk/CMakeLists.txt
```diff
@@ -132,6 +132,7 @@ target_sources(viamsdk
     rpc/dial.cpp
     rpc/server.cpp
     rpc/private/viam_grpc_channel.cpp
+    rpc/private/viam_rust_utils_stubs.cpp
```
lia-viam, line 135: instead of doing an `#if WIN32` in the cpp file, what if we did this at the CMake level:
```
if (NOT WIN32)
  target_link_libraries(viamsdk
    PRIVATE viam_rust_utils
  )
else()
    target_sources(viamsdk PRIVATE rust_utils_stubs.cpp)
endif()

```

lia-viam (approved):
I think this looks fine assuming the main tests pass. Conan is failing but it's only as broken as it already was on main (which seems to have gotten worse recently) so not a blocker for merging. 

Probably want to look into github CI for the conan package at some point however

src/viam/sdk/CMakeLists.txt
acmorrow: Sure, I'll give that a try, and as long as it works (which I expect it will) I'll merge with that.

src/viam/sdk/module/service.cpp
acmorrow: Filed https://viam.atlassian.net/browse/RSDK-10389

PR #406, opened 2025-04-08
[RSDK-10385] Misc conan improvements

acmorrow:
- The presets shouldn't get checked in, so, gitignore them
- We use xtensor publicly, so the dependency should have transitive headers.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -12,3 +12,6 @@
 
 # ignore doxygen output
 /etc/docs/api/*
+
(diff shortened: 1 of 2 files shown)


acmorrow:
@lia-viam - There's a failed workflow run here, and I don't see immediately how it could have been caused by my changes. Any thoughts?

lia-viam:
@acmorrow that's actually our current best level of build passing for this pipeline so you're all good!

PR #407, opened 2025-04-09
[RSDK-10385] Windows signal handling

acmorrow:
Claude got this one started, I tidied it up a bit. Overall, I'd like to find a better way to do this that integrates with the new `Instance` mechanism, but getting something working is currently a higher priority than elegance/consistency. I'm happy to file a follow-up ticket later to do this more gracefully if we can accept this as an interim solution.

### DIFF
--- a/src/viam/sdk/module/signal_manager.cpp
+++ b/src/viam/sdk/module/signal_manager.cpp
@@ -1,22 +1,79 @@
 #include <viam/sdk/module/signal_manager.hpp>
 
+#ifdef _WIN32
+#include <windows.h>
+#else
 #include <csignal>
 #include <pthread.h>
+#endif
+
+#include <stdexcept>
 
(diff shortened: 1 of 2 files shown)


lia-viam (approved):
fine with merging this as is, but i wonder if going forward this may be past a certain threshold of `ifdef` proliferation where we want to split sources into `signal_manager_win32.cpp` and `signal_manager_posix.cpp`

acmorrow:
> fine with merging this as is, but i wonder if going forward this may be past a certain threshold of `ifdef` proliferation where we want to split sources into `signal_manager_win32.cpp` and `signal_manager_posix.cpp`

I totally agree. I'm going to file an issue to revisit the signal handler infrastructure, and note both this idea and the idea of connecting it to the instance singleton (where I suspect it belongs).

PR #408, opened 2025-04-09
try to override b2

lia-viam:
See https://github.com/conan-io/conan-center-index/issues/27165


stuqdog (commented):
abstractly this looks reasonable to me but tests are failing, so 😢 

please re-request review when tests pass and I can look again/approve!

lia-viam:
@stuqdog thanks for taking a look! i am going to close this one, it seems for now that we might be screwed because i don't know if it's possible to override versions of a build dependency for another target. just waiting on developments in the conan ticket for now

PR #409, opened 2025-04-09
RC version bump

lia-viam:


acmorrow:
@lia-viam - Looks like something doesn't like that release tag format.

acmorrow:
Let's not move forward with this since CMake seems unable to handle rc versions (what?). Anyway, I'll find another way to do testing without this.

lia-viam:
@acmorrow did some digging and it seems like all the components must be nonnegative integers, see https://cmake.org/cmake/help/latest/command/project.html

i wouldn't be opposed to using the `[tweak]` to do this, eg `0.8.0.1` or something

acmorrow:
Let's just get #410 merged and call it 0.8.0. Meanwhile, could you possibly also take a quick look at https://github.com/viam-modules/mlmodel-tflite/pull/25, which is the same thing? I can't seem to add you as a reviewer - maybe you aren't part of that org?

acmorrow (approved):
LGTM

PR #411, opened 2025-04-11
[RSDK-10385] Don't enforce compiler minima or complain about warnings for conan builds

acmorrow:
@lia-viam - Speculatively putting this PR up while I wait for a workflow run to validate that it gets the results I need. If so, I'd like to merge this and tag an 0.8.1 release. I'll follow up either way on this PR with results.


lia-viam:
the bullseye failure is expected, bookworm is worth re-running but not a blocker

acmorrow (commented):
@lia-viam - Updated, PTAL. This fixes the compilation issue with VS2019.

src/viam/sdk/services/motion.hpp
```diff
@@ -142,7 +136,7 @@ class Motion : public Service {
         std::string execution_id;
 
         /// @brief An ordered list of plan steps.
-        struct steps steps;
+        steps steps;
```
lia-viam, line 139: looks like this is still causing issues, now with additional compilers. a couple thoughts. so i think usually our alias declarations would use camel case, and if the declarations above were `using Step  = ...` and `using Steps = ...` there would be no issue here.

that said, i'm sort of thinking we should just write `std::vector<[Ss]tep>` and call it a day. the alias for `unordered_map` feels ergonomically justified, a `vector` alias less so. and generally speaking we don't have many instances in the sdk of doing `using meows = std::vector<meow>` and i suppose i'm not crazy about the practice in general.

acmorrow: Yeah, that makes sense. I've made that change.

acmorrow:
Updated, PTAL. Hopefully the builders are happier with this, i'm trying it on VS 2019 right now as well.

lia-viam (approved):
lgtm, good to merge if the builds pass.

lia-viam (approved):

PR #416, opened 2025-04-18
RSDK-10366 - rust utils for windows

stuqdog:
Adds support for tcp-based rust-utils, updates rust-utils downloading for x86_64 windows.

Tested tcp connection on mac successfully.

Tested on windows: with a properly built `viam_rust_utils.lib` at the root of the directory, I am able to connect as a client via webRTC. ~~Without a `viam_rust_utils.lib` at the root of the directory, I instead link in the stubs that were previously being linked, and `abort()` out whenever making a `rust-utils` call (as we were doing previously).~~ edit: with the recent update to `rust-utils`, we now successfully download `rust-utils` if it's not in the root, so long as we're on x86_64.


src/viam/sdk/rpc/dial.cpp
```diff
@@ -148,20 +149,27 @@ std::shared_ptr<ViamChannel> ViamChannel::dial(const char* uri,
...
         uri, entity, type, payload, opts.allows_insecure_downgrade(), float_timeout.count(), ptr);
-    if (socket_path == NULL) {
+    if (proxy_path == NULL) {
         free_rust_runtime(ptr);
         throw Exception(ErrorCondition::k_connection, "Unable to establish connecting path");
     }
 
-    std::string address("unix://");
-    address += socket_path;
+    std::string localhost_prefix("127.0.0.1");
```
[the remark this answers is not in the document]
stuqdog: Yeah, the address we get back from `rust-utils` is consistently of the form `127.0.0.1:{port}` so I opted for that.

src/viam/sdk/rpc/dial.cpp
```diff
@@ -148,20 +149,27 @@ std::shared_ptr<ViamChannel> ViamChannel::dial(const char* uri,
...
+    auto tcp_check = std::mismatch(localhost_prefix.begin(), localhost_prefix.end(), proxy_path);
```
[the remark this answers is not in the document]
stuqdog: no good reason why! I asked the internet how to find a substring in C++ and somehow the first thing I landed on was an SO post recommending `std::mismatch` 😅 I'll switch to `std::string::find`.

acmorrow (approved):
LGTM!

src/viam/sdk/rpc/dial.cpp
```diff
@@ -148,20 +149,26 @@ std::shared_ptr<ViamChannel> ViamChannel::dial(const char* uri,
...
         throw Exception(ErrorCondition::k_connection, "Unable to establish connecting path");
     }
 
-    std::string address("unix://");
-    address += socket_path;
+    const std::string localhost_prefix("127.0.0.1");
+    std::string address;
+    if (std::string(proxy_path).find(localhost_prefix) == std::string::npos) {
+        // proxy path is not a localhost address and is therefore a unix domain socket (UDS)
+        address += "unix:";
```
acmorrow, line 163: There is another instance of `unix://` in this repo at `RobotClient::at_local_socket`, which should probably also move to the new form, as well as a comment in `robot/client.hpp` and in `src/viam/examples/README.md`. Maybe get them all as part of this review if this one is being changed?

acmorrow:
@stuqdog - I'd like to take another look through this before it merges, if that is OK?

acmorrow:
If it is ready, that is!

.github/workflows/conan.yml
```diff
@@ -40,6 +40,63 @@ jobs:
...
+      - name: Install python
+        uses: actions/setup-python@v5
+        with:
+          python-version: '3.12'
+
+      - name: Install Scoop
+        uses: MinoruSekine/setup-scoop@v4.0.1
+
+      - name: install buf 'n stuff
+        run: scoop install buf ninja
```
acmorrow, line 71: Do you still need any of this if you have choco?

.github/workflows/conan.yml
```diff
@@ -40,6 +40,63 @@ jobs:
...
+
+      - name: Install dependencies
+        uses: crazy-max/ghaction-chocolatey@v3
+        with:
+          args: install -y conan git 7zip jq
```
acmorrow, line 76: Do you actually need 7zip or jq for this?

.github/workflows/conan.yml
```diff
@@ -40,6 +40,63 @@ jobs:
...
+
+      # Ensures that things installed with choco are visible to us
+      - name: refresh env
+        run: |
+          Import-Module $env:ChocolateyInstall\helpers\chocolateyProfile.psm1
+          refreshenv
```
acmorrow, line 82: I'm not sure this does anything, actually, when it is in its own named step.

.github/workflows/conan.yml
```diff
@@ -40,6 +40,63 @@ jobs:
...
+          conan profile detect
+          conan create . `
+            --build=missing `
+            -o "&:shared=False" `
+            -s:h compiler.cppstd=17 `
+            -c:h tools.microsoft:winsdk_version=10.0.17763.0 `
+            -s:h compiler.runtime=static
```
acmorrow, line 95: I wouldn't expect any of these four arguments to be needed for CI. For tflite_cpu they were required in order to get a static executable that would run on Windows 10 Build 1089 without requiring the VC redistributable, but those constraints are not in effect here.

.github/workflows/release.yml
```diff
@@ -32,9 +33,88 @@ jobs:
...
+        run: |
+          Import-Module $env:ChocolateyInstall\helpers\chocolateyProfile.psm1
+          refreshenv
+          conan profile detect
+          conan create . `
+            --build=missing `
+            -o "&:shared=False" `
+            -s:h compiler.cppstd=17 `
+            -c:h tools.microsoft:winsdk_version=10.0.17763.0 `
+            -s:h compiler.runtime=static
```
acmorrow, line 95: As above

CMakeLists.txt
```diff
@@ -37,6 +37,9 @@ if (WIN32)
   # Use a newer version of CMake on Windows, so we are NEW for
   # https://cmake.org/cmake/help/latest/policy/CMP0149.html
   cmake_minimum_required(VERSION 3.27 FATAL_ERROR)
+
+  # set boost compiler to avoid errors with boost log
+  set(Boost_COMPILER "vc143")
```
acmorrow, line 42: This should't be necessary: the boost infra should be making decisions based on the compiler actually in use.

CMakeLists.txt
```diff
@@ -256,6 +259,11 @@ set(viam_rust_utils_file ${CMAKE_CURRENT_BINARY_DIR}/${CMAKE_SHARED_LIBRARY_PREF
 
 file(GLOB viam_rust_utils_files ${PROJECT_SOURCE_DIR}/${CMAKE_SHARED_LIBRARY_PREFIX}viam_rust_utils*${CMAKE_STATIC_LIBRARY_SUFFIX})
 
+# Check if the architecture is one that we support in rust-utils
+set(WINDOWS_ARCHS "AMD64" "X86" "X64")
```
acmorrow, line 263: Per this SO answer: https://stackoverflow.com/a/70498851, there are only four values possible on windows: `AMD64`, `IA64`, `ARM64`, `EM64T`, and `X86`.

- You can safely ignore `IA64` (Itanium) and `EM64T` (not seen since XP).
- That leaves `AMD64` and `ARM64` for 64-bit Intel and ARM, and `x86` for 32-bit Intel Windows.

I doubt rust-utils is going to support 32-bit Windows or 64-bit ARM windows?

So it feels to me like the only value that matters here is `AMD64`, in which case you can just test that directly down on 277.

CMakeLists.txt
```diff
@@ -266,24 +274,43 @@ if (viam_rust_utils_files)
     ${viam_rust_utils_file}
     ONLY_IF_DIFFERENT
   )
-elseif(NOT WIN32) # TODO(RSDK-10366): Currently, rust_utils is not published for windows, so don't even try downloading
+elseif(NOT WIN32 OR rust_utils_supported_windows_arch) # TODO(RSDK-10366): Currently, rust_utils is not published for windows aarch, so don't even try downloading
   set(lvru_system_name ${CMAKE_SYSTEM_NAME})
-  if (CMAKE_SYSTEM_NAME STREQUAL "Darwin")
+  if (CMAKE_SYSTEM_NAME STREQUAL "Cygwin")
```
acmorrow, line 279: Is supporting cygwin something important?

CMakeLists.txt
```diff
@@ -266,24 +274,43 @@ if (viam_rust_utils_files)
...
+    set(lvru_system_name "windows")
+  elseif (CMAKE_SYSTEM_NAME STREQUAL "Darwin")
     set(lvru_system_name "macosx")
   endif()
+
+  set(lvru_system_processor ${CMAKE_SYSTEM_PROCESSOR})
+  if (WIN32 AND rust_utils_supported_windows_arch)
+    message(WARNING "windows and rust-utils-supported architecture is true")
+    set(lvru_system_processor "x86_64")
```
acmorrow, line 288: This sort of reinforces my point above about `WINDOWS_ARCHS`.

CMakeLists.txt
```diff
@@ -266,24 +274,43 @@ if (viam_rust_utils_files)
...
+  endif()
+  message(WARNING "system processor is ${lvru_system_processor}")
+
+  set(shared_library_prefix ${CMAKE_SHARED_LIBRARY_PREFIX})
+  if (WIN32)
+    set(shared_library_prefix "lib")
+  endif()
+  message(WARNING "shared library prefix is ${CMAKE_SHARED_LIBRARY_PREFIX}")
+  message(WARNING "trying to download https://github.com/viamrobotics/rust-utils/releases/latest/download/libviam_rust_utils-${lvru_system_name}_${lvru_system_processor}${CMAKE_STATIC_LIBRARY_SUFFIX}")
```
acmorrow, line 297: There is an inherent conflict in windows library naming between shared (DLL) and static, because if you have `foo.dll` then you need an "import library" for that DLL, and that is normally called `foo.lib`. But if you want a static library as well then you can't call it `foo.lib` because that would conflict with the DLL import library. So, often, the static library has a `foo_static.lib` to distinguish it from the dynamic one. Other conventions are possible, but there is no strict rule here. From what I understand, the C++ SDK prefers to obtain the static version. So, the suffix is going to depend on how rust-utils publishes the relevant artifacts.

conanfile.py
```diff
@@ -93,8 +93,8 @@ def package_info(self):
 
         # TODO(RSDK-10366): Currently, rust_utils is not published for windows
         # and the C++ SDK just doesn't include it as a dependency on that platform
-        if not self.settings.os == "Windows":
```
stuqdog, line 96: @lia-viam can you help me understand what this is doing? Based on testing on the windows laptop, it doesn't seem to matter whether we run this always or not. Same below.

CMakeLists.txt
```diff
@@ -283,25 +284,20 @@ elseif(NOT WIN32) # TODO(RSDK-10366): Currently, rust_utils is not published for
...
     "$<BUILD_INTERFACE:${CMAKE_CURRENT_BINARY_DIR}>"
-    "$<INSTALL_INTERFACE:${CMAKE_INSTALL_LIBDIR}>"
   )
 
   set_property(TARGET viam_rust_utils PROPERTY IMPORTED_LOCATION ${viam_rust_utils_file})
-
-  install(
-    IMPORTED_RUNTIME_ARTIFACTS viam_rust_utils
-    LIBRARY COMPONENT viam-cpp-sdk_runtime
-  )
```
stuqdog, line 304: per offline conversation with @acmorrow, this appears to largely be artifacts of when we dynamically linked rust-utils.

src/viam/examples/mlmodel/example_audio_classification_client.cpp
```diff
@@ -377,7 +377,7 @@ int main(int argc, char* argv[]) try {
             });
 
             // Print out the top 5 (or fewer) label/score pairs.
-            for (size_t i = 0; i != std::min(5UL, scored_labels.size()); ++i) {
+            for (size_t i = 0; i != 5UL && i != scored_labels.size(); ++i) {
```
stuqdog, line 380: For some reason windows didn't like the `std::min` invocation; this should be logically equivalent.

src/viam/examples/modules/complex/gizmo/api.hpp
```diff
@@ -30,12 +30,14 @@ class Gizmo : public Component {
     explicit Gizmo(std::string name);
 };
 
-namespace viam::sdk {
+namespace viam {
+namespace sdk {
```
stuqdog, line 34: Make compiles on C++14.

src/viam/examples/modules/complex/summation/api.hpp
```diff
@@ -29,12 +29,14 @@ class Summation : public Service {
     explicit Summation(std::string name);
 };
 
-namespace viam::sdk {
+namespace viam {
+namespace sdk {
```
stuqdog, line 33: make compile on C++14

src/viam/sdk/tests/mocks/mock_robot.cpp
```diff
@@ -299,7 +299,8 @@ ::grpc::Status MockRobotService::FrameSystemConfig(
     const ::viam::robot::v1::FrameSystemConfigRequest*,
     ::viam::robot::v1::FrameSystemConfigResponse* response) {
     auto client_md = context->client_metadata();
-    if (auto client_info = client_md.find("viam_client"); client_info == client_md.end()) {
+    auto client_info = client_md.find("viam_client");
+    if (client_info == client_md.end()) {
```
stuqdog, line 303: make compile on C++14

conanfile.py
lia-viam: See https://docs.conan.io/2/examples/conanfile/package_info/components.html for more info, but basically we are telling conan that the SDK depends on a library called `viam_rust_utils`, and the name of the library to be linked is also `viam_rust_utils`

src/viam/examples/mlmodel/example_audio_classification_client.cpp
lia-viam: See https://stackoverflow.com/questions/13416418/define-nominmax-using-stdmin-max#13420838

I think we can safely put `#define NOMINMAX` at the top of an example cpp file like this, maybe within an `#ifdef WIN32`; that's what i'd recommend

lia-viam (approved):

.github/workflows/conan.yml
stuqdog: unfortunately I think so. possibly `choco` can take care of `ninja` for us but I was unable to install buf with anything except `scoop` (in fact, the official instructions for `buf` are to use `scoop`).

.github/workflows/conan.yml
stuqdog: probably not! I'll get rid of it.

CMakeLists.txt
stuqdog: Not that I'm aware of, I wasn't sure so I was trying to be conservative.

CMakeLists.txt
```diff
@@ -288,16 +288,17 @@ else()
     message(WARNING "Currently running on Windows with no rust-utils file. Module code should work as expected, but client code may fail unexpectedly.")
 endif()
 
-# TODO(RSDK-10637): Currently, rust_utils doesn't build correctly for windows in CI, so don't declare the library unless a user has provided their own build
-if (NOT WIN32)
+if (NOT WIN32) # build `SHARED` on unix-based systems
```
stuqdog, line 291: Building `STATIC` here caused tests to fail. So I reverted to `SHARED` (which we were already doing). But, building `SHARED` causes the builds to fail entirely on windows! So, we build `STATIC` on windows, below.

CMakeLists.txt
```diff
@@ -283,21 +289,30 @@ elseif(NOT WIN32) # TODO(RSDK-10366): Currently, rust_utils is not published for
...
+  add_library(viam_rust_utils STATIC IMPORTED)
+  target_link_directories(viam_rust_utils
+    INTERFACE
+    "$<BUILD_INTERFACE:${CMAKE_CURRENT_BINARY_DIR}>"
+  )
   set_property(TARGET viam_rust_utils PROPERTY IMPORTED_LOCATION ${viam_rust_utils_file})
+endif()
+
 
+if (NOT WIN32) # installing seems to be necessary for tests to pass in CI
```
acmorrow, line 315: I'd imagine that it should only be necessary to install `viam_rust_utils` for a static build of the C++ SDK. In `BUILD_SHARED_LIBS=ON` mode, the whole of `viam_rust_utils` will get baked into the `libviamsdk` shared library

src/viam/examples/mlmodel/example_audio_classification_client.cpp
```diff
@@ -377,7 +381,7 @@ int main(int argc, char* argv[]) try {
             });
 
             // Print out the top 5 (or fewer) label/score pairs.
-            for (size_t i = 0; i != std::min(5UL, scored_labels.size()); ++i) {
+            for (size_t i = 0; i != std::min<std::size_t>(5UL, scored_labels.size()); ++i) {
```
acmorrow, line 384: Why the explicit <std::size_t>?

src/viam/sdk/CMakeLists.txt
```diff
@@ -289,9 +289,14 @@ endif()
 
 if (APPLE)
   target_link_libraries(viamsdk PUBLIC "-framework Security")
-else()
+elseif (NOT WIN32)
```
acmorrow, line 292: I think that some of these should actually be declared as `target_link_libraries` on `viam_rust_utils`, if that is the library that actually brings these dependencies in.

.github/workflows/conan.yml
acmorrow: I'm a little puzzled about this, because as far as I know I was able to build the C++ SDK as part of the windows tflite build without scoop, or many of these other steps.

CMakeLists.txt
```diff
@@ -283,21 +289,30 @@ elseif(NOT WIN32) # TODO(RSDK-10366): Currently, rust_utils is not published for
...
     "$<BUILD_INTERFACE:${CMAKE_CURRENT_BINARY_DIR}>"
     "$<INSTALL_INTERFACE:${CMAKE_INSTALL_LIBDIR}>"
   )
+  set_property(TARGET viam_rust_utils PROPERTY IMPORTED_LOCATION ${viam_rust_utils_file})
 
+elseif (${CMAKE_SYSTEM_PROCESSOR} STREQUAL "AMD64") # build `STATIC` for windows
+  add_library(viam_rust_utils STATIC IMPORTED)
+  target_link_directories(viam_rust_utils
+    INTERFACE
+    "$<BUILD_INTERFACE:${CMAKE_CURRENT_BINARY_DIR}>"
```
acmorrow, line 309: Apologies: I've changed my mind here, I think you still do need `INSTALL_INTERFACE` here for the `BUILD_SHARED_LIBS=OFF` case.

.github/workflows/conan.yml
acmorrow: Also, this page, describing `windows-latest` claims that `cmake`, `python`, and `choco` come pre-installed:

https://github.com/actions/runner-images/blob/main/images/windows/Windows2022-Readme.md

acmorrow: Oh: I made it so the C++ SDK _downloads_ buf. So you shouldn't need to install it at all!:

https://github.com/viamrobotics/viam-cpp-sdk/blob/main/src/viam/api/CMakeLists.txt#L64-L94

acmorrow (changes requested):

CMakeLists.txt
stuqdog: I don't immediately recall, but I'll switch back to `STATIC` now and see if/how it fails!

.github/workflows/conan.yml
```diff
@@ -84,7 +84,11 @@ jobs:
           Import-Module $env:ChocolateyInstall\helpers\chocolateyProfile.psm1
           refreshenv
           conan profile detect
-          conan create .
+          conan create . `
+            --build=missing `
+            -o "&:shared=False" `
+            -s:h compiler.cppstd=17 `
+            -s:h compiler.runtime=static
```
stuqdog, line 91: @acmorrow I _think_ this (or some subset of it) is necessary for or else I get
```ERROR: There are invalid packages:
grpc/1.69.0: Invalid: grpc/1.69.0 shared not supported by Visual Studio
```
in CI.

CMakeLists.txt
stuqdog:
```
CMake Error at /__w/viam-cpp-sdk/viam-cpp-sdk/test_package/build/gcc-12-x86_64-14-release/generators/cmakedeps_macros.cmake:66 (message):
  Library 'viam_rust_utils' not found in package.  If 'viam_rust_utils' is a
  system library, declare it with 'cpp_info.system_libs' property
```

src/viam/examples/mlmodel/example_audio_classification_client.cpp
lia-viam: @acmorrow since the template wants both its arguments to be the same `typename T`, a call to `std::min(5UL, std::size_t{})` was ambiguous...vexing but I think MSVC may have been right there since `size_t` is `unsigned long long` not `unsigned long`

src/viam/examples/mlmodel/example_audio_classification_client.cpp
```diff
@@ -12,6 +12,10 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+#ifdef _WIN32
+#define NOMINMAX
```
lia-viam, line 16: i think this may no longer be necessary because of the `std::min<std::size_t>` below

conanfile.py
```diff
@@ -91,7 +91,8 @@ def package(self):
 
     def package_info(self):
 
-        self.cpp_info.components["viam_rust_utils"].libs = ["viam_rust_utils"]
+        if self.options.shared:
```
lia-viam, line 94: maybe this was covered elsewhere but what's the logic here with only doing this for a shared build?

stuqdog: Ahh yeah, so 1) I think I had it backwards and should've had `if not self.options.shared`. But possibly both are wrong! This comes from a conversation with @acmorrow about how dynamic libraries will sort of naturally include the metadata about what needs to be linked, so doing it ourselves is only necessary in specific cases.

But, I might be off-base here, or be misremembering/misunderstanding what Drew was saying! So happy to revisit as necessary.

.github/workflows/conan.yml
```diff
@@ -61,13 +61,15 @@ jobs:
...
+            #--build=missing `
+            #-o "&:shared=False"
         run: |
           Import-Module $env:ChocolateyInstall\helpers\chocolateyProfile.psm1
           refreshenv
           conan profile detect
-          conan create . `
-            --build=missing `
-            -o "&:shared=False"
+          cmake . --preset conan-default
```
lia-viam: rehashing an offline discussion, this type of invocation is suitable for `release.yml` but with `conan.yml` we do want to test that we can get a successful `conan create`, because that packages and tests the SDK conan package

.github/workflows/conan.yml
```diff
@@ -40,6 +40,39 @@ jobs:
...
+        #run: |
+          #Import-Module $env:ChocolateyInstall\helpers\chocolateyProfile.psm1
+          #refreshenv
+          #conan profile detect
+          #conan create . `
+            #--build=missing `
+            #-o "&:shared=False"
+        #env:
+          #CONAN_USER_HOME: c:/cache
+          #CONAN_USER_HOME_SHORT: c:/cache/conan_shortpaths
```
stuqdog, line 74: Not sure when we'll get around to fixing this so I don't know that I feel quite as strongly about leaving this in for future reference. Happy to just remove if preferred.

We can definitely try and fix now, but unless someone has a very clear picture of what's wrong I think I'd prefer to do that in a separate PR so we can lock in the wins here now.

lia-viam: I think maybe just leave in the TODO/ticket number but the commented code can be removed, we can always consult this diff to see what you had written I believe

lia-viam (approved):

lia-viam:
oh also were you able to run the release pipeline using this updated yaml?

.github/workflows/release.yml
```diff
@@ -36,6 +36,54 @@ jobs:
...
+    steps:
+      - name: Checkout Code
+        uses: actions/checkout@v4
+        with:
+          ref: ${{ needs.prepare.outputs.sha }}
+
+      - name: Install dependencies
+        uses: crazy-max/ghaction-chocolatey@v3
+        with:
+          args: install -y conan git
```
acmorrow, line 58: I'm still surprised you need this. Chocolatey is pre-installed on windows builders, from what I know. The tflite cpu build uses choco to install conan without doing anything special to install choco itself:

https://github.com/viam-modules/mlmodel-tflite/blob/main/bin/setup.ps1#L5

.github/workflows/release.yml
```diff
@@ -36,6 +36,54 @@ jobs:
...
+
+      - name: Create package
+        shell: powershell
+        run: |
+          Import-Module $env:ChocolateyInstall\helpers\chocolateyProfile.psm1
+          refreshenv
+          conan profile detect
+          conan install . --output-folder=build-conan --build=missing -o "&:shared=False"
+          cmake . --preset conan-default
+          cmake --build --preset=conan-release --target ALL_BUILD install -j 8
```
acmorrow, line 68: Just curious: but why direct invocation of `cmake` here rather than `conan build`?

.github/workflows/release.yml
```diff
@@ -36,6 +36,54 @@ jobs:
...
+        with:
+          args: install -y conan git
+
+      - name: Create package
+        shell: powershell
+        run: |
+          Import-Module $env:ChocolateyInstall\helpers\chocolateyProfile.psm1
+          refreshenv
+          conan profile detect
+          conan install . --output-folder=build-conan --build=missing -o "&:shared=False"
```
acmorrow, line 66: Does `conan install` actually take an `--output-folder`? I'm not sure what it would mean, since I believe this is building the dependencies.

CMakeLists.txt
```diff
@@ -283,25 +289,33 @@ elseif(NOT WIN32) # TODO(RSDK-10366): Currently, rust_utils is not published for
...
     "$<INSTALL_INTERFACE:${CMAKE_INSTALL_LIBDIR}>"
   )
-
   set_property(TARGET viam_rust_utils PROPERTY IMPORTED_LOCATION ${viam_rust_utils_file})
 
+  #if (NOT BUILD_SHARED_LIBS) # installing seems to be necessary for tests to pass in CI
   install(
-    IMPORTED_RUNTIME_ARTIFACTS viam_rust_utils
-    LIBRARY COMPONENT viam-cpp-sdk_runtime
+    FILES ${viam_rust_utils_file} DESTINATION ${CMAKE_INSTALL_LIBDIR} COMPONENT viam-cpp-sdk_runtime
```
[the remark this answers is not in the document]
acmorrow: OK. I may try it out myself after this merges to understand better, because it doesn't align with my understanding of how this ought to be working.

src/viam/examples/mlmodel/example_audio_classification_client.cpp
acmorrow: Or just make the 5 a size_t? I'd expect `std::min(size_t{5}, scored_labels.size())` to do the right thing.

src/viam/sdk/CMakeLists.txt
acmorrow: Could you, before merging, add a note here explaining that we would have liked to attach several of these dependencies to `viam_rust_utils`, but currently lack an understanding of how to do so?

src/viam/sdk/CMakeLists.txt
```diff
@@ -276,10 +276,10 @@ target_link_libraries(viamsdk
...
 )
 
-# TODO(RSDK-10366): Currently, rust_utils is not published for
-# windows, so don't link to it.  Instead, link a stub implementation
-# that just calls `abort`.
-if (NOT WIN32)
+# rust-utils doesn't build in CI on windows other than x86_64, so confirm that we're either not
+# windows, or if we are windows that we're AMD64, or else that the user has provided their own
+# build of rust_utils.
+if (NOT WIN32 OR num_viam_rust_utils_files GREATER 0 OR ${CMAKE_SYSTEM_PROCESSOR} STREQUAL "AMD64")
```
acmorrow, line 282: Could this test just look only at `num_viam_rust_utils_files` now? I'd assume that if it is greater than zero then one was found and it should be used.

acmorrow (approved):
If this is working I think you should merge it more or less as is, though I have a few comments and questions here and there that I'd appreciate your taking a look at. I don't think any of them amount to something requiring another round of review.

.github/workflows/release.yml
lia-viam: yeah output folder is where all the toolchain/build system files get put, the dependencies are built in the cache (`~/.conan2`) and then output folder contains `conan_toolchain.cmake` and the various `findLibrary.cmake` files, cmake setup scripts, etc

.github/workflows/release.yml
stuqdog: Yeah you're probably right that it's not necessary! I kept it in just because we needed to do a `choco` install _some_ way or another, it seemed totally innocuous to include, and I didn't want to do another round of testing to confirm. If you think there's a good reason to switch to a `choco install` invocation instead, let me know!

.github/workflows/release.yml
stuqdog: For reasons that weren't entirely clear, the `conan build` invocation resulted in a number of undefined symbols coming from the libraries that we were importing in the `CMakeLists.txt`. @lia-viam and I spent a short time trying to debug and then just determined that the `cmake` invocation was good enough to get releases built. Will fix in [RSDK-10666](https://viam.atlassian.net/browse/RSDK-10666).

src/viam/sdk/CMakeLists.txt
stuqdog: I may be wrong, but is this a correct assumption? I thought `num_viam_rust_utils_files` was evaluated _before_ we downloaded, in order to determine if we _should_ download a `rust_utils` file. So the `not WIN32 or AMD64` cases are important for the cases where the user hasn't provided a `rust-utils` file but we can download one for them.

.github/workflows/release.yml
acmorrow: Maybe throw a `# TODO(RSDK-10666): ...` on there.

src/viam/sdk/CMakeLists.txt
acmorrow: Sorry, you are right, I got myself tangled up. I was reacting to having complex logic to decide whether to use rust utils or build the stub here. It just feels like it would be better if all of that tortured logic was isolated up in the top level CMakeLists.txt and this file had a way of making a simple yes/no decision about whether to use the library or build the stubs.

Actually, I wonder if `if (TARGET...` is what I'm after: https://cmake.org/cmake/help/latest/command/if.html#target

Again, I know you are eager to land this, but could you please add a TODO (doesn't need a ticket, leave it for someone to clean up later).

acmorrow (approved):
LGTM with a few requests for additional comments / TODO's tickets before merge, but no need to post an update with those.

.github/workflows/release.yml
stuqdog: done!

.github/workflows/release.yml
stuqdog: Ahh that's a good point! this is reason enough to just make the change now imo, tested and it works.

src/viam/sdk/CMakeLists.txt
stuqdog: went ahead and switched to `if (TARGET viam_rust_utils)`, confirmed that it builds correctly both with a user-provided `rust-utils` file and without one.

PR #420, opened 2025-04-22
[RSDK-10412] add conan.lock

penguinland:
I was looking into RDSK-10413 for the tflite_cpu repo (see https://github.com/viam-modules/mlmodel-tflite/pull/28), and noticed this is basically the same ticket for a different repo.

This file was generated by running `conan lock create .`. I don't like that the file doesn't end with a newline, but I also don't like manually editing files created by machines.

I think pinning all versions of libraries we depend on is a useful feature, though I'm not good at modern C++ (let alone Conan itself), and might have missed something. I'm tagging @lia-viam, as our resident Conan expert, and @acmorrow as the ticket creator.


conan.lock
```diff
@@ -0,0 +1,64 @@
...
+        "gnu-config/cci.20210814#dc430d754f465e8c74463019672fb97b%1701248168.479",
+        "flex/2.6.4#e35bc44b3fcbcd661e0af0dc5b5b1ad4%1674818991.113",
+        "cmake/3.31.6#ed0e6c1d49bd564ce6fed1a19653b86d%1740594968.467",
+        "c-ares/1.34.3#55aa867c731f23300cd544f6d90a794d%1731428505.375",
+        "bzip2/1.0.8#d00dac990f08d991998d624be81a9526%1724661266.998",
+        "b2/5.2.1#91bc73931a0acb655947a81569ed8b80%1718416612.241",
+        "automake/1.16.5#058bda3e21c36c9aa8425daf3c1faf50%1688481772.751",
+        "autoconf/2.71#f9307992909d7fb3df459340f1932809%1711983104.648",
+        "abseil/20250127.0#7f3643d6f847196fe0bdc2b83c109831%1742212255.256",
+        "abseil/20240116.2#dfc978c35194fee939d236253044b967%1740151184.642"
```
acmorrow, line 60: This is very interesting, and is *maybe* concerning, though I can't quite tell what the implications for having two versions of `abseil` in `build_requires` are w.r.t. to the scary ODR consequences of having two abseils in the graph. Hopefully @lia-viam can provide further insights.

I note though that there are other duplications in this list, particularly `meson` and `pkgconf`, which do make more sense as things that you could have two of without issue since they are presumably only used at build time.

acmorrow (commented):
Having thought about this some more, I'm now doubting my self about how I wrote that ticket. For rust-y things at least, you want lockfiles on programs not libraries. Maybe what I was really thinking was that _tflite_cpu_ needed a lockfile so builds of that binary were reproducible?

lia-viam:
@acmorrow 
> Maybe what I was really thinking was that tflite_cpu needed a lockfile so builds of that binary were reproducible?

This is what I vaguely remember of our discussion as well. I'm not super familiar with lockfiles but tentatively going to say I'm opposed to doing it here, and for `tflite_cpu` we could maybe do it, or provide guidance for how to do it if you want a local, reproducible build, but not necessarily check it in to source control. 

Will try to research conan lockfiles more and report back

conan.lock
penguinland: Stuff in the `build_requires` section is only used in tests and compile time tools (e.g., the `cmake` and `pkgconf` packages). 

I've tried running `conan lock create . -vtrace`, but haven't yet been able to track down where these dependencies are coming from. but my guess is that we have two tools used at build time that depend on the two different versions.

The ones I'd be concerned about are up in the `requires` section, and they're all unique.

penguinland: Okay, `conan graph info` is more useful! `protobuf` depends on any version of `abseil` between 20230802.1 and 20250127.0, whereas literally everything else (including `grpc`, oddly) depends on the much narrower range 20240116.1 to 20240117.0. The issue is just that `protobuf` is being very permissive with which versions it accepts.

My hope is that `conan` is smart enough to realize that a single version can satisfy all these requirements, though I don't actually know.

penguinland:
Lia and I talked: she has convinced me that this PR should be closed but https://github.com/viam-modules/mlmodel-tflite/pull/28 should remain open.

The versioning specified in conanfile.py is good enough for most things, and the lock file is only for when you absolutely need the exact same bytes every time. When you're building a standalone executable, you want the build to succeed every time, so make a lockfile so it's always the same. but when you're building a library that will get linked into someone else's code, that someone else might want slightly different versions of the dependencies. As long as those different versions still match our conanfile.py, we should allow them. 

So, get rid of this PR, but keep the one for the tflite module.

PR #421, opened 2025-04-22
Fix sigabrt in module integration tests

lia-viam:
There was a sigabrt in recent module integration tests that revealed some UB where we sometimes accidentally double-free the memory managed by a `std::string`. Zero clue as to why logging brought this up but it is indeed something we'd want to fix anyway.

Contains some drive-by changes as well

### DIFF
--- a/src/viam/sdk/robot/client.cpp
+++ b/src/viam/sdk/robot/client.cpp
@@ -131,7 +131,7 @@ void RobotClient::close() {
     should_refresh_.store(false);
     threads_.clear();
     stop_all();
-    viam_channel_->close();
+    viam_channel_.close();
 }
 
 bool is_error_response(const grpc::Status& response) {
@@ -209,7 +209,7 @@ void RobotClient::refresh() {
         if (rs) {
             try {
                 const std::shared_ptr<Resource> rpc_client =
-                    rs->create_rpc_client(name.name(), channel_);
+                    rs->create_rpc_client(name.name(), viam_channel_.channel());
                 const Name name_({name.namespace_(), name.type(), name.subtype()}, "", name.name());
                 new_resources.emplace(name_, rpc_client);
             } catch (const std::exception& exc) {
@@ -248,11 +248,10 @@ void RobotClient::refresh_every() {
     }
 };
 
-RobotClient::RobotClient(std::shared_ptr<ViamChannel> channel)
-    : channel_(channel->channel()),
-      viam_channel_(std::move(channel)),
+RobotClient::RobotClient(ViamChannel channel)
+    : viam_channel_(std::move(channel)),
       should_close_channel_(false),
-      impl_(std::make_unique<impl>(RobotService::NewStub(channel_))) {}
+      impl_(std::make_unique<impl>(RobotService::NewStub(viam_channel_.channel()))) {}
 
 std::vector<Name> RobotClient::resource_names() const {
     const std::lock_guard<std::mutex> lock(lock_);
@@ -284,9 +283,9 @@ void RobotClient::log(const std::string& name,
     }
 }
 
-std::shared_ptr<RobotClient> RobotClient::with_channel(std::shared_ptr<ViamChannel> channel,
+std::shared_ptr<RobotClient> RobotClient::with_channel(ViamChannel channel,
                                                        const Options& options) {
-    std::shared_ptr<RobotClient> robot = std::make_shared<RobotClient>(std::move(channel));
+    auto robot = std::make_shared<RobotClient>(std::move(channel));
     robot->refresh_interval_ = options.refresh_interval();
     robot->should_refresh_ = (robot->refresh_interval_ > 0);
     if (robot->should_refresh_) {
@@ -305,8 +304,8 @@ std::shared_ptr<RobotClient> RobotClient::with_channel(std::shared_ptr<ViamChann
 std::shared_ptr<RobotClient> RobotClient::at_address(const std::string& address,
                                                      const Options& options) {
     const char* uri = address.c_str();
-    auto channel = ViamChannel::dial_initial(uri, options.dial_options());
-    std::shared_ptr<RobotClient> robot = RobotClient::with_channel(channel, options);
+    std::shared_ptr<RobotClient> robot =
+        RobotClient::with_channel(ViamChannel::dial_initial(uri, options.dial_options()), options);
     robot->should_close_channel_ = true;
 
     return robot;
@@ -315,10 +314,9 @@ std::shared_ptr<RobotClient> RobotClient::at_address(const std::string& address,
 std::shared_ptr<RobotClient> RobotClient::at_local_socket(const std::string& address,
                                                           const Options& options) {
     const std::string addr = "unix://" + address;
-    const std::shared_ptr<grpc::Channel> channel =
-        sdk::impl::create_viam_channel(addr, grpc::InsecureChannelCredentials());
-    std::shared_ptr<RobotClient> robot =
-        RobotClient::with_channel(std::make_shared<ViamChannel>(channel), options);
+    std::shared_ptr<RobotClient> robot = RobotClient::with_channel(
+        ViamChannel(sdk::impl::create_viam_channel(addr, grpc::InsecureChannelCredentials())),
+        options);
     robot->should_close_channel_ = true;
 
     return robot;
--- a/src/viam/sdk/robot/client.cpp
+++ b/src/viam/sdk/robot/client.cpp
@@ -120,18 +120,16 @@ RobotClient::~RobotClient() {
         try {
             this->close();
         } catch (const std::exception& e) {
-            VIAM_SDK_LOG(error) << "Received err while closing RobotClient: " << e.what();
+            std::cerr << "Received err while closing RobotClient: " << e.what() << "\n";
         } catch (...) {
-            VIAM_SDK_LOG(error) << "Received unknown err while closing RobotClient";
+            std::cerr << "Received unknown err while closing RobotClient\n";
         }
     }
 }
 
 void RobotClient::close() {
     should_refresh_.store(false);
-    for (const std::shared_ptr<std::thread>& t : threads_) {
-        t->~thread();
-    }
+    threads_.clear();
     stop_all();
     viam_channel_->close();
 }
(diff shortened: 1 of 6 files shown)


src/viam/sdk/rpc/dial.hpp
```diff
@@ -38,11 +41,27 @@ class ViamChannel {
...
+        RustDialData(const RustDialData&) = delete;
+        RustDialData(RustDialData&&) noexcept;
+
+        RustDialData& operator=(const RustDialData&) = delete;
+        RustDialData& operator=(RustDialData&&) noexcept;
+
+        ~RustDialData();
+
+        const char* path;
+        void* rust_runtime;
```
stuqdog, line 59: (q) is there some C++ reason I don't understand why we aren't using the `pimpl` idiom here? It seems like this is all private implementation details that we don't really need to tell users about at all, since they have no way of accessing the dial data anyway.

src/viam/sdk/rpc/dial.cpp
```diff
@@ -18,18 +18,33 @@
...
+ViamChannel::RustDialData::RustDialData(RustDialData&& other) noexcept
+    : path(std::exchange(other.path, nullptr)),
+      rust_runtime(std::exchange(other.rust_runtime, nullptr)) {}
+
+ViamChannel::RustDialData& ViamChannel::RustDialData::operator=(RustDialData&& other) noexcept {
+    path = std::exchange(other.path, nullptr);
+    rust_runtime = std::exchange(other.rust_runtime, nullptr);
+
+    return *this;
 }
```
stuqdog, line 33: (q) when/how do these get used? It's not clear to me that we do any assignment of a `RustDialData` object except when creating a new one in the `ViamChannel` constructor.

src/viam/sdk/rpc/dial.hpp
```diff
@@ -38,11 +41,27 @@ class ViamChannel {
 
     const std::shared_ptr<GrpcChannel>& channel() const;
 
+    void close();
+
    private:
+    struct RustDialData {
```
stuqdog, line 47: minor gotcha, I think we've typically used `snake_case` for structs.

stuqdog (commented):
A couple minor things and we'll want to figure out why tests are failing, otherwise lgtm!

src/viam/sdk/robot/client.cpp
```diff
@@ -129,9 +129,7 @@ RobotClient::~RobotClient() {
 
 void RobotClient::close() {
     should_refresh_.store(false);
-    for (const std::shared_ptr<std::thread>& t : threads_) {
-        t->~thread();
-    }
+    threads_.clear();
```
acmorrow, line 132: Woah. This was also a double destroy for all these `std::thread` objects, right? I'm sort of amazed this went unnoticed for so long. Are we doing ASAN / UBSAN builds anywhere in CI? If not, I think it might be time to do so.

src/viam/sdk/robot/client.cpp
```diff
@@ -292,13 +290,12 @@ std::shared_ptr<RobotClient> RobotClient::with_channel(std::shared_ptr<ViamChann
...
     if (robot->should_refresh_) {
-        const std::shared_ptr<std::thread> t =
-            std::make_shared<std::thread>(&RobotClient::refresh_every, robot);
+        auto t = std::thread(&RobotClient::refresh_every, robot);
         // TODO(RSDK-1743): this was leaking, confirm that adding thread catching in
         // close/destructor lets us shutdown gracefully. See also address sanitizer,
         // UB sanitizer
-        t->detach();
-        robot->threads_.push_back(t);
+        t.detach();
```
acmorrow, line 297: Well, this is why the double delete didn't really do anything harmful: the thread was detached. But, how do they terminate then?

I think if you call `~RobotClient` you could have real problems: `close` doesn't wait for the threads, and then the `should_refresh` variable will could get destroyed out from under.

I have a real distrust of detached threads for reasons exactly like this. Perhaps this review is an opportunity to do the necessary refactoring to bound the lifetime of the refresh threads so they are a subset of the lifetime of the `RobotClient`.

src/viam/sdk/robot/client.cpp
```diff
@@ -318,11 +315,10 @@ std::shared_ptr<RobotClient> RobotClient::at_address(const std::string& address,
 std::shared_ptr<RobotClient> RobotClient::at_local_socket(const std::string& address,
                                                           const Options& options) {
     const std::string addr = "unix://" + address;
-    const char* uri = addr.c_str();
     const std::shared_ptr<grpc::Channel> channel =
-        sdk::impl::create_viam_channel(uri, grpc::InsecureChannelCredentials());
-    auto viam_channel = std::make_shared<ViamChannel>(channel, address.c_str(), nullptr);
-    std::shared_ptr<RobotClient> robot = RobotClient::with_channel(viam_channel, options);
+        sdk::impl::create_viam_channel(addr, grpc::InsecureChannelCredentials());
+    std::shared_ptr<RobotClient> robot =
```
acmorrow, line 320: `auto robot = `

src/viam/sdk/rpc/dial.cpp
lia-viam: so I had mentioned to you wanting to change instances of `shared_ptr<ViamChannel>` to `ViamChannel` and for that we would want correctly operating special member functions, otherwise we could recreate anew the issue with double deletes or frees. even without that though it's generally a good idea to program defensively with resource-managing classes, see eg https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#c21-if-you-define-or-delete-any-copy-move-or-destructor-function-define-or-delete-them-all

src/viam/sdk/rpc/dial.hpp
lia-viam: 🐍

src/viam/sdk/rpc/dial.hpp
acmorrow: It does feel like an implementation detail that could be hidden. As written it will affect the ABI since is held "by size" inside the optional and therefore inside `ViamChannel`, so if it grew new fields it would be an ABI break.

lia-viam: nope this does seem like a good candidate for pimpl and i considered doing it that way, only didn't do it bc it wasn't that way already but this seems like a good candidate for a drive-by

src/viam/sdk/rpc/dial.cpp
```diff
@@ -162,7 +174,15 @@ std::shared_ptr<ViamChannel> ViamChannel::dial(const char* uri,
     const std::unique_ptr<viam::robot::v1::RobotService::Stub> st =
         viam::robot::v1::RobotService::NewStub(channel);
     return std::make_shared<ViamChannel>(channel, socket_path, ptr);
-};
+}
+
+const std::shared_ptr<grpc::Channel>& ViamChannel::channel() const {
```
acmorrow, line 179: Was this just unimplemented before? If so, maybe that means nobody was using it, and therefore it could be renamed to something like `grpc::Channel`? Or, simply deleted if we don't want gRPC types in our ABI?

src/viam/sdk/rpc/dial.hpp
```diff
@@ -38,11 +41,27 @@ class ViamChannel {
...
+        RustDialData(const char* path_, void* runtime);
```
acmorrow, line 48: Why the trailing `_` on `path_`?

src/viam/sdk/robot/client.cpp
lia-viam: our unit tests in CI run with santized builds for shared libraries, but our sanitizer seems to only be ubsan, not asan. i'm pretty surprised this hasn't been caught before either, and same with the other main issue being resolved in this PR which is that we had an instance of rust-utils `free_string` being called on a `c_str` managed by a `std::string`. perhaps a blessing in disguise that logging directed us to these!

src/viam/sdk/robot/client.cpp
lia-viam: I suppose our alternative would be having all the threads `join` in `close`? This is a little dicey for throwing-in-dtor reasons, but I think in practice it may be fairly common

src/viam/sdk/rpc/dial.hpp
lia-viam: It could just as well be `a_path` or similar, but I think ctor arg with the same name as member may cause a variable shadow warning

acmorrow: I'm fairly sure it doesn't actually

src/viam/sdk/robot/client.cpp
acmorrow: Yes, I think joining the threads in `close` is the right thing to do.

src/viam/sdk/robot/client.cpp
acmorrow: ASAN and UBSAN work together just fine so it might be pretty easy to light it up.

stuqdog: iirc turning on ASAN causes compilation to fail because it modifies the underlying memory shape but gRPC is not being compiled with ASAN, leading to ABI errors (I might have the details here somewhat wrong). Definitely fixable but it means compiling gRPC by hand in tests, which I believe is why we haven't done it thus far.

lia-viam: Realizing I encountered this too yesterday when trying to debug the original issue in this PR. As for conan, here is a discussion from 1.x but the answer is 'kinda sorta' https://github.com/conan-io/conan/issues/4754

src/viam/sdk/rpc/dial.cpp
lia-viam: So yes this was used, but I have taken this opportunity to remove the data member `shared_ptr<grpc::Channel>` in `RobotClient` which was redundant and replace all its uses with this method

src/viam/sdk/robot/client.hpp
```diff
@@ -87,10 +87,9 @@ class RobotClient {
     /// @param options Options for connecting and refreshing.
     /// Connects directly to a pre-existing channel. A robot created this way must be
     /// `close()`d manually.
-    static std::shared_ptr<RobotClient> with_channel(std::shared_ptr<ViamChannel> channel,
-                                                     const Options& options);
+    static std::shared_ptr<RobotClient> with_channel(ViamChannel channel, const Options& options);
 
-    RobotClient(std::shared_ptr<ViamChannel> channel);
+    RobotClient(ViamChannel channel);
```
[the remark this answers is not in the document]
stuqdog: I think this should probably be private, good callout!

stuqdog (approved):
lgtm!

src/viam/sdk/robot/client.hpp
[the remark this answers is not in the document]
lia-viam: totally agreed but this ends up breaking `make_shared` bc it needs a public ctor ... ugh...going to just make it explicit and public for now, although in the future it may be that we can un-share `RobotClient` as we did with `ViamChannel`, tbd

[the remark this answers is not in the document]
acmorrow: Oh right. That always gets me. You could always do the thing with a private dummy struct argument if you wanted to lock it down so it was still public but not callable from outside the class.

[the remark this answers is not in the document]
lia-viam: I think for now we can probably merge it as is if the builds pass so as to unblock this week's release, and I will futz around in another branch to see if we can clean this up further

PR #425, opened 2025-04-28
RSDK-10579: Robot client update

lia-viam:
This is an update/minor overhaul to robot client that fixes some of the arch issues (and hopefully crashes) that were showing up after the logging PR. Tbd if the `atomic<bool>` should be replaced with something else.

### DIFF
--- a/src/viam/sdk/common/client_helper.cpp
+++ b/src/viam/sdk/common/client_helper.cpp
@@ -23,7 +23,7 @@ bool isStatusCancelled(int status) noexcept {
     return status == ::grpc::StatusCode::CANCELLED;
 }
 
-void set_name(...) {}
+void set_name(...) {}  // NOLINT(cert-dcl50-cpp)
 
 }  // namespace client_helper_details
 
--- a/src/viam/sdk/common/client_helper.cpp
+++ b/src/viam/sdk/common/client_helper.cpp
@@ -23,6 +23,8 @@ bool isStatusCancelled(int status) noexcept {
     return status == ::grpc::StatusCode::CANCELLED;
 }
 
+void set_name(...) {}
+
 }  // namespace client_helper_details
 
 ClientContext::ClientContext() : wrapped_context_(std::make_unique<GrpcClientContext>()) {
--- a/src/viam/sdk/robot/client.cpp
+++ b/src/viam/sdk/robot/client.cpp
@@ -91,6 +91,11 @@ struct RobotClient::impl {
         }
     }
 
+    template <typename Method>
+    auto client_helper(Method m) {
+        return make_client_helper(this, *stub_, m);
+    }
+
     std::unique_ptr<RobotService::Stub> stub_;
 
     boost::shared_ptr<viam::sdk::impl::SinkType> log_sink;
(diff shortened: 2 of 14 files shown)


acmorrow:
Just a general question, but does this render #424 obsolete?

src/viam/sdk/robot/client.cpp
```diff
@@ -295,7 +275,7 @@ std::shared_ptr<RobotClient> RobotClient::with_channel(ViamChannel channel,
     robot->refresh_interval_ = options.refresh_interval();
     robot->should_refresh_ = (robot->refresh_interval_ > 0);
     if (robot->should_refresh_) {
-        robot->threads_.emplace_back(&RobotClient::refresh_every, robot);
+        robot->threads_.emplace_back(&RobotClient::refresh_every, robot.get());
```
acmorrow, line 278: Apologies if this ends up being a duplicate comment: I know I left a note here but github seems to have eaten it somehow.

I think this means that the RobotClient will no longer be rendered immortal by its background refresh threads. So, doesn't that mean that this PR would need to include the changes to switch to a condvar and notification in order to interrupt and join them? I think this goes to my question of how this PR relates to #424 .

lia-viam: So I think the key here is "**no longer** rendered immortal" -- I don't think that was ever the case even beforehand. I mean of course our old approach was UB, what with `~thread` getting called manually in `close`, but even without that, automatic destructors would be destructing `vector<thread>` before destructing the `RobotClient` itself, so all the `shared_ptr` would be gone by then

src/viam/sdk/robot/client.hpp
```diff
@@ -69,6 +76,10 @@ class RobotClient {
     ~RobotClient();
 
     void refresh();
+
+    /// @brief Disconnect this robot client from any robot to which it is connected.
```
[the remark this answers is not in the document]
lia-viam: yes definitely, going to add a bunch of docs to this PR

lia-viam:
also @acmorrow  re: #424 the answer is maybe, going to run some tests after the changes in this PR are polished/finalized

src/viam/sdk/robot/client.cpp
```diff
@@ -116,91 +133,60 @@ void RobotClient::connect_logging() {
...
     }
 }
 
 void RobotClient::close() {
     should_refresh_.store(false);
 
-    for (auto& thread : threads_) {
-        thread.join();
+    if (refresh_thread_.joinable()) {
+        refresh_thread_.join();
```
acmorrow, line 149: I think this `join` will need to block until the refresh thread awakens from `std::this_thread::sleep_for(refresh_interval_)` in `refresh_every`.

acmorrow (approved):
This LGTM if the intention is that this land in a state where tearing down the client will endure a wait until the refresh thread wakes again. I think I'd want to see that work ticketed and acted on sooner rather than later though.

Alterntively, this review could go another round and switch over to using a condvar to effect the sleep, and then the thread could be awakened by signaling the condvar.

src/viam/sdk/common/client_helper.cpp
```diff
@@ -23,6 +23,8 @@ bool isStatusCancelled(int status) noexcept {
     return status == ::grpc::StatusCode::CANCELLED;
 }
 
+void set_name(...) {}  // NOLINT(cert-dcl50-cpp)
```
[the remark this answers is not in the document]
lia-viam: exactly! a variadic is always considered the worst/last choice for overload resolution so it's needed in this case

src/viam/sdk/robot/client.cpp
lia-viam: correct, recording for posterity that we will try to swap with a proper cond var wait but merging now to fix the latest release

PR #430, opened 2025-05-02
RSDK-10435: Dial direct

lia-viam:
Enables dialing directly to gRPC, potentially circumventing rust-utils entirely.

This includes some QoL/drive-by changes, such as `struct`ifying `DialOptions` and removing the `shared_ptr` from `RobotClient`


src/viam/sdk/module/service.cpp
```diff
@@ -254,13 +254,12 @@ ModuleService::ModuleService(int argc,
...
 
 ModuleService::~ModuleService() {
-    // TODO(RSDK-5509): Run registered cleanup functions here.
-    VIAM_SDK_LOG(info) << "Shutting down gracefully.";
+    VIAM_SDK_LOG(info) << "Shutting down gracefully.\n";
     server_->shutdown();
 
     if (parent_) {
         try {
-            parent_->close();
```
lia-viam, line 263: this was causing a double call to `stop_all`, the second one happening after disconnecting from the robot

src/viam/sdk/rpc/dial.cpp
```diff
@@ -1,3 +1,4 @@
+#include <grpc/grpc.h>
```
stuqdog, line 1: (nit) should move this to a separate include block

src/viam/sdk/rpc/dial.cpp
```diff
@@ -6,12 +7,18 @@
...
 
+#include <viam/api/proto/rpc/v1/auth.grpc.pb.h>
+#include <viam/api/proto/rpc/v1/auth.pb.h>
 #include <viam/api/robot/v1/robot.grpc.pb.h>
 #include <viam/api/robot/v1/robot.pb.h>
+#include <viam/sdk/common/client_helper.hpp>
 #include <viam/sdk/common/exception.hpp>
+#include <viam/sdk/common/private/instance.hpp>
 #include <viam/sdk/rpc/private/viam_grpc_channel.hpp>
 #include <viam/sdk/rpc/private/viam_rust_utils.h>
```
stuqdog, line 23: (nit) Hm we did this wrong here (my bad!) but in most places in the SDK we've got the `api` includes in a separate block before the `sdk` includes. Could we do the same here?

src/viam/sdk/rpc/private/viam_grpc_channel.cpp
```diff
@@ -9,6 +9,14 @@ namespace viam {
...
 
+std::shared_ptr<grpc::Channel> create_viam_auth_channel(const std::string& address) {
+    grpc::experimental::TlsChannelCredentialsOptions opts;
+    opts.set_check_call_host(false);
+    auto tls_creds = grpc::experimental::TlsCredentials(opts);
+
+    return grpc::CreateChannel(address, tls_creds);
+}
+
 std::shared_ptr<grpc::Channel> create_viam_channel(
```
stuqdog, line 20: probably not worth fixing here, but it's unfortunate and a little confusing that `create_viam_channel` creates a `grpc::Channel`, not a `ViamChannel`. :(

src/viam/sdk/rpc/dial.cpp
```diff
@@ -186,6 +138,31 @@ ViamChannel ViamChannel::dial(const char* uri, const boost::optional<DialOptions
...
+    ClientContext ctx;
+    AuthenticateRequest req;
+
+    *req.mutable_entity() = opts.auth_entity.get();
+    *req.mutable_credentials()->mutable_payload() = opts.credentials->payload();
+    *req.mutable_credentials()->mutable_type() = opts.credentials->type();
+
+    AuthenticateResponse resp;
+
+    auto status = auth_stub->Authenticate(ctx, req, &resp);
```
stuqdog, line 155: Can we factor this out into a helper function? I think it's a bit cleaner, and also will be relevant if/when we want to have app client support in C++.

src/viam/sdk/rpc/dial.hpp
```diff
@@ -66,49 +69,32 @@ class Credentials {
...
+    boost::optional<Credentials> credentials;
 
     /// @brief Allows the RPC connection to be downgraded to an insecure connection if detected.
     /// This is only used when credentials are not present.
-    bool allow_insecure_downgrade_ = false;
+    bool allow_insecure_downgrade = false;
+
+    /// @brief Bypass WebRTC and connect directly to the robot.
+    /// This dials directly through grpc bypassing rust utils.
+    bool disable_webrtc = false;
```
stuqdog, line 85: (opt) Maybe not a big deal (I don't think we do this in other SDKs), but it might be good to add a note here about the specialized `uri` form that is required (`<part>.<location>.local.viam.cloud:8080`). I could definitely imagine that being a gotcha.

stuqdog (commented):
Nice! A couple small things but generally looks good to me.

src/viam/sdk/rpc/dial.hpp
lia-viam: Oh yeah I wanted to ask you about that, should we try to validate the URI under this code path and error out of it's not of that form?

src/viam/sdk/rpc/private/viam_grpc_channel.cpp
lia-viam: fixed it :)

src/viam/sdk/rpc/dial.hpp
stuqdog: Hmm good question. I think probably no. My thinking is:
1. that format might change in the future and that could lead to some issues that might be annoying to track down, and
2. that format is currently required assuming the user isn't doing anything weird or bespoke. If, e.g., they were to use a different app instance than `app.viam.com` then maybe the format would be different? I have no idea but I don't want to just break things for someone if such a case were to arise

On the other hand, it _is_ a little annoying if we have users saying "why doesn't this work??" and we say "oh it says in the comments" and they say "that wasn't helpful, if you'd just given me an error message I would've known much faster". So while I'm inclined _not_ to error out, I'm open to being convinced.

Maybe we add a `warning` log message? Probably excessive 😅

src/viam/sdk/rpc/dial.hpp
```diff
@@ -11,7 +11,8 @@
 namespace viam {
 namespace sdk {
 
-class DialOptions;
+struct DialOptions;
```
acmorrow, line 14: Are these really `ViamChannelOptions`? Or `ViamChannel::Options`? In which case, this header could be `viam_channel.h`?

src/viam/sdk/rpc/dial.cpp
```diff
@@ -157,22 +102,30 @@ ViamChannel ViamChannel::dial_initial(const char* uri,
...
-    void* ptr = init_rust_runtime();
     const DialOptions opts = options.get_value_or(DialOptions());
-    const std::chrono::duration<float> float_timeout = opts.timeout();
+
+    if (opts.disable_webrtc) {
+        return dial_direct(uri, opts);
+    }
+
+    const std::chrono::duration<float> float_timeout = opts.timeout;
+    void* ptr = init_rust_runtime();
```
acmorrow, line 112: If this was captured by `unique_ptr` with `free_rust_runtime` as the deleter type, some things could be simplified.

src/viam/sdk/common/client_helper.cpp
```diff
@@ -30,6 +31,12 @@ void set_name(...) {}  // NOLINT(cert-dcl50-cpp)
 ClientContext::ClientContext() : wrapped_context_(std::make_unique<GrpcClientContext>()) {
     set_client_ctx_authority_();
     add_viam_client_version_();
+
+    const std::string& token =
+        Instance::current(Instance::Creation::open_existing).impl_->direct_dial_token;
+    if (!token.empty()) {
```
acmorrow, line 37: I'm a little surprised to see this as a global, but maybe I'm misunderstanding. But if you had one channel that was doing direct dial and another that wasn't, wouldn't this by trying to add this metadata to the contexts for both channels?

src/viam/sdk/module/service.cpp
acmorrow: Does `conan` now give us a vehicle for building the entire dependency tree under ASAN and UBSAN?

acmorrow (commented):
I took a look out of interest. Please don't let these comments get in the way of moving this forward, they are just observations and notes on things that have bugged me about the code that is under modification here, so this seemed like a time to raise them. I'm happy for any or none of these issues to be deferred into future work.

src/viam/sdk/module/service.cpp
lia-viam: I think yes in theory but it's probably pretty tricky, see here for discussion (old conan docs) https://docs.conan.io/1/howtos/sanitizers.html

src/viam/sdk/common/client_helper.cpp
lia-viam: I think yes, although I'm not sure how common/feasible such a situation would be in practice, or if we should error out if we detect that kind of thing? cc @stuqdog

src/viam/sdk/rpc/dial.cpp
lia-viam: that's a good idea i'll try to do it that way

src/viam/sdk/rpc/dial.hpp
lia-viam: I think they probably are indeed `ViamChannel::Options`

src/viam/sdk/rpc/dial.hpp
```diff
@@ -66,49 +73,34 @@ class Credentials {
...
-    DialOptions& set_allow_insecure_downgrade(bool allow);
-    DialOptions& set_timeout(std::chrono::duration<float> timeout);
-    DialOptions& set_initial_connection_attempts(int attempts);
-    DialOptions& set_initial_connection_attempt_timeout(std::chrono::duration<float> timeout);
-
-   private:
-    // TODO (RSDK-917): We currently don't provide a flag for disabling webRTC, instead relying on a
-    // `local` uri. We should update dial logic to consider such a flag.
-
+struct DialOptions {
```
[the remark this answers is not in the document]
lia-viam: Ah so personally I was applying the C++ core guideline against trivial getters and setters in a class which hint that something probably should have just been a `struct`. With newer C++ (or older C++ with gnu extensions) I'm much more in favor of using designated initializers, eg

```
ViamChannel::dial(where, {
    .entity = ...,
    .credentials = ...,
});
```

this PR has exposed a bunch of weird stuff around these classes and `optional`, I was going to try to clean up the example code because there's a lot of unnecessary instances of constructing an `options` and then a `optional<options>` which is set to the previous `options`

src/viam/sdk/rpc/dial.hpp
```diff
@@ -66,49 +73,34 @@ class Credentials {
...
 
     /// @brief Number of attempts to make when initially connecting to a robot
     /// If set to 0 or a negative integer, will attempt to reconnect forever.
-    int initial_connection_attempts_ = 3;
+    int initial_connection_attempts = 3;
 
     /// @brief Timeout of connection attempts when initially dialing a robot
     /// Defaults to 20sec to match the default timeout duration
-    std::chrono::duration<float> initial_connection_attempt_timeout_{20};
+    std::chrono::duration<float> initial_connection_attempt_timeout{20};
```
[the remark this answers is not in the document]
lia-viam: I suspect a lot of this may be cruft that was not thought over a ton and hasn't been updated. I agree no refresh seems like an odd default

src/viam/sdk/common/client_helper.cpp
stuqdog: Hmm this is a good question. I'm not sure how common such a situation is in practice currently, but I do think it's certainly feasible and I could imagine a case where, e.g., someone is connecting to two machines at once and would prefer to use direct gRPC but is required to use webRTC for some of its added functionality on one of the machines only.

For that matter, even if both channels were dialing direct, if they were dialing to different machines then the token would only be valid for one of them! So yeah, probably tokens should either not be global or (if that's not feasible) should be keyed to a particular channel.

lia-viam: hmm ok so this is probably something we should handle better. i'll try to look into the architecture and see how to make this work probably on a per channel basis

src/viam/sdk/rpc/dial.hpp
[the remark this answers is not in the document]
lia-viam: doing an overhaul of this API (commit forthcoming) and made the troubling discovery that a _lot_ of our example code says that passing a refresh interval of zero seconds does auto refresh. I'm not sure what auto refresh means or if this was ever the case!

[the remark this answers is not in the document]
lia-viam: update: documentation has been updated to reflect that zero seconds means 'only on config update'

src/viam/sdk/common/client_helper.cpp
lia-viam: @acmorrow after much wailing and gnashing of teeth we figured out a way to do this without a global, thereby ballooning the surface area of this PR considerably.

acmorrow (commented):
Sorry for the long lag on this, I finally got time to sit down and go through it.

src/viam/sdk/rpc/dial.cpp
lia-viam: Life comes at you slow....

PR #438, opened 2025-05-21
conan: handle newer packages with c++17

lia-viam:
trying to fix windows release build failures


lia-viam:
this will need a proper fix, but i used this branch just to be able to build a release binary for windows

PR #439, opened 2025-05-21
[RSDK-10768] Use the expected name for single input tensors

acmorrow:


lia-viam (approved):

penguinland (approved):
LGTM. Thanks for the test!

lia-viam:
linter is complaining but otherwise lgtm

PR #442, opened 2025-05-28
fix detection of c++14

lia-viam:
turns out you are not allowed to string compare against `14` and `gnu14` because `gnu14` is not supported on windows


acmorrow (approved):
LGTM

PR #443, opened 2025-05-29
work around boost/1.88.0 issue on windows

lia-viam:


acmorrow (approved):
LGTM, and thanks for fixing.

PR #448, opened 2025-06-02
RSDK-10842 - shut down cpp modules properly when viam-server is hard killed

stuqdog:


src/viam/sdk/robot/client.cpp
```diff
@@ -231,6 +236,62 @@ void RobotClient::refresh_every() {
...
+                refresh();
+                connected = true;
+            } catch (const std::exception& e) {
+                viam_channel_.close();
+                std::this_thread::sleep_for(std::chrono::seconds{reconnect_every});
+            }
+        }
+        if (!connected) {
+            // NOLINTNEXTLINE
+            close();
```
stuqdog, line 289: @lia-viam here's somewhere that your C++ expertise would be greatly appreciated! `close` seemed better than just `exit`ing to ensure graceful shutdown (though we do ultimately want to `exit` if we've reached this point), but I'm not totally confident that calling `close` here, from within this thread, is safe behavior.

src/viam/sdk/robot/client.cpp
```diff
@@ -231,6 +236,62 @@ void RobotClient::refresh_every() {
...
+    unsigned int check_every = check_every_interval_;
+    unsigned int reconnect_every = reconnect_every_interval_;
+    if (check_every == 0) {
+        check_every = reconnect_every;
+    }
+    if (check_every == 0 && reconnect_every == 0) {
+        should_check_connection_.store(false);
+    }
+    bool connected(true);
+    while (should_check_connection_) {
```
stuqdog, line 249: @lia-viam another ignorant C++ question, `should_check_connection_` is set to false when we want to shutdown (makes sense!) but also above when `check_every` and `reconnect_every` are set to zero. In such a case this thread will return pretty much immediately, but the client code in such a case could be running for an indefinite amount of time. Are there any safety concerns here?

src/viam/sdk/rpc/dial.hpp
```diff
@@ -117,12 +120,32 @@ class Options {
...
+    /// check, will default to the `reconnect_every_interval_` value. Defaults to 0.
+    /// @note Setting to a non-zero value is useful in modules but may result in delays shutting
+    /// down client code
+    unsigned int check_every_interval_ = 0;
+
+    /// @brief How often to attempt to reconnect to the robot when disconnected. If set to 0,
+    /// will not attempt to reconnect. Defaults to 0.
+    /// @note Setting to a non-zero value is useful in modules but may result in delays shutting
+    /// down client code
+    unsigned int reconnect_every_interval_ = 0;
```
stuqdog, line 148: I opted to set these both to `0` by default. It leads to no meaningful behavioral change for existing clients, and sidesteps the issue of having to wait for the `check_connection` loop to finish before shutting down client code. Also if a client is having issues connecting, they probably can figure out what they want to do and stop the process as necessary by interacting directly, without us making any assumptions.

this is overridden for modules in `modules/service.cpp`, to ensure that modules don't get orphaned.

src/viam/sdk/robot/client.cpp
lia-viam: I think the issue calling `close` from here is that `joinable` will return false because it checks `thread.get_id() == this_thread::id()`. we should probably be storing false in the `should_check_connection_` thing if we establish that we've been disconnected, so that even if we're not calling `join` we can be sure that the thread will have exited

another thought would be to have a flag variable that says if we've been disconnected, and have `impl::client_helper` check that the same way it always checks if `impl` is non-null. That might exit with an exception throw, but I'm not sure if we're too pressed about "graceful" shutdown in the highly ungraceful case of the process being orphaned by a `kill -9`

lia-viam: also maybe cc @acmorrow

lia-viam (approved):

src/viam/sdk/robot/client.cpp
stuqdog: fwiw, the python SDK at this point just calls `sys.exit()`, so if we're not too concerned about gracefulness (and this is a case where the module is an orphan process anyway and nothing it logs or does will be trivially inspectable), we could just `exit(0)` and call it a day.

lia-viam: per offline discussion--this is actually fine bc when the `check_connection` thread calls `close` it will not `join` itself, but the later destructor call will

PR #449, opened 2025-06-03
[RSDK-10836] Use pimpl idiom to split signal_manager

gabegottlob:
testing


gabegottlob:
Successfully built on all platforms (see test commit 2). The final commit reverted the changes we made to the conan.yml (file in order to test whether it would successfully build on windows) so it is back to normal.

src/viam/sdk/tests/Testing/Temporary/CTestCostData.txt
lia-viam, line 1: it looks like these files were accidentally added to the commit, you should do a `git rm` on this `Testing` directory

src/viam/sdk/module/signal_manager.cpp
lia-viam, line 1: so this is gonna be a little annoying but i want to have you `git mv` this rather than deleting the file outright, that way we don't lose the git history on the file. my suggestion would be to undo this deletion with `git checkout main signal_manager.cpp`, verify that the git history is still present using `git log`, and then rename `signal_manager.cpp` to `signal_manager_posix.cpp` with `git mv

src/viam/sdk/module/signal_manager_posix.cpp
```diff
@@ -0,0 +1,29 @@
...
+
+SignalManager::~SignalManager() {}
+
+int SignalManager::wait() {
+    int sig = 0;
+    return sigwait(&((pImpl->sigset_)), &sig);
+}
+
+}  // namespace sdk
+}  // namespace viam
```
lia-viam, line 29: this is how git tells you you're missing a newline at the end of the file, try adding one manually or changing your vscode settings to automatically insert them

src/viam/sdk/module/signal_manager_posix.cpp
```diff
@@ -0,0 +1,29 @@
...
+};
+
+SignalManager::SignalManager() : pImpl(std::make_unique<Impl>()) {
+    sigemptyset(&(pImpl->sigset_));
+    sigaddset(&(pImpl->sigset_), SIGINT);
+    sigaddset(&(pImpl->sigset_), SIGTERM);
+    pthread_sigmask(SIG_BLOCK, &(pImpl->sigset_), NULL);
+}
+
+SignalManager::~SignalManager() {}
```
lia-viam, line 21:
```suggestion
SignalManager::~SignalManager = default;
```
try looking up this syntax if you're not familiar with it!

src/viam/sdk/module/signal_manager.hpp
```diff
@@ -1,10 +1,6 @@
 #pragma once
 
-#include <signal.h>
-
-#ifdef _WIN32
-#include <windows.h>
-#endif
+#include <memory>  //must include for std::unique_ptr
```
lia-viam, line 3: i think it's safe to omit this comment and the ones below :)

lia-viam (changes requested):
good work overall! here are some changes for this round of review

src/viam/sdk/module/signal_manager_posix.cpp
```diff
@@ -0,0 +1,29 @@
+#include <viam/sdk/module/signal_manager.hpp>
+
+#include <csignal>
+#include <pthread.h>
+#include <signal.h>
```
acmorrow, line 5: Put these two system headers into their own group, before including `<csignal>`:

```
#include <pthread.h>
#include <signal.h>

#include <csignal>
```

Also, why do you need both `<csignal>` and `signal.h`?

gabegottlob: The original signal_manager.hpp included signal.h, and the cpp file included csignal. Was there a reason there is both in the first place?

src/viam/sdk/module/signal_manager.cpp
lia-viam: turns out this is not possible to do 🤯

stuqdog (approved):

PR #454, opened 2025-06-18
[RSDK-10923] Add IsHoldingSomething to Gripper API

gvaradarajan:


src/viam/sdk/tests/mocks/mock_gripper.cpp
```diff
@@ -28,6 +28,13 @@ bool MockGripper::is_moving() {
     return false;
 }
 
+Gripper::holding_status MockGripper::is_holding_something(const ProtoStruct&) {
+    Gripper::holding_status res;
+    res.is_holding_something = false;
+    res.meta = {};
+    return res;
+}
+
```
stuqdog, line 37: can we set `res.meta` to the protoStruct that's passed into the function, and then write a test of this method in `test_gripper`?

stuqdog (commented):
Request to add a test and also fix compilation failures, otherwise lgtm!

acmorrow (approved):
LGTM

PR #458, opened 2025-06-27
RSDK-11067: remove resource on alwaysrebuild

lia-viam:


src/viam/sdk/module/service.cpp
```diff
@@ -128,16 +128,14 @@ struct ModuleService::ServiceImpl : viam::module::v1::ModuleService::Service {
...
-            Registry::get().lookup_model(cfg.name());
+            Registry::get().lookup_model(cfg.api(), cfg.model());
 
         // TODO RSDK-11067 new resource gets constructed while old one is still alive.
         if (reg) {
             try {
-                const std::shared_ptr<Resource> resource = reg->construct_resource(deps, cfg);
-                manager->replace_one(cfg.resource_name(), resource);
+                std::shared_ptr<Resource> resource = reg->construct_resource(deps, cfg);
+                manager->replace_one(cfg.resource_name(), std::move(resource));
```
acmorrow, line 137: - Arguably, this is part of RSDK-11067, but the behavior here means that if `construct_resource` fails, the old one sticks around. Is that OK?
- Maybe add some warn logging here after the call to `replace_one` that if `res.use_count != 1` that the old `Resource` has been held alive somewhere? I know it isn't perfect due to multithreading, but at least it gives some indication that things haven't gone according to plan.

acmorrow (approved):
LGTM, but please file a new ticket about investigating reconfiguration behavior and CC me on it. I'll write up my observations there.

PR #463, opened 2025-07-11
RSDK-11289: Fix logic in replace_one

lia-viam:
third time's the charm? this should finally fix the  _`always-rebuild`_ logic in `ReconfigureResource`, making sure that the old resource is destructed before the new one is created. this was causing issues with realsense and would cause issues for any other unique or global resource which has sole ownership of a device, socket, etc.

this adds an overload of `replace_one` which takes a construction callback, bringing the logic closer to that of eg python
https://github.com/viamrobotics/viam-python-sdk/blob/main/src/viam/module/module.py#L253

i marked the old one as deprecated for now but we may just want to remove it entirely 

cc @hexbabe @acmorrow

### DIFF
--- a/src/viam/sdk/resource/resource_manager.hpp
+++ b/src/viam/sdk/resource/resource_manager.hpp
@@ -58,7 +58,14 @@ class ResourceManager {
     /// @brief Replaces an existing resource. No-op if the named resource does not exist.
     /// @param name The name of the resource to replace.
     /// @param resource The new resource that is replacing the existing one.
+    [[deprecated("Use the callback overload as it destroys before constructing")]]
     void replace_one(const Name& name, std::shared_ptr<Resource> resource);
+
+    /// @brief Replaces an existing resource. No-op if the named resource does not exist.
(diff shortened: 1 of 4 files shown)


src/viam/sdk/resource/resource_manager.hpp
```diff
@@ -58,7 +58,15 @@ class ResourceManager {
     /// @brief Replaces an existing resource. No-op if the named resource does not exist.
     /// @param name The name of the resource to replace.
     /// @param resource The new resource that is replacing the existing one.
-    void replace_one(const Name& name, std::shared_ptr<Resource> resource);
+    [[deprecated("Use the callback overload as it destroys before constructing")]] void replace_one(
```
stuqdog, line 61: I think if this overload doesn't work correctly we should just get rid of it entirely.

stuqdog (approved):
lgtm though I think we should probably just get rid of the deprecated `replace_one` if it doesn't work.

PR #479, opened 2025-08-25
RSDK-1150: Mlmodel data type logging

lia-viam:
Addresses the ticket by adding an ostream operator which is used in the error stringstreams rather than casting to `uint8_t`. Note that the to_string conversion returns `nullptr` on error, so we cast to a numeric value in this case


src/viam/sdk/services/mlmodel.cpp
```diff
@@ -100,6 +100,20 @@ const char* MLModelService::tensor_info::data_type_to_string(const data_types da
...
 
+std::ostream& operator<<(std::ostream& os, MLModelService::tensor_info::data_types dt) {
+    const char* str = MLModelService::tensor_info::data_type_to_string(dt);
+
+    if (str) {
+        os << str;
+    } else {
+        // Cast to unsigned because uint8_t is unsigned char, and 0-9 are whitespace or non printing
+        // characters
+        os << static_cast<unsigned>(static_cast<std::underlying_type_t<decltype(dt)>>(dt));
```
penguinland, line 111: Naive question: why did you need to `static_cast` this twice? Couldn't you just cast straight to `unsigned` and be done?

penguinland (approved):
LGTM!

I'm glad you wrote this instead of me: this is much more elegant than what I would have done.

PR #485, opened 2025-09-25
Draft: RSDK-11274 Strings in motion service

lia-viam:
Draft PR for getting rid of `ResourceName` in motion service; holding off until we fix the proto update job


lia-viam:
this was merged into the latest proto update PR

PR #488, opened 2025-10-01
use protobuf/<host_version> in build requirements

lia-viam:
see https://docs.conan.io/2/reference/conanfile/methods/build_requirements.html#host-version


acmorrow:
LGTM!

PR #539, opened 2026-01-20
Fix conan build breakage from buf v2 upgrade

lia-viam:
See https://github.com/viamrobotics/viam-cpp-sdk/actions/runs/21167290773/job/60875067417


stuqdog (approved):
if it fixes issues, lgtm!

PR #608, opened 2026-04-02
RSDK-13622: Add lockfile to corresponding release

lia-viam:
Fixes the use and publishing of lockfiles so we do not have a situation where the lockfile is out of step with its corresponding release. Now
- the lockfile in `etc/conan/conan.lock` does not mention `viam-cpp-sdk`
- the publish conan packages job has an option to generate a lockfile and add it to a corresponding release
- release lockfiles _will_ contain `viam-cpp-sdk`, with the version and package revision pinned to the release in which the lockfile artifact is contained

This lets us distinguish the lockfile used to _build_ the C++ SDK from lockfiles used to _build modules_ which use the C++ SDK.


lia-viam:
see https://github.com/viamrobotics/viam-cpp-sdk/releases/tag/releases%2Fv0.33.1 which had a conan.lock added after the fact

stuqdog (approved):

PR #626, opened 2026-05-11
RSDK-13854: OpenTelemetry support

lia-viam:
Basic support for OpenTelemetry in viam-cpp-sdk. This hooks opentel into client/server API boundaries, so that, eg, a `Sensor.GetReadings` request will generate a span in RDK with a subspan showing how much time was spent just in viam-cpp-sdk.

The architecture is similar to the log backend and logging through RobotClient: there is an impl class that lives in the `Instance`, some setup logic on startup, and a RobotClient method which sends RPCs to RDK. 

Config options should be documented in the build code, but basically this is off by default and effectively a conan exclusive, because system package managers have spotty support for opentel. Note that in addition to the compile-time build switch, there is also a runtime env var check


src/viam/api/CMakeLists.txt
```diff
@@ -499,6 +493,26 @@ target_link_libraries(viamapi
   PRIVATE Threads::Threads
 )
 
+# `robot.proto` imports `opentelemetry/proto/trace/v1/trace.proto`, so the
+# OTel proto descriptors are part of viamapi's ABI. When tracing is enabled,
+# get them from the system `opentelemetry-cpp::proto` library to avoid
+# double-registering them alongside `opentelemetry-cpp::otlp_grpc_exporter`
+# (which links the same library transitively into viamsdk). When tracing is
+# disabled, compile our locally-generated copies into viamapi instead.
```
acmorrow, line 509: Are there version skew concerns here? Or does it all just work out because if we are building with otel then we don't care what buf did anyway?

src/viam/sdk/tracing/private/span_guard.hpp
```diff
@@ -0,0 +1,48 @@
...
+
+    /// @brief Record the final gRPC status before destruction and return it unchanged.
+    ::grpc::Status commit(::grpc::Status status) noexcept;
+
+    ServerSpanGuard(const ServerSpanGuard&) = delete;
+    ServerSpanGuard& operator=(const ServerSpanGuard&) = delete;
+
+   private:
+    struct Impl;
+    std::unique_ptr<Impl> impl_;
```
acmorrow: If this is a `private` type, does it need the pimpl?

src/viam/sdk/tracing/private/span_guard.cpp
```diff
@@ -0,0 +1,162 @@
+#include <viam/sdk/tracing/private/span_guard.hpp>
+
+#include <grpcpp/client_context.h>
+#include <grpcpp/server_context.h>
+
+#ifdef VIAMCPPSDK_OPENTELEMETRY_TRACING
```
acmorrow, line 6: Seems like this should span the whole file?

src/viam/sdk/tracing/private/span_guard.cpp
```diff
@@ -0,0 +1,162 @@
...
+        impl_->span->SetStatus(otel_trace::StatusCode::kError);
+        impl_->span->SetAttribute("rpc.grpc.status_code", static_cast<int>(status.error_code()));
+    }
+    return status;
+}
+
+void inject_trace_context(GrpcClientContext* ctx) noexcept {
+    GrpcClientCarrier carrier{ctx};
+    otel_prop::GlobalTextMapPropagator::GetGlobalPropagator()->Inject(
+        carrier, otel_ctx::RuntimeContext::GetCurrent());
```
acmorrow, line 120: What's the lifetime of carrier? Does it copy it?

src/viam/sdk/tracing/private/span_guard.cpp
```diff
@@ -0,0 +1,162 @@
...
+}
+
+}  // namespace impl
+}  // namespace sdk
+}  // namespace viam
+
+#else  // VIAMCPPSDK_OPENTELEMETRY_TRACING — provide no-op implementations
```
acmorrow, line 127: Oh, I see. Carry on then.

src/viam/sdk/tracing/private/tracer.hpp
```diff
@@ -0,0 +1,40 @@
...
+
+namespace viam {
+namespace sdk {
+
+class RobotClient;
+
+namespace impl {
+
+/// @brief Install the W3C Trace Context propagator. Called once during @c Instance construction.
+void initialize_trace_propagator() noexcept;
```
acmorrow: Could be static on `Tracer` maybe?

src/viam/sdk/tracing/private/tracer.hpp
```diff
@@ -0,0 +1,40 @@
...
+    static Tracer& get();
+
+    /// @brief Install an exporter sourced from @p client. The caller must keep @p client alive
+    /// until @c shutdown_provider returns.
+    void initialize_provider(RobotClient* client) noexcept;
+    void shutdown_provider() noexcept;
+
+   private:
+    struct Impl;
+    std::unique_ptr<Impl> impl_;
```
acmorrow: Ditto why pimpl if it private?

acmorrow (approved):
I went through this pretty fast: I'm not familiar with the otel APIs so doing a careful read there probably isn't a good use of time.

Overall, this looks great. I have some comments here and there but they are all small things or questions.

Basically, if this is working, I'd say you can ship it more or less as is.

src/viam/sdk/tracing/private/span_guard.hpp
```diff
@@ -0,0 +1,48 @@
...
+};
+
+/// @brief Inject the currently-active OpenTelemetry trace context as W3C @c traceparent /
+/// @c tracestate metadata entries into an outgoing gRPC client context.
+///
+/// This propagates trace context from any server span that is currently active on the calling
+/// thread (e.g., one created by @c ServerSpanGuard) to downstream gRPC calls. It is a no-op
+/// when no span is active or when OpenTelemetry tracing is not compiled in.
+void inject_trace_context(GrpcClientContext* ctx) noexcept;
```
acmorrow, line 54: This looks a little misplaced. It isn't really part of the span guard stuff. Is there a more specific home for it?

src/viam/api/CMakeLists.txt
lia-viam: I think the latter

src/viam/sdk/tracing/private/span_guard.cpp
lia-viam: good catch, this may be a lifetime issue will investigate further

src/viam/sdk/tracing/private/span_guard.hpp
lia-viam: good question. it's possible to do it without pimpl, but just makes it so all the `private` data of the class is in an `#ifdef`. which I think is fine, we probably don't care about ABI for private classes anyway

src/viam/sdk/tracing/private/span_guard.cpp
lia-viam: Ah so `GrpcClientCarrier` is defined above as basically an interface layer on top of the ctx to allow setting some underlying fields. It's not copied into Inject but doesn't need to be because all you can do through the interface is add metadata to the stored ctx.

PR #629, opened 2026-05-11
RSDK-13547- dh params from urdf

nfranczak:
I've added a `latex` subfolder which contains the `.tex` file and the `.pdf` file which guide the implementation of this code as well as explain the math.

Here is a very good video (short and easy to understand) to gain intuition on what: a, d, \alpha, \theta are:
https://www.youtube.com/watch?v=rA9tm0gTln8

Should it be chosen to review the mathy code here, this is the order of functions I'd recommend:
1. urdf_to_dh_params
2. parse_urdf
3. walk_urdf_chain
4. joint_axes_at_rest
5. build_dh_frames
6. pick_base_x
7. common_normal
8. validate_end_effector_dh
9. extract_dh_row

I've also added plenty of comments to show how certain parts of the code tie into the underlying mathematical document


lia-viam:
I think let's omit the latex/pdf, if you want to host it elsewhere maybe it could be in a doc comment or something

Other than that my question is more, what is this doing and does it belong in viam-cpp-sdk? Without having a super good grasp of what's going on, this seems a little more domain specific than what we usually include here. I'm also wondering if we need to add a dependency on eigen or if we can't get away with using xtensor or some of the existing AxisAngles or Vector3 classes

nfranczak:
> I think let's omit the latex/pdf...

I feel comfortable removing the pdf. Can you please clarify if the tex file can stay?

> what is this doing and does it belong in viam-cpp-sdk?

**What it does:** Converts a URDF (one way of describing arm kinematics) into DH parameters (more compact representation). 

**Why here:** Planned extensions to [trajex](https://github.com/viam-modules/trajex) consume DH parameters, so we need a clean URDF to DH path to feed the module.
Two reasons this belongs in the SDK rather than a module:
1. Trajex is itself a module. Having one module depend on another to get its inputs obfuscates its usage.
2. Putting it in the SDK makes it directly reusable by C++ arm drivers (universal-robots, yaskawa) and by any future C++ motion planning code we write.


> I'm also wondering if we need to add a dependency on eigen or if we can't get away with using xtensor or some of the existing AxisAngles or Vector3 classes

I can look into this.

lia-viam:
ok that makes sense. if we want to keep the tex file i think it probably belongs in the doc folder rather than the sdk tree directly

nfranczak:
Opting to do model table instead.
Will place another PR on this repo.

PR #635, opened 2026-05-13
RSDK-13961: Default to static conan builds

lia-viam:
Tagging reviews for some people who work on modules that consume viam-cpp-sdk, hopefully this should make your build code and readmes simpler! 

This PR was claude-generated with the following prompt: 
> default_options.shared has been changed to False, make changes to all conan code that was manually passing shared=False for viam-cpp-sdk, especially in .github/workflows

The same can probably be done in your own repos to simplify all the documentation and build script logic to set shared=False.


seanavery (approved):
Thanks for the heads up!

Will try removing the [explicit shared option setter](https://github.com/viam-modules/viam-camera-realsense/blob/ce7ee03b8d9f0ac2ba860b7dc77416a239f4dfc9/conanfile.py#L24) after bumping the sdk

acmorrow (approved):
LGTM.

PR #637, opened 2026-05-14
RSDK-13974 Implement play stream

oliviamiller:
Implement PlayStream RPC in audioout.


src/viam/sdk/components/private/audio_out_client.cpp
```diff
@@ -28,6 +29,46 @@ AudioOutClient::AudioOutClient(std::string name, const ViamChannel& channel)
           std::static_pointer_cast<::grpc::ChannelInterface>(channel.channel()))),
       channel_(&channel) {}
 
+void AudioOutClient::play_stream(
+    audio_info info,
+    std::function<boost::optional<std::vector<uint8_t>>()> chunk_source,
+    const ProtoStruct& extra) {
+    ::grpc::ClientContext ctx;
+    ::viam::component::audioout::v1::PlayStreamResponse response;
+    auto writer = stub_->PlayStream(&ctx, &response);
```
acmorrow, line 38: Just driving by, but, I was curious why `AudioClient::play_stream` doesn't use the client helper, and why `AudioOutServer::PlayStream` doesn't use the service helper.

If there is a use-case they don't support, we should be looking into how to extend support, not bypass them. Going forward, use of the client/service helpers is what will drive automatic open telemetry support integration in this project. As written `play_stream` / `PlayStream` will not be traceable.

oliviamiller: the current helpers don't work with client-side streaming RPCs, I'll look into extending the helpers to support this in this PR.

acmorrow: @oliviamiller - I thought that was added here: https://github.com/viamrobotics/viam-cpp-sdk/commit/669b48bbcd6372ae5f138fcc03a7aae1a27da199

oliviamiller: that added server-side streaming support

acmorrow: I see: he linked commit changed the client helper in furtherance of _server side_ streaming (to the client), but what is required for play stream is the opposite direction: we need the ability for a client to stream to the server. Maybe the thing to do is continue with what you have here but queue up the work to extend the helpers to cover this case.

acmorrow: @lia-viam - Is there anything on the roadmap for adding support to the helpers for this streaming mode?

lia-viam: last time we considered this problem there weren't enough streaming methods to warrant it. i think that is no longer the case and it would be good to have. note in particular that manually implemented streaming methods will need to follow the pattern in [board_server](https://github.com/viamrobotics/viam-cpp-sdk/commit/f0a1731680d324c8f8ac11b8b4eb30eb609c4d54#diff-155a9222edcd35d68e5477f430157822498e49326fb50f097a7bc421eb48f5ea) to have spans created, which is not ideal going forward

acmorrow: I filed https://viam.atlassian.net/browse/RSDK-13989 so we don't lose track of the need for this.

src/viam/sdk/module/service.cpp
```diff
@@ -215,7 +215,7 @@ struct ModuleService::ServiceImpl : viam::module::v1::ModuleService::Service {
         *response->mutable_handlermap() = hm;
 
         if (!is_env_var_true("VIAM_NO_MODULE_PARENT")) {
-            auto new_parent_addr = parent.grpc_conn_protocol_ + request->parent_address();
+            auto new_parent_addr = parent.grpc_conn_protocol_ + request->raw_parent_address();
```
oliviamiller, line 218: had to add this to bump API version due to https://github.com/viamrobotics/api/pull/837/changes

src/viam/sdk/components/private/audio_out_client.cpp
oliviamiller: Thank you! @lia-viam I added the spans to playstream the same way its implemented in board server for now.

lia-viam (approved):

PR #638, opened 2026-05-14
RSDK-13855: Span creation

lia-viam:
This was branched off of opentel-prototype, but as compared to that basically the only relevant changes are the public `tracing/span.xpp` files, and the two-line addition to the example code to show module author subspan creation. 

The example is a bit trivial and there's no unit testing, I would have to think about it but I'm not sure if there's a good way to meaningfully unit test this stuff given that it relies on there being a module parent process creating spans. It might be nice to have some of our test/example code in CI compiled with tracing enabled, but that would mean we'd have to use conan to build the module tests that run against viam-server.


src/viam/examples/modules/simple/main.cpp
```diff
@@ -71,6 +72,11 @@ ProtoStruct MySensor::do_command(const ProtoStruct& command) {
 }
 
 ProtoStruct MySensor::get_readings(const ProtoStruct&) {
+    // Example of simple instrumentation with opentelemetry.
+    // These are no-ops if tracing is not compiled in.
+    Span s("readings");
```
acmorrow: In C++20 `span` is an important concept. I think maybe it is worth disambiguating the name more. Maybe `TracingSpan` or `[viam::sdk::]tracing::Span`? Something.

src/viam/sdk/tracing/span.hpp
```diff
@@ -0,0 +1,74 @@
...
+
+   private:
+    struct Impl;
+    std::unique_ptr<Impl> impl_;
+};
+
+template <>
+void Span::set_attribute<std::string>(const char*, std::string) noexcept;
+
+// -- Explicit instantiation declarations for the supported attribute value types -- //
```
acmorrow, line 81: Nice

src/viam/sdk/tracing/span.hpp
```diff
@@ -0,0 +1,74 @@
...
+   public:
+    /// @brief Open a child span named @p name under the currently-active span.
+    explicit Span(const char* name) noexcept;
+
+    ~Span() noexcept;
+
+    Span(const Span&) = delete;
+    Span& operator=(const Span&) = delete;
+    Span(Span&&) = delete;
+    Span& operator=(Span&&) = delete;
```
acmorrow: The copy deletion should be automatic due to the unique_ptr member. What is the argument for disallowing move?

src/viam/sdk/tracing/span.hpp
```diff
@@ -0,0 +1,74 @@
...
+    void set_attribute(const char* key, T value) noexcept;
+
+    /// @brief Record a timestamped event on the span.
+    void add_event(const char* name) noexcept;
+
+    /// @brief Mark the span as having completed successfully.
+    void set_status_ok() noexcept;
+
+    /// @brief Mark the span as having failed. @p description is optional context.
+    void set_status_error(const char* description = "") noexcept;
```
acmorrow, line 57: Am I obligated to call one of these? Or is `status_ok` implied? How should this work with exceptions? I'm worried there will suddenly be tons of logic like:

```
Span x;
try {
    thing();
    x.set_status_ok();
catch(...) {
    x.set_status_error(std::current_exception().what());
    throw;
}
```

If possible I'd prefer for most of that to get automated away, so that it can be written like:

```
Span x;
thing();
```

And do the right thing almost all the time.

src/viam/sdk/tracing/span.cpp
```diff
@@ -0,0 +1,115 @@
...
+
+namespace otel_ctx = opentelemetry::context;
+namespace otel_trace = opentelemetry::trace;
+
+namespace viam {
+namespace sdk {
+
+namespace {
+
+constexpr const char* k_instrumentation_scope = "viam-cpp-sdk.module";
```
acmorrow: Small preference for declaring these as `constexpr const char k_instrumentation_scope[] = ...`.

acmorrow (changes requested):

src/viam/sdk/tracing/span.hpp
lia-viam: Good point about the copies. I think the moves are to enforce the way that we detect and associate spans with the active parent span on the same thread, so as to not have people moving stuff around in a way that breaks that association

src/viam/sdk/tracing/span.hpp
lia-viam: I think the latter case is indeed what happens but I will add a comment to that effect
https://opentelemetry.io/docs/concepts/signals/traces/#span-status

src/viam/sdk/tracing/span.hpp
```diff
@@ -0,0 +1,74 @@
...
+/// @ingroup Tracing
+class TracingSpan {
+   public:
+    /// @brief Open a child span named @p name under the currently-active span.
+    explicit TracingSpan(const char* name) noexcept;
+
+    ~TracingSpan() noexcept;
+
+    TracingSpan(TracingSpan&&) = delete;
+    TracingSpan& operator=(TracingSpan&&) = delete;
```
acmorrow, line 40: It is imperfect, but, as a hint, you could also delete the various forms of `new` here, making it much harder to actually misuse TracingSpan. That'd somewhat help the LIFO order requirement.

src/viam/sdk/tracing/span.hpp
acmorrow: That doesn't handle the case of wanting to mark a span as having completed with an error though. I believe you would still need the explicit `try` / `catch` machinery in order to do that.

src/viam/examples/modules/simple/main.cpp
```diff
@@ -71,6 +72,11 @@ ProtoStruct MySensor::do_command(const ProtoStruct& command) {
 }
 
 ProtoStruct MySensor::get_readings(const ProtoStruct&) {
+    // Example of simple instrumentation with opentelemetry.
+    // These are no-ops if tracing is not compiled in.
+    TracingSpan tspan("readings");
```
acmorrow: Should probably be `MySensor::get_readings` for the string.

acmorrow (approved):
LGTM.

My only quibble with it is the issue about forcing try/catch on all code that uses spans.

Maybe some machinery that let you write:

```
TracingSpan("MySensor::get_readings").wrap([&](auto& tspan) {
    tspan.add_event("computing signal");
});
```

And wrapped the whole thing in a try/catch block and set the appropriate outcome for you?

PR #639, opened 2026-05-18
feat: add vision service

HipsterBrown:
I was testing out the C++ module generation for the vision service module I wanted to build and discovered that service template was missing due to the API not being provided by the SDK. So I worked with Claude to implement that service API in the SDK and reuse as much of the existing types (like those from MLModel service and Camera component) as reasonable.


src/viam/sdk/services/vision.hpp
```diff
@@ -0,0 +1,211 @@
...
+    };
+
+    /// @struct point_cloud_object
+    /// @brief A point cloud and its associated geometry.
+    /// @note The proto field is named `point_cloud`; we use `cloud` here to avoid shadowing
+    /// the `point_cloud` type alias (GCC rejects the shadowing under -fpermissive).
+    struct point_cloud_object {
+        point_cloud cloud;
+        std::vector<GeometryConfig> geometries;
+    };
```
HipsterBrown, line 74: This might be a point of contention. Do we want to use `cloud` for the field name to avoid shadowing the type or something else?

src/viam/sdk/common/private/raw_image.hpp
```diff
@@ -0,0 +1,21 @@
...
+
+#include <viam/api/component/camera/v1/camera.pb.h>
+#include <viam/sdk/components/camera.hpp>
+
+namespace viam {
+namespace sdk {
+namespace impl {
+
+/// @brief Convert a proto Image to Camera::raw_image.
+Camera::raw_image from_proto(const ::viam::component::camera::v1::Image& proto);
```
lia-viam, line 14: I think these conversions should probably be public in `camera.hpp`, see summary comment

src/viam/sdk/services/private/vision.hpp
```diff
@@ -0,0 +1,41 @@
...
+#include <viam/api/common/v1/common.pb.h>
+#include <viam/api/service/vision/v1/vision.pb.h>
+#include <viam/sdk/services/vision.hpp>
+
+namespace viam {
+namespace sdk {
+namespace impl {
+namespace vision {
+
+::viam::service::vision::v1::Detection to_proto(const Vision::detection&);
```
lia-viam, line 26: similarly detection, classification, and point_cloud_object should be public i think

src/viam/sdk/services/private/vision_client.hpp
```diff
@@ -0,0 +1,72 @@
...
+    using interface_type = Vision;
+    using service_type = ::viam::service::vision::v1::VisionService;
+
+    VisionClient(std::string name, const ViamChannel& channel);
+
+    const ViamChannel& channel() const {
+        return *channel_;
+    }
+
+    std::vector<Vision::detection> get_detections_from_camera(const std::string& camera_name,
```
lia-viam, line 37: nit but can you put a newline in between the end of the last function definition and the subsequent ones

src/viam/sdk/services/vision.hpp
```diff
@@ -0,0 +1,211 @@
...
+///
+/// This acts as an abstract base class to be inherited from by any drivers representing
+/// specific vision implementations. This class cannot be used on its own.
+class Vision : public Service {
+   public:
+    API api() const override;
+
+    /// @struct detection
+    /// @brief Result of a single object detection.
+    /// Pixel bbox fields are optional int64; normalized bbox fields are optional double.
```
lia-viam, line 30: This comment might make more sense split into the body of the struct
```suggestion
    struct detection {
    // pixel bbox fields
    ...
    
    // normalized bbox fields
    ...
    };
```

src/viam/sdk/services/vision.hpp
lia-viam: pt_cloud would work, and I'd say remove the `@note` comment

src/viam/sdk/services/private/vision.hpp
```diff
@@ -0,0 +1,41 @@
+// Copyright 2024 Viam Inc.
```
lia-viam, line 1: a separate file for these conversions isn't really in line with other patterns we use in the sdk, but see comments below regarding the actual conversion functions

src/viam/sdk/tests/test_vision.cpp
```diff
@@ -0,0 +1,467 @@
...
+
+}  // namespace vision
+}  // namespace sdktests
+}  // namespace viam
+
+BOOST_TEST_DONT_PRINT_LOG_VALUE(Vision::detection)
+BOOST_TEST_DONT_PRINT_LOG_VALUE(Vision::classification)
+BOOST_TEST_DONT_PRINT_LOG_VALUE(Vision::properties)
+
+BOOST_AUTO_TEST_SUITE(vision_proto_conv)
```
lia-viam, line 84: harmless but generally speaking we don't bother testing the to_proto/from_proto conversions

src/viam/sdk/tests/test_vision.cpp
```diff
@@ -0,0 +1,467 @@
...
+
+namespace vimpl = viam::sdk::impl::vision;
+namespace vpb = viam::service::vision::v1;
+using viam::sdk::Vision;
+
+namespace viam {
+namespace sdktests {
+namespace vision {
+
+// vision_fixture stands up an in-process gRPC server backed by MockVision and a
```
lia-viam, line 38: this is interesting but i wonder if it's necessary, can you (personally, or pointing the agent to it) compare this file to test_motion or test_mlmodel and see if we don't end up with something simpler without this separate fixture

lia-viam (changes requested):
So this seems difficult for claude and for some human developers truthfully, but we've taken an approach to proto type conversion in the sdk that is very easy/ergonomic for module authors to use, at the cost of being kind of cumbersome to implement on the sdk-side.

Try pointing the agent to, eg, https://github.com/viamrobotics/viam-cpp-sdk/blob/d54d7fbb877edce12bb201eac6e1803d0511d8fb/src/viam/sdk/common/pose.hpp#L53 as well as https://github.com/viamrobotics/viam-cpp-sdk/blob/main/src/viam/sdk/common/proto_convert.hpp

The idea is if you put struct specializations in a header, along with appropriate forward declarations, then end-users can include the `viam/api` files and be able to just have `to_proto` and `from_proto` work automatically.

There are some cases in the codebase where we do manually implement to_proto and from_proto in a cpp file or in the `private/` directory; in these cases it is judged that conversion to/from these types is not likely to be needed outside of implementation details. However, since, eg, `raw_image` is now used by both `Camera` and `Vision` it certainly passes the test for providing public conversion functions

There are also cases as I've commented in one of the files where we don't bother implementing to_ and from_proto because the function will only be used once, so we just do the conversion inline in the respective _client and _server files

PR #645, opened 2026-05-20
RSDK-13982 - construct model table from urdf

nfranczak:
as follow up will do sva


lia-viam:
Mentioned this in the previous PR but I'm not convinced we need Eigen? I would be in favor of reusing `Vector3` from `linear_algebra.hpp` or an appropriate [xtensor type](https://xtensor.readthedocs.io/en/latest/quickref/basic.html)

lia-viam (commented):
lgtm, try running `bin/run_clang_format.sh` if linter still isn't passing

conanfile.py
```diff
@@ -91,6 +91,7 @@ def requirements(self):
         self.requires(self._grpc_requires())
         self.requires('protobuf/[>=3.17.1 <6.30.0]')
         self.requires(self._xtensor_requires(), transitive_headers=True)
+        self.requires('eigen/[>=3.3.0]', transitive_headers=True)
```
acmorrow: Is this still true? Ditto below?

src/viam/sdk/common/linear_algebra.hpp
```diff
@@ -38,6 +38,13 @@ struct Vector3 {
     std::array<scalar_type, 3> data;
 };
 
+inline bool operator==(const Vector3& a, const Vector3& b) {
+    return a.data == b.data;
+}
+inline bool operator!=(const Vector3& a, const Vector3& b) {
```
acmorrow, line 45: I'm surprised the formatter didn't do it for you, but please throw a blank line inbetween these definitions.

src/viam/sdk/referenceframe/kinematics_model_table.hpp
```diff
@@ -0,0 +1,58 @@
...
+#endif
+
+#include <viam/sdk/common/kinematics.hpp>
+#include <viam/sdk/common/linear_algebra.hpp>
+
+namespace viam {
+namespace sdk {
+
+/// @brief URDF joint type, restricted to arm-relevant joints.
+enum class JointType { revolute, continuous, prismatic, fixed };
```
acmorrow: These should be k_revolute, etc.

src/viam/sdk/referenceframe/kinematics_model_table.hpp
```diff
@@ -0,0 +1,58 @@
...
+/// @brief Parse a kinematics description and return the per-joint model
+/// table in chain order. Currently only URDF input is supported; SVA
+/// support is planned.
+/// @throws viam::sdk::Exception on parse error, branching / multi-root /
+///         disconnected topology, unsupported joint types, or zero axis.
+ModelTable kinematics_to_model_table(const KinematicsDataURDF& urdf);
+
+/// @brief Convert a model table to a double tensor of shape (n, 10).
+/// Columns: 0..2 xyz, 3..5 rpy, 6..8 axis, 9 joint type as
+/// underlying enum value (0=revolute, 1=continuous, 2=prismatic, 3=fixed).
```
acmorrow: If you want to do that, I'd suggest giving the JointType enum values matching explicit values.

src/viam/sdk/referenceframe/kinematics_model_table.hpp
```diff
@@ -0,0 +1,58 @@
...
+
+namespace viam {
+namespace sdk {
+
+/// @brief URDF joint type, restricted to arm-relevant joints.
+enum class JointType { revolute, continuous, prismatic, fixed };
+
+/// @brief One row of the model table: the per-joint URDF fields.
+/// @note `xyz`/`rpy`/`axis` are taken directly from the URDF.
+struct JointRow {
```
acmorrow: You could put `JointType `in here, so you had `JointRow::[T|t]ype::k_[revolute,continuous,prismatic,fixed]`.

src/viam/sdk/referenceframe/private/kinematics_model_table_internals.hpp
```diff
@@ -0,0 +1,43 @@
...
+
+#include <boost/optional.hpp>
+
+#include <viam/sdk/common/kinematics.hpp>
+#include <viam/sdk/common/linear_algebra.hpp>
+#include <viam/sdk/referenceframe/kinematics_model_table.hpp>
+
+namespace viam {
+namespace sdk {
+namespace urdf_model_table_internals {
```
acmorrow: Elsewhere, where we have had private headers, we have gathered the inner namespace as `impl` (we can't call it `private` since that is a keyword). It might be better to stick to that convention.

src/viam/sdk/referenceframe/kinematics_model_table.cpp
```diff
@@ -0,0 +1,263 @@
...
+        throw Exception(ErrorCondition::k_not_supported,
+                        "URDFToModelTable: joint '" + parsed.name + "' has unsupported type '" +
+                            parsed.type_str +
+                            "' (supported: revolute, continuous, prismatic, fixed)");
+    }
+
+    if (row.type == JointType::fixed) {
+        row.axis = Vector3{0, 0, 0};
+    } else if (parsed.axis_opt) {
+        if (magnitude(*parsed.axis_opt) < 1e-12) {
```
acmorrow: Why this value and not some other?

src/viam/sdk/referenceframe/kinematics_model_table.cpp
```diff
@@ -0,0 +1,263 @@
...
+    }
+    if (tensor.shape()[0] == 0) {
+        throw Exception(ErrorCondition::k_general, "tensor_to_model_table: empty tensor");
+    }
+    ModelTable table;
+    table.reserve(tensor.shape()[0]);
+    for (std::size_t i = 0; i < tensor.shape()[0]; ++i) {
+        const double type_v = tensor(i, 9);
+        const int type_i = static_cast<int>(type_v);
+        if (static_cast<double>(type_i) != type_v || type_i < 0 || type_i > 3) {
```
acmorrow: Probably want to compare against the enum values here, not constants.

acmorrow (changes requested):
This looks good, I have some comments to consider though.

src/viam/sdk/referenceframe/kinematics_model_table.hpp
```diff
@@ -0,0 +1,65 @@
...
+/// @brief URDF joint type, restricted to arm-relevant joints.
+/// @note Underlying values are stable and form the wire encoding for
+/// column 9 of the tensor produced by `model_table_to_tensor`.
+enum class JointType {
+    k_revolute = 0,
+    k_continuous = 1,
+    k_prismatic = 2,
+    k_fixed = 3,
+};
+
```
acmorrow, line 35: Are there any connections between the set of possible values of these vectors, and the types of joints?

You could consider adding static factory methods on `JointRow`, like `JointRow::fixed(std::string name, ...)`.

src/viam/sdk/referenceframe/kinematics_model_table.hpp
acmorrow: Ping?

src/viam/sdk/referenceframe/kinematics_model_table.hpp
```diff
@@ -0,0 +1,65 @@
...
+/// @note `xyz`/`rpy`/`axis` are taken directly from the URDF.
+struct JointRow {
+    std::string name;
+    Vector3 xyz{};
+    Vector3 rpy{};
+    Vector3 axis{};
+    JointType type = JointType::k_fixed;
+};
+
+using ModelTable = std::vector<JointRow>;
```
acmorrow: Should this be a class? Should `JointRow` be nested within it? Then:

```
ModelTable ModelTable::from(const KinematicsDataURDF& urdf);  //static 
ModelTable ModelTable::from(const xt::xarray<double>& tensor);  // static
xt::xarray<double> ModelTable::to_tensor() const;
```

Etc?

Then just give it some accessors:

```
const std::vector<ModelTable::JointRow>& ModelTable::rows();
```

src/viam/sdk/referenceframe/private/kinematics_model_table_internals.hpp
```diff
@@ -0,0 +1,43 @@
+/// @file referenceframe/private/kinematics_model_table_internals.hpp
```
acmorrow: I don't think you should doc comment it if it is under `private`, since otherwise doxygen docs will I think end up picking it up, even though consumers of the SDK can't use it.

src/viam/sdk/referenceframe/kinematics_model_table.cpp
```diff
@@ -0,0 +1,271 @@
...
+        }
+        joints.push_back(std::move(j));
+    }
+    return joints;
+}
+
+std::vector<ParsedJoint> walk_urdf_chain(const std::vector<ParsedJoint>& joints) {
+    std::map<std::string, std::vector<const ParsedJoint*>> by_parent;
+    std::set<std::string> all_parents;
+    std::set<std::string> all_children;
```
acmorrow: Do you need ordering on these? Or would the `unordered` form work?

src/viam/sdk/referenceframe/kinematics_model_table.cpp
```diff
@@ -0,0 +1,271 @@
...
+            list += "'" + r + "' ";
+        }
+        throw Exception(ErrorCondition::k_general,
+                        "URDFToModelTable: expected exactly one root link, found " +
+                            std::to_string(roots.size()) + ": " + list);
+    }
+
+    std::vector<ParsedJoint> ordered;
+    ordered.reserve(joints.size());
+    std::set<std::string> visited;
```
acmorrow: Do you need ordering on `visited`?

acmorrow (changes requested):
Similar thoughts as the UR Arm review about pushing harder to make these classes, rather than structs with helper functions.

src/viam/sdk/referenceframe/kinematics_model_table.cpp
nfranczak: `by_parent` and `all_children` can be `unordered` since we only find the values that they hold.
I'd like `all_parents` to remain ordered for this reason: we iterate through `all_parents` to see if there are multiple roots in the urdf (i.e. multiple ties to the world frame), by having an ordering that makes the error message clearer and not vary across runs.

src/viam/sdk/referenceframe/kinematics_model_table.cpp
nfranczak: yep, will change.

src/viam/sdk/referenceframe/kinematics_model_table.hpp
nfranczak: xyz and rpy just tell us the position and orientation of "things", eg. links.
axis is our biggest hint here.
If it's `(0, 0, 0)` then we can deduce that it's a fixed joint with certainty.
But on the other hand if it's a non-zero unit vector we do not have a way of knowing if the joint type is revolute, continuous or prismatic.

src/viam/sdk/referenceframe/kinematics_model_table.hpp
nfranczak: made it into a class.

src/viam/sdk/referenceframe/kinematics_model_table.hpp
nfranczak: I am not a fan of this.
It feels like there is less cognitive overhead with keep it this way.
edit: and with your proposal the jointtype is hidden one level deeper right?

acmorrow: I think it comes down to this question: Is `JointType` something intrinsically useful, or is it just a discriminator to understand how to interpret the fields of `JointRow`. If it is intrinsically valuable, then I'm more OK with making it a top-level type on its own. If it is just the descriminator for `JointRow`, I'd be in favor of nesting it.

src/viam/sdk/referenceframe/private/kinematics_model_table_internals.hpp
acmorrow: You still ahve the rest of this file doc commented though.

It should be easy, just change /// to //

src/viam/sdk/referenceframe/kinematics_model_table.hpp
acmorrow: FYI now that they are both at least in `ModelTable` I'm less inclined to push `JointType` into `JointRow` though I do think the argument holds.

I'm resisting the urge to say that `JointRow` should be some sort of variant, because if I see it correctly, the only real structural variant is for the fixed case.

src/viam/sdk/referenceframe/kinematics_model_table.hpp
```diff
@@ -0,0 +1,76 @@
...
+
+/// @brief A per-joint model table for a serial kinematic chain, held in chain
+/// order. Build one with the `from` factories and read it back via `rows()` or
+/// `to_tensor()`.
+class ModelTable {
+   public:
+    /// @brief URDF joint type, restricted to arm-relevant joints.
+    /// @note Underlying values are stable and form the wire encoding for
+    /// column 9 of the tensor produced by `to_tensor()`.
+    enum class JointType {
```
acmorrow, line 29: I suspect the linter is going to want you to give this a smaller integral type.

acmorrow (approved):
LGTM with a few lingering comments. @lia-viam can you please also take a look?

src/viam/sdk/referenceframe/kinematics_model_table.cpp
```diff
@@ -0,0 +1,283 @@
...
+
+// The row/type are nested in ModelTable; alias them for brevity in the
+// URDF-pipeline helpers below.
+using JointRow = ModelTable::JointRow;
+using JointType = ModelTable::JointType;
+
+std::vector<ParsedJoint> parse_urdf(const KinematicsDataURDF& urdf) {
+    namespace pt = boost::property_tree;
+    pt::ptree tree;
+    const std::string text(reinterpret_cast<const char*>(urdf.bytes.data()), urdf.bytes.size());
```
lia-viam: we have [`bytes_to_string`](https://github.com/viamrobotics/viam-cpp-sdk/blob/745b456cc863e8684f9888c1f769f4219a807645/src/viam/sdk/common/utils.hpp#L84) for this

src/viam/sdk/referenceframe/kinematics_model_table.cpp
```diff
@@ -0,0 +1,283 @@
...
+    return ordered;
+}
+
+JointRow to_row(const ParsedJoint& parsed) {
+    JointRow row;
+    row.name = parsed.name;
+    row.xyz = parsed.xyz;
+    row.rpy = parsed.rpy;
+
+    if (parsed.type_str == "revolute") {
```
lia-viam, line 165: I wonder if we want a public `to_string` or `joint_type_from_string` ?

src/viam/sdk/referenceframe/kinematics_model_table.cpp
```diff
@@ -0,0 +1,283 @@
...
+}
+
+xt::xarray<double> ModelTable::to_tensor() const {
+    if (rows_.empty()) {
+        throw Exception(ErrorCondition::k_general, "ModelTable::to_tensor: empty model table");
+    }
+    xt::xarray<double> out = xt::zeros<double>({rows_.size(), std::size_t{10}});
+    for (std::size_t i = 0; i < rows_.size(); ++i) {
+        const auto& r = rows_[i];
+        out(i, 0) = r.xyz.x();
```
lia-viam, line 216: I don't love this but struggling to come up with an alternative

```cpp
    int idx = 0;
    for (const auto& vec3 : {std::cref(r.xyz), std::cref(r.rpy), std::cref(r.axis)}) {
        for (const auto& val : arr.data) {
            out(i, idx++) = val;
        }
    }
```

honestly not sure if this is preferable, your call

lia-viam (commented):
mostly minor/optional stuff but still lgtm

src/viam/sdk/referenceframe/kinematics_model_table.hpp
nfranczak: A bit confused by this, the linter did not complain.

src/viam/sdk/referenceframe/kinematics_model_table.cpp
nfranczak: I am not convinced that we need a public `to_string` or  `joint_type_from_string`.
My reasoning is that users always get a parsed `JointType` values values via `from(urdf)` / `from(tensor)` and  never see  the strings.
The `to_string` has no user anywhere in this flow.

src/viam/sdk/referenceframe/kinematics_model_table.cpp
nfranczak: while more compact/abstract, I think that what we have now is slightly clearer to read.

src/viam/sdk/referenceframe/private/kinematics_model_table_internals.hpp
nfranczak: Sorry a bit confused by this: in the current version of the code I just removed all the lines which were prefixed with : `///`

src/viam/sdk/referenceframe/kinematics_model_table.cpp
lia-viam: in that case maybe just a line break between the blocks where you're operating on the different vec3's

nfranczak: done!

src/viam/sdk/referenceframe/kinematics_model_table.hpp
acmorrow: OK, maybe we don't do that anymore.

src/viam/sdk/referenceframe/private/kinematics_model_table_internals.hpp
```diff
@@ -0,0 +1,40 @@
...
+
+#include <viam/sdk/common/kinematics.hpp>
+#include <viam/sdk/common/linear_algebra.hpp>
+#include <viam/sdk/referenceframe/kinematics_model_table.hpp>
+
+namespace viam {
+namespace sdk {
+namespace impl {
+
+/// @brief Per-joint parse result, pre-validation.
```
acmorrow: Here, for instance, there is still a doxygen style comment on an internal header.

nfranczak: ok I can change.
does that mean that this for example is incorrect?
https://github.com/viamrobotics/viam-cpp-sdk/blob/main/src/viam/sdk/rpc/private/viam_grpc_channel.hpp#L14

PR #652, opened 2026-05-29
RSDK-14047: gate options on standalone build

lia-viam:
We were already doing this for test/example building; add wall_werror and unify the approach.


acmorrow (approved):
LGTM, thanks!

PR #663, opened 2026-06-15
RSDK-14090: Docker bake base

lia-viam:
First step of the next gen build docker stuff: replacing the base images with docker bake file, single dockerfile, and shell scripts handling platform-specific stuff. I have tested most permutations of this locally, but only on my arm64 mac due to the stated goal of not using QEMU. 

TBD if we want to dogfood this by replacing some of the CI pipelines with it now, or save that for a future PR.

I also created a ticket to track the docs work in deleting the old stuff, but I'm going to try to do that later this week after getting a bunch of the base work out of the way.


etc/docker/scripts/install-base.sh
```diff
@@ -0,0 +1,32 @@
...
+    gcc \
+    gdb \
+    git \
+    gnupg \
+    gpg \
+    less \
+    ninja-build \
+    pkg-config \
+    sudo \
+    wget
```
acmorrow: Do we need: `gdb`, `less`, `doxygen`, and both of `curl` and `wget`? The lighter we can make these containers the better!

etc/docker/scripts/install-base.sh
```diff
@@ -0,0 +1,32 @@
...
+apt-get -y dist-upgrade
+rm -rf /var/lib/apt/lists/*
+
+apt_install \
+    build-essential \
+    ca-certificates \
+    curl \
+    doxygen \
+    g++ \
+    gcc \
```
acmorrow, line 21: Fairly sure that `build-essential` already includes gcc and c++

etc/docker/scripts/install-cmake.sh
```diff
@@ -0,0 +1,45 @@
...
+
+# shellcheck disable=SC1091
+. "$(dirname "$0")/lib/common.sh"
+
+: "${CMAKE_MIN_VERSION:?CMAKE_MIN_VERSION must be set}"
+
+# Candidate version apt would install, e.g. "3.25.1-1". Empty if no candidate.
+apt-get update
+candidate="$(apt-cache policy cmake 2>/dev/null | awk '/Candidate:/ {print $2}')"
+rm -rf /var/lib/apt/lists/*
```
acmorrow: I wonder if it would be better to just include `cmake` among the packages in `install_base`, and then query its version directly. Then, if needed, we can replace with newer, and the query can just be by running `cmake --version` or similar.

You might even be able to here doc up a trivial CMakeLists.txt with a requirement on that version and then run it?

etc/docker/scripts/install-grpc-from-source.sh
```diff
@@ -0,0 +1,39 @@
...
+# source, against the system zlib/c-ares/re2/openssl. Used on distros whose apt
+# grpc is missing or too old.
+
+set -euo pipefail
+[[ "${DEBUG:-}" ]] && set -x
+
+# shellcheck disable=SC1091
+. "$(dirname "$0")/lib/common.sh"
+
+: "${GRPC_VERSION:?GRPC_VERSION must be set}"
```
acmorrow: I sort of liked how the CMake setup handled the "system vs install" aspect internally. Is there a reason the grpc one doesn't work the same way?

etc/docker/scripts/install-grpc-from-source.sh
```diff
@@ -0,0 +1,39 @@
...
+    -DgRPC_BUILD_TESTS=OFF \
+    -DBUILD_SHARED_LIBS=ON \
+    -DCMAKE_BUILD_TYPE=Release \
+    -DCMAKE_INSTALL_PREFIX=/usr/local \
+    -DCMAKE_INSTALL_RPATH=/usr/local/lib
+ninja -C build install
+
+cd "${SRC_DIR}"
+rm -rf grpc
+ldconfig
```
acmorrow: Why `ldconfig`?

etc/docker/scripts/install-llvm.sh
```diff
@@ -0,0 +1,61 @@
...
+# update-alternatives below.
+LLVM_PKGS=(
+    "clang-${LLVM_VERSION}"
+    "clang-tidy-${LLVM_VERSION}"
+    "clang-format-${LLVM_VERSION}"
+)
+
+# Point the unversioned tool names at this LLVM_VERSION so callers that invoke
+# `clang` / `clang-tidy` / `clang-format` get the pinned toolchain.
+register_alternatives() {
```
acmorrow: This is a divergence from our current practice, so, most build scripts and such explicitly call clang-format-19 for instance. Maybe this is for the better, or, maybe the explicit versioning is safer. Not sure.

etc/docker/scripts/install-llvm.sh
```diff
@@ -0,0 +1,61 @@
...
+    update-alternatives --install /usr/bin/clang-format clang-format \
+        "/usr/bin/clang-format-${LLVM_VERSION}" 100
+}
+
+# Gate on the installable candidate, not `apt-cache show`: on some distros
+# clang-${LLVM_VERSION} is a known-but-uninstallable name ("referred to by
+# another package") that `show` reports yet apt cannot install.
+apt-get update
+clang_candidate="$(apt-cache policy "clang-${LLVM_VERSION}" 2>/dev/null | awk '/Candidate:/ {print $2}')"
+rm -rf /var/lib/apt/lists/*
```
acmorrow: I don't love this "grep for candidate" thing. There must be a better way to ask "what version do you use"?

etc/docker/scripts/install-llvm.sh
```diff
@@ -0,0 +1,61 @@
...
+
+if [[ -n "${clang_candidate}" && "${clang_candidate}" != "(none)" ]]; then
```
acmorrow: Here too, I like the pattern that the script makes its own decision, like the cmake one, leaving grpc as the odd one out.

etc/docker/scripts/install-system-deps.sh
```diff
@@ -0,0 +1,42 @@
...
+# from source at ${GRPC_VERSION} instead (against the system zlib/c-ares/re2/ssl).
+
+set -euo pipefail
+[[ "${DEBUG:-}" ]] && set -x
+
+# shellcheck disable=SC1091
+. "$(dirname "$0")/lib/common.sh"
+
+# Minimum apt grpc we accept before falling back to a source build.
+GRPC_MIN_VERSION="${GRPC_MIN_VERSION:-1.51}"
```
acmorrow: I feel like this belongs with the other MIN_VERSION things for LLVM and CMake

etc/docker/scripts/install-system-deps.sh
```diff
@@ -0,0 +1,42 @@
...
+    libboost-all-dev
+    libc-ares-dev
+    libre2-dev
+    libssl-dev
+    zlib1g-dev
+)
+
+apt-get update
+grpc_candidate="$(apt-cache policy libgrpc++-dev 2>/dev/null | awk '/Candidate:/ {print $2}')"
+rm -rf /var/lib/apt/lists/*
```
acmorrow: ditto re version detection.

etc/docker/docker-bake.hcl
```diff
@@ -0,0 +1,83 @@
...
+  default = "19"
+}
+
+variable "GRPC_VERSION" {
+  default = "v1.62.1"
+}
+
+// Native arch throughout. No QEMU emulation.
+variable "ARCH" {
+  default = "amd64"
```
acmorrow: What will the workflow at gh look like to get multiarch images across the different named layers?

etc/docker/docker-bake.hcl
```diff
@@ -0,0 +1,83 @@
...
+  // it from this strategy without touching any other matrix.
+  matrix = {
+    cell = [
+      { distro = "debian", version = "bullseye" },
+      { distro = "debian", version = "bookworm" },
+      { distro = "debian", version = "trixie" },
+      { distro = "debian", version = "sid" },
+      { distro = "ubuntu", version = "focal" },
+      { distro = "ubuntu", version = "jammy" },
+      { distro = "ubuntu", version = "noble" },
```
acmorrow, line 45: Raccoon!

etc/docker/Dockerfile
```diff
@@ -0,0 +1,42 @@
...
+# One Dockerfile, parameterized over distro, version, and strategy.
+# Stages:
+#   base   - distro + universal devkit + LLVM + cmake. Internal only, not published.
+#   system - base + apt -dev packages (grpc-from-source fallback on old distros). Published.
+#
+# Strategy is selected by bake `target = "system"`, not a build arg.
+# Distro-conditional logic lives in scripts/, not in RUN steps here.
+
+ARG DISTRO=debian
+ARG VERSION=bookworm
```
acmorrow: Why these as the defaults? Should these always be user provided?

acmorrow (changes requested):
I really like where this is going. I think there is an inconsistency around how gRPC vs version works as compared with cmake and llvm, and I think I like the way the latter did it better.

Also curious about what the workflow on GH will be to get these published as multiarch. If the answer is "I don't know yet" that's fine.

Have you been able to at least build all the amd64 slices on an amd64 machine and all the arm64 slices on an arm machine?

etc/docker/scripts/install-grpc-from-source.sh
lia-viam: apparently this rebuilds `/etc/ld.so.cache` so the loader will find the newly installed lib!

etc/docker/scripts/install-base.sh
lia-viam: was able to remove all these, and the `gcc` and `g++` below

etc/docker/scripts/install-grpc-from-source.sh
lia-viam: nope, changed!

etc/docker/scripts/install-llvm.sh
lia-viam: I thought so too but I think this may be our best option, the other approaches have a bunch of caveats that don't quite work here

etc/docker/scripts/install-grpc-from-source.sh
acmorrow: I'd sort of expect we wouldn't need to do this. The install prefix is /usr/local, and, if all went well, appropriate rpathing is in place.

etc/docker/docker-bake.hcl
lia-viam: done! 🦝

etc/docker/scripts/install-grpc-from-source.sh
lia-viam: upon further inspection yes that's accurate

etc/docker/scripts/install-cmake.sh
lia-viam: I wonder if this wouldn't just add some extra confusion around having to uninstall or `update-alternatives` on the new version? Perhaps better to not have installed the wrong one in the first place

lia-viam:
@acmorrow I think this is ready for another look; most of your comments should be addressed. meanwhile starting on the second ticket which does the CI build and merge flow

etc/docker/scripts/install-grpc.sh
```diff
@@ -0,0 +1,58 @@
...
+    -DgRPC_CARES_PROVIDER=package \
+    -DgRPC_RE2_PROVIDER=package \
+    -DgRPC_SSL_PROVIDER=package \
+    -DgRPC_INSTALL=ON \
+    -DgRPC_BUILD_TESTS=OFF \
+    -DBUILD_SHARED_LIBS=ON \
+    -DCMAKE_BUILD_TYPE=Release \
+    -DCMAKE_INSTALL_PREFIX=/usr/local \
+    -DCMAKE_INSTALL_RPATH=/usr/local/lib
+ninja -C build install
```
acmorrow: Slight preference for `cmake --build build --target install` over directly invoking Ninja

etc/docker/scripts/install-grpc.sh
```diff
@@ -0,0 +1,58 @@
...
+
+cmake -S . -B build -G Ninja \
+    -DgRPC_ZLIB_PROVIDER=package \
+    -DgRPC_CARES_PROVIDER=package \
+    -DgRPC_RE2_PROVIDER=package \
+    -DgRPC_SSL_PROVIDER=package \
+    -DgRPC_INSTALL=ON \
+    -DgRPC_BUILD_TESTS=OFF \
+    -DBUILD_SHARED_LIBS=ON \
+    -DCMAKE_BUILD_TYPE=Release \
```
acmorrow: Slight preference for `RelWithDebInfo` on this. At least with the system verisons, there is the possibility of installing the dbg package for postmortem debugging of a core and seeing into the symbols, but in the case we build grpc from source with Release we lose that. Yes, it'll cost us some real space in the image, but hopefully it's just Focal that is so affected?

etc/docker/Dockerfile
```diff
@@ -0,0 +1,42 @@
...
+
+# DEBIAN_FRONTEND set here once, not per-call in scripts.
+ENV DEBIAN_FRONTEND=noninteractive \
+    HOME=/root
+
+# Versions and toggles ride in on build args; distro identity comes from /etc/os-release.
+ARG LLVM_VERSION=19
+ARG CMAKE_MIN_VERSION=3.25
+# DEBUG=1 turns on `set -x` tracing in every install script.
+ARG DEBUG=
```
acmorrow: Maybe pull all the `ARG`s up top, and group them by theme? Something like:

```
# Required parameters
ARG DISTRO
ARG VERSION

# Options to control the build
ARG DEBUG=

# Default verison (minima)
ARG CMAKE_MIN_VERSION=3.25
ARG LLVM_VERSION=19
```

And then the `ENV` section hangs together naturally.

etc/docker/Dockerfile
```diff
@@ -0,0 +1,42 @@
...
+ENV LLVM_VERSION=${LLVM_VERSION} \
+    CMAKE_MIN_VERSION=${CMAKE_MIN_VERSION} \
+    DEBUG=${DEBUG}
+
+COPY scripts/ /opt/scripts/
```
acmorrow, line 27: Does this mean these files persist into the published containers? Do we want that?

etc/docker/Dockerfile
```diff
@@ -0,0 +1,42 @@
...
+
+RUN /opt/scripts/install-base.sh
+RUN /opt/scripts/install-llvm.sh
+RUN /opt/scripts/install-cmake.sh
+
+
+FROM base AS system
+
+ARG GRPC_VERSION=v1.62.1
+ENV GRPC_VERSION=${GRPC_VERSION}
```
acmorrow: I'd move these up too. Also, is 1.62.1 the version, or the version minimum?

etc/docker/Dockerfile
```diff
@@ -0,0 +1,42 @@
...
+
+FROM ${DISTRO}:${VERSION} AS base
+
+# DEBIAN_FRONTEND set here once, not per-call in scripts.
+ENV DEBIAN_FRONTEND=noninteractive \
+    HOME=/root
+
+# Versions and toggles ride in on build args; distro identity comes from /etc/os-release.
+ARG LLVM_VERSION=19
+ARG CMAKE_MIN_VERSION=3.25
```
acmorrow: If the `bake` file sets defaults for these, do they belong in the docker file with defaults too?

acmorrow (changes requested):
This is looking very close to done to me. Mostly, my push back here is about grouping things. I do have one more general area of concern, which is having one source of truth for versions and verison minima. Does it belong in the bake file, or the docker file? And I'd like to see it handled entirely uniformly across all the things for which we want to set a version (minimum), which I think is just three: llvm, cmake, grpc?

stuqdog (commented):
this seems reasonable to me but I don't think I have sufficient context to provide a good review here, I defer to Drew's thoughts.

etc/docker/Dockerfile
lia-viam: Closing the loop on this--1.62.1 is chosen somewhat arbitrarily as a semi-recent stable build. We could have it just build the apt version, which is 1.51, on all platforms, but there is now an override build for building the exact version from source

etc/docker/Dockerfile
```diff
@@ -0,0 +1,45 @@
...
+RUN /opt/scripts/install-llvm.sh
+RUN /opt/scripts/install-cmake.sh
+
+
+FROM base AS system
+
+# Optional override for the grpc version built from source (numeric, e.g. 1.65.0).
+# When set, forces a source build at that version on every distro, even where the
+# apt grpc would otherwise be accepted. Empty -> default behavior.
+ARG GRPC_SOURCE_VERSION_OVERRIDE=
```
acmorrow, line 40: [nit]: Personal preference would be to have all the `ARG`s up at the top for discoverability. The `ENV` application of it can come later, as you have it.

etc/docker/Dockerfile
```diff
@@ -0,0 +1,45 @@
...
+ENV GRPC_SOURCE_VERSION_OVERRIDE=${GRPC_SOURCE_VERSION_OVERRIDE}
+
+RUN /opt/scripts/install-system-deps.sh
```
acmorrow, line 43: [minor]: Does `RUN GRPC_SOURCE_VERSION_OVERRIDE=${GRPC_SOURCE_VERSION_OVERRIDE} /opt/scripts/install-system-deps.sh` work? That'd let you avoid polluting the global env for commands after this one with the grpc version override environment variable.

etc/docker/docker-bake.hcl
```diff
@@ -0,0 +1,71 @@
...
+group "default" {
+  targets = ["system"]
+}
+
+// Internal stage only; not published. Holds everything common to the strategies
+// stacked on top of it. `system` inherits this and points `target` at its own
+// Dockerfile stage.
+target "base" {
+  dockerfile = "Dockerfile"
+  context    = "."
```
acmorrow, line 23: What does it mean that this `target` has `context = '.'` but the `target` `system` does not?

etc/docker/docker-bake.hcl
```diff
@@ -0,0 +1,71 @@
...
+  target     = "base"
+
+  // amd64 / arm64 built on native runners; one manifest list per cell.
+  platforms = ["linux/${ARCH}"]
```
acmorrow: Does this mean that the bake matrix is only aware of the slices for the platform on which it runs? I sort of feel like the bake declaration should be declaring the whole matrix, so that you could, in principle, build it entirely locally under emulation. But then in practice in GH we use runners and matrix things to do native runs and combine later.

etc/docker/docker-bake.hcl
```diff
@@ -0,0 +1,71 @@
...
+}
+
+target "system" {
+  inherits = ["base"]
+  name     = "system-${cell.distro}-${cell.version}"
```
acmorrow, line 32: What does this `name` do? How does it relate to the `tags` below?

etc/docker/docker-bake.hcl
```diff
@@ -0,0 +1,71 @@
...
+    ]
+  }
+
+  args = {
+    DISTRO  = cell.distro
+    VERSION = cell.version
+  }
+
+  tags = [
+    "${REGISTRY}/cpp-sdk-system:${cell.distro}-${cell.version}",
```
acmorrow: - Can this re-use `name` from the target declaration, at least in part?
- Slight preference for `${cell.distro}:${cell.version}"` since that is the same pattern that the base distros use (e.g. `ubuntu:focal`). So maybe `cpp-sdk-system-${...` so there is only one `:`

etc/docker/docker-bake.hcl
```diff
@@ -0,0 +1,71 @@
...
+// Native arch throughout. No QEMU emulation.
+variable "ARCH" {
+  default = "amd64"
+}
+
+group "default" {
+  targets = ["system"]
+}
+
+// Internal stage only; not published. Holds everything common to the strategies
```
acmorrow: I don't see much harm in publishing these, maybe even useful in some cases, and probably costs nothing since the layers are getting pushed anyway.

etc/docker/scripts/install-system-deps.sh
```diff
@@ -0,0 +1,25 @@
...
+# from source instead (against the system zlib/c-ares/re2/ssl). See install-grpc.sh.
+
+set -euo pipefail
+[[ "${DEBUG:-}" ]] && set -x
+
+# shellcheck disable=SC1091
+. "$(dirname "$0")/lib/common.sh"
+
+# Libraries needed regardless of how grpc arrives.
+COMMON_DEV_PKGS=(
```
acmorrow, line 14: Later, we may want to do a dev vs runtime split. That'd let us build up a very light container with the C++ SDK preinstalled but no dev dependencies dragged along, just the runtime bits.

etc/docker/scripts/install-grpc.sh
```diff
@@ -0,0 +1,72 @@
...
+    -DgRPC_ZLIB_PROVIDER=package \
+    -DgRPC_CARES_PROVIDER=package \
+    -DgRPC_RE2_PROVIDER=package \
+    -DgRPC_SSL_PROVIDER=package \
+    -DgRPC_INSTALL=ON \
+    -DgRPC_BUILD_TESTS=OFF \
+    -DBUILD_SHARED_LIBS=ON \
+    -DCMAKE_BUILD_TYPE=RelWithDebInfo \
+    -DCMAKE_INSTALL_PREFIX=/usr/local \
+    -DCMAKE_INSTALL_RPATH=/usr/local/lib
```
acmorrow, line 68: Maybe later we want to canonicalize `/opt/viam` as where we install stuff. Again, thinking ahead to some images with a pre-installed Viam C++ SDK at each release level.

acmorrow (approved):
Mostly questions and nits.

etc/docker/docker-bake.hcl
lia-viam: updated the tag format, apparently it can't reuse the target name decl however

etc/docker/docker-bake.hcl
lia-viam: this creates per-cell bake targets that we can build, eg `system-debian-bullseye`, and ends up getting used to refer to the cells in the CI job for building these

PR #665, opened 2026-06-23
Bump quantstack minima to version with minimum required CMake 3.5

lia-viam:
This fixes builds with CMake 4, which dropped support for cmake below 3.5, because the older tags we were pointing to had a minimum required of 3.1.  

This means we no longer need to manually install CMake 3.25 in a lot of places, but a forthcoming PR built on top of #663 will take care of that


lia-viam:
currently running module integration tests to make sure this doesn't break anything

acmorrow (approved):
LGTM if CI and module build tests support it.

PR #667, opened 2026-06-29
bump conan cppstd min

lia-viam:
fixes build breakage from #665


conanfile.py
```diff
@@ -65,18 +65,12 @@ def validate(self):
 
 
     def _xtensor_requires(self):
-        if valid_max_cppstd(self, 14, False):
-            return 'xtensor/[>=0.24.3 <0.26.0]'
-
         if valid_max_cppstd(self, 17, False):
-            return 'xtensor/[>=0.24.3 <0.27.0]'
+            return 'xtensor/[>=0.25.0 <0.27.0]'
```
acmorrow, line 69: - Is 27 still an upper bound? If so, do we know why?
- Do you still need the valid_max_cppstd check against 17?
- Depending on the answer to that, maybe you don't need _xtensor_requies anymore?

conanfile.py
```diff
@@ -90,7 +84,9 @@ def requirements(self):
         # maintained conan packages.
         self.requires(self._grpc_requires())
         self.requires('protobuf/[>=3.17.1 <6.30.0]')
+
         self.requires(self._xtensor_requires(), transitive_headers=True)
+        self.requires('xtl/[>=0.7.7]', override=True)
```
acmorrow, line 92: Why `override=True`? I think I know why, but a comment might save future us some archaeological efforts.

acmorrow (approved):
LGTM with some minor questions and a comment suggestion

conanfile.py
lia-viam: yes, 0.27.x was when they start to require a c++20 compiler. i believe we still need valid_max_cppstd but the naming of these helpers is highly counter intuitive to me

conanfile.py
lia-viam: good call, will add a comment

stuqdog (approved):

PR #675, opened 2026-07-07
[RSDK-14159] Add support for streaming kinematic points to arms

acmorrow:
This depends on https://github.com/viamrobotics/api/pull/874.

I'm opening this as a draft because claude wrote most of this. I plan to do a self-review here, then point claude at my comments, as an experiment in delivering feedback to claude.

Once I'm satisfied with its revisions above this base, I will push an update and promote this to a real PR.


src/viam/sdk/components/private/arm_client.cpp
```diff
@@ -1,13 +1,20 @@
 #include <boost/variant/apply_visitor.hpp>
```
acmorrow: claude: [nit] not your doing but this visitor header is misplaced. The `arm_client.hpp` header should on its own as the first header. Please move the visitor header down into its own group, immediately before or after the grpc headers.

src/viam/sdk/components/private/arm_trajectory_validation.hpp
```diff
@@ -0,0 +1,78 @@
...
+//   - when constraints are present, velocities carries one entry per position;
+//   - when accelerations are present, they carry one entry per velocity.
+//
+// Dimensions are only checked for internal consistency; the trajectory is not
+// validated against the arm's actual degrees of freedom.
+class TrajectoryStreamValidator {
+   public:
+    // Returns a description of the first invariant the point violates, or
+    // boost::none if the point is valid in sequence.
+    boost::optional<std::string> check(const Arm::trajectory_point& point) {
```
acmorrow: claude: [question] - should this just throw on error instead? Do we ever not promote an engaged return into an exception somehow? I see the comment above explaining THAT this is handled, but it doesn't really clarify. If returning an error string is in fact the right choice, the comment should say WHY the client and server handling differs such that an error return is preferable to a throw exception.

acmorrow: claude: [improvement] - unless needed for performance reasons, this function should be defined out of line in a .cpp file.

src/viam/sdk/components/private/arm_trajectory_validation.hpp
```diff
@@ -0,0 +1,78 @@
+/// @file components/private/arm_trajectory_validation.hpp
+///
+/// @brief Shared invariant checking for streamed arm trajectories.
```
acmorrow: claude: [change] - no doc comments in `private` files.

src/viam/sdk/components/arm.hpp
```diff
@@ -105,6 +121,137 @@ class Arm : public Component, public Stoppable {
...
+    ///
+    /// The first point of a stream must have @ref time equal to zero; subsequent
+    /// points must be strictly increasing in time across the entire stream.
+    struct trajectory_point {
+        /// @brief Target kinematic constraints at a waypoint.
+        ///
+        /// `velocities` is required whenever this struct is present;
+        /// `accelerations` is independently optional. The
+        /// accelerations-require-velocities invariant is enforced by the type
+        /// system, not just documentation.
```
acmorrow: claude: [nit] - this comment about the type system is unnecessary

src/viam/sdk/components/arm.hpp
```diff
@@ -105,6 +121,137 @@ class Arm : public Component, public Stoppable {
...
+
+        /// @brief Offset of this waypoint from the start of the trajectory.
+        ///
+        /// Must be zero at the first point of a stream and strictly increasing
+        /// thereafter. Microsecond resolution matches the SDK's
+        /// `google.protobuf.Duration` conversion; sub-microsecond distinctions
+        /// are not preserved on the wire.
+        std::chrono::microseconds time;
+        std::vector<double> positions;
+        boost::optional<kinematic_constraints> constraints;
```
acmorrow, line 154: claude: [change] - if you are going to doc comment one of the fields, you should doc comment all of them.

src/viam/sdk/components/arm.hpp
```diff
@@ -105,6 +121,137 @@ class Arm : public Component, public Stoppable {
...
+    };
+
+    /// @brief An update emitted by the implementation as a streamed trajectory
+    /// executes.
+    ///
+    /// Updates flow back independently of the input batches: the implementation
+    /// may emit them at any cadence, or not at all, and a future implementation
+    /// may deliver them as a truly asynchronous channel. Currently empty; it
+    /// will grow fields carrying per-execution status.
```
acmorrow: claude: [change] - "an arm may emit them at any cadence" is all you need to say. Strike the rest of the comment.

src/viam/sdk/components/arm.hpp
```diff
@@ -105,6 +121,137 @@ class Arm : public Component, public Stoppable {
...
+    struct trajectory_update {};
+
+    /// @brief How a streamed trajectory execution ended.
+    enum class stream_outcome : std::uint8_t {
+        k_completed,                 ///< The trajectory ran to its natural end.
+        k_halted_by_update_handler,  ///< update_handler returned false, stopping the stream early.
```
acmorrow: claude: [change] - please give these explicit values starting at zero for `k_completed`.

src/viam/sdk/tests/mocks/mock_arm.cpp
```diff
@@ -2,6 +2,10 @@
 #include "viam/sdk/rpc/grpc_context_observer.hpp"
```
acmorrow: claude: [nit] - not your change, but `mock_arm.hpp` should be the first header, in its own group. Rehome `grpc_context_observer.hpp` into the group of sdk headers below. Also note that it has the wrong quote style.

src/viam/sdk/tests/mocks/mock_arm.cpp
```diff
@@ -2,6 +2,10 @@
...
 #include <viam/sdk/tests/mocks/mock_arm.hpp>
 
+#include <stdexcept>
+
+#include <grpcpp/support/status.h>
+
 #include "mock_arm.hpp"
```
acmorrow: claude: [nit] - This `mock_arm.hpp` looks wrong and duplicate.

src/viam/sdk/tests/mocks/mock_arm.hpp
```diff
@@ -39,6 +45,17 @@ class MockArm : public sdk::Arm {
...
     sdk::ProtoStruct peek_command;
     std::vector<std::string> viam_client_metadata;
+
+    // Streaming trajectory hooks. move_through_joint_positions_streamed records
+    // each batch it receives into peek_streamed_batches and counts the updates
+    // it emits in peek_streamed_ack_count. If streamed_fault is set, it throws
+    // after draining to exercise terminal-error propagation: runtime_error maps
+    // to a std::exception (INTERNAL at the wire), grpc_status to a thrown
+    // grpc::Status carried through verbatim.
+    enum class StreamFault : std::uint8_t { none, runtime_error, grpc_status };
```
acmorrow: claude: [change] - enum should be `stream_fault`, and values should be `k_none`, etc. Probably best to give them explicit values starting at zero as well.

src/viam/sdk/tests/mocks/mock_arm.hpp
```diff
@@ -39,6 +45,17 @@ class MockArm : public sdk::Arm {
     bool peek_stop_called;
     sdk::ProtoStruct peek_command;
     std::vector<std::string> viam_client_metadata;
+
+    // Streaming trajectory hooks. move_through_joint_positions_streamed records
```
acmorrow: claude: [change] - this comment needs help. It reads like AI, and it doesn't really seem to say anything.

src/viam/sdk/tests/test_arm.cpp
```diff
@@ -7,6 +7,15 @@
 #include <boost/variant/apply_visitor.hpp>
 #include <boost/variant/static_visitor.hpp>
```
acmorrow, line 10: claude: [nit] - not your change, but these headers are sorted wrong. They should be a group down around grpc.

src/viam/sdk/components/private/arm_client.cpp
```diff
@@ -104,6 +111,119 @@ void ArmClient::move_through_joint_positions(const std::vector<std::vector<doubl
         .invoke();
 }
 
+Arm::stream_outcome ArmClient::move_through_joint_positions_streamed(
+    const std::function<boost::optional<std::vector<Arm::trajectory_point>>()>& batch_source,
+    const std::function<bool(Arm::trajectory_update)>& update_handler,
+    const ProtoStruct& extra) {
+    // Use the SDK's ClientContext wrapper rather than a raw grpc::ClientContext.
```
acmorrow: claude: [change] - Tag this whole comment as `TODO(RSDK-14164)` since that is the ticket to add the BiDi client/server helpers.

src/viam/sdk/components/private/arm_client.cpp
```diff
@@ -104,6 +111,119 @@ void ArmClient::move_through_joint_positions(const std::vector<std::vector<doubl
...
+    ClientContext ctx(*channel_);
+    auto stream = stub_->MoveThroughJointPositionsStreamed(ctx);
+
+    // Send Init on the caller's thread before any other thread is involved.
+    ::viam::component::arm::v1::MoveThroughJointPositionsStreamedRequest init_msg;
+    init_msg.set_name(this->name());
+    auto* init_payload = init_msg.mutable_init();
+    *init_payload->mutable_extra() = to_proto(extra);
+    if (!stream->Write(init_msg)) {
+        const auto status = stream->Finish();
```
acmorrow: claude: [change] - It looks to me like the `Finish` call should be handled by an RAII guard.

src/viam/sdk/components/private/arm_client.cpp
```diff
@@ -104,6 +111,119 @@ void ArmClient::move_through_joint_positions(const std::vector<std::vector<doubl
...
+    // If update_handler returns false the caller is electing to halt; record it
+    // so teardown reports k_halted_by_update_handler rather than the CANCELLED
+    // status our own TryCancel will produce.
+    bool update_handler_halted = false;
+    std::exception_ptr reader_exception;
+    std::thread reader([&] {
+        try {
+            ::viam::component::arm::v1::MoveThroughJointPositionsStreamedResponse pb;
+            while (stream->Read(&pb)) {
+                if (!update_handler(Arm::trajectory_update{})) {
```
acmorrow: claude: [change] - even though `trajectory_update` is vacuous right now, it may not always be. We should do the work to have proto conversion for it, and feed the actual data that came back in the response pb to it. That way, no changes here will be required if/when the response actually carries data.

src/viam/sdk/components/private/arm_client.cpp
```diff
@@ -104,6 +111,119 @@ void ArmClient::move_through_joint_positions(const std::vector<std::vector<doubl
...
+            auto* trajectory_batch = msg.mutable_batch();
+            for (const auto& point : *batch) {
+                if (const auto err = validator.check(point)) {
+                    throw Exception(ErrorCondition::k_general,
+                                    "move_through_joint_positions_streamed: " + *err);
+                }
+                *trajectory_batch->add_points() = to_proto(point);
+            }
+            if (!stream->Write(msg)) {
+                break;
```
acmorrow: claude: [question] - what does it mean if `Write` fails? Why is `break` correct? Is there no error we should be propagating in such an event?

src/viam/sdk/components/private/arm_client.cpp
```diff
@@ -104,6 +111,119 @@ void ArmClient::move_through_joint_positions(const std::vector<std::vector<doubl
...
+    // Reap the call on every exit path, even the ones that TryCancel'd and are
+    // about to rethrow. Finish() completes the cancelled RPC and releases its
+    // resources; skipping it (as the error paths previously did) leaks an
+    // in-flight call per errored stream. It is safe to call now: the reader has
+    // been joined, so no other thread touches the stream.
+    const auto status = stream->Finish();
+
+    // Prefer a callback exception over Finish()'s gRPC status: after a
+    // TryCancel that status is a generic CANCELLED that hides the real cause,
+    // whereas the stashed exception carries it.
```
acmorrow: claude: [nit] - this comment should maybe speak to why a preference for the writer exception over the reader one, as well, if it matters. Or state that it doesn't matter and the choice is arbitrary.

src/viam/sdk/components/private/arm_server.cpp
```diff
@@ -92,6 +98,122 @@ ::grpc::Status ArmServer::MoveThroughJointPositions(
...
 }
 
+::grpc::Status ArmServer::MoveThroughJointPositionsStreamed(
+    ::grpc::ServerContext* context,
+    ::grpc::ServerReaderWriter<
+        ::viam::component::arm::v1::MoveThroughJointPositionsStreamedResponse,
+        ::viam::component::arm::v1::MoveThroughJointPositionsStreamedRequest>* stream) noexcept {
+    using RequestT = ::viam::component::arm::v1::MoveThroughJointPositionsStreamedRequest;
+    using ResponseT = ::viam::component::arm::v1::MoveThroughJointPositionsStreamedResponse;
+
```
acmorrow: claude: [change] - This implementation should also have a TODO referencing the bidi client/server helper support.

src/viam/sdk/components/private/arm_server.cpp
```diff
@@ -92,6 +98,122 @@ ::grpc::Status ArmServer::MoveThroughJointPositions(
...
+    ServerSpanGuard span_guard{context, "ArmServer::MoveThroughJointPositionsStreamed"};
+
+    // 1. Read the Init message. The stream must start with exactly one Init.
```
acmorrow: claude: [nit] - don't number stages like this. They get out of sync. Just describe what is being done. The code is the ordering.

src/viam/sdk/components/private/arm_server.cpp
```diff
@@ -92,6 +98,122 @@ ::grpc::Status ArmServer::MoveThroughJointPositions(
         });
 }
 
+::grpc::Status ArmServer::MoveThroughJointPositionsStreamed(
+    ::grpc::ServerContext* context,
+    ::grpc::ServerReaderWriter<
+        ::viam::component::arm::v1::MoveThroughJointPositionsStreamedResponse,
+        ::viam::component::arm::v1::MoveThroughJointPositionsStreamedRequest>* stream) noexcept {
+    using RequestT = ::viam::component::arm::v1::MoveThroughJointPositionsStreamedRequest;
```
acmorrow: claude: [nit]: I'd prefer `request_type` and `response_type` for these.

src/viam/sdk/components/private/arm_server.cpp
```diff
@@ -92,6 +98,122 @@ ::grpc::Status ArmServer::MoveThroughJointPositions(
...
+    const auto& init = first.init();
+
+    // Everything downstream of the Init read may throw (proto conversions,
+    // resource lookup, the impl itself, and the batch_source closure which
+    // throws grpc::Status on protocol violations). We are noexcept, so wrap
+    // it all in a single try and catch every throw type.
+    try {
+        const ProtoStruct extra = from_proto(init.extra());
+
+        // 2. Resource lookup by name.
```
acmorrow: claude: [question] - I wonder if maybe the lookup by name should come even earlier. Also, check the server helper to see what other resource validation checks are done (e.g. name is just missing, rather than we don't find by that name).

src/viam/sdk/components/private/arm_server.cpp
```diff
@@ -92,6 +98,122 @@ ::grpc::Status ArmServer::MoveThroughJointPositions(
...
+            return span_guard.commit(::grpc::Status(
+                ::grpc::StatusCode::NOT_FOUND,
+                "MoveThroughJointPositionsStreamed: arm not found: " + first.name()));
+        }
+
+        // 3. batch_source: pull the next batch off the stream. Empty batches
+        // are tolerated as wire no-ops (the impl never sees them). A second
+        // Init or a missing oneof is a protocol violation surfaced via a
+        // thrown grpc::Status so the client gets a clear INVALID_ARGUMENT. Each
+        // point is validated against the trajectory invariants as it arrives;
```
acmorrow: claude: [question] - Are we doing too much validation by validating on both the RPC client and the RPC server? Clients can come from anywhere, and the server shouldn't be relying on it anyway. Maybe clients shouldn't validate, and only servers should?

src/viam/sdk/components/private/arm_server.cpp
```diff
@@ -92,6 +98,122 @@ ::grpc::Status ArmServer::MoveThroughJointPositions(
...
+        auto batch_source = [stream, context, validator = TrajectoryStreamValidator{}]() mutable
+            -> boost::optional<std::vector<Arm::trajectory_point>> {
+            while (true) {
+                if (context->IsCancelled()) {
+                    return boost::none;
+                }
+                RequestT msg;
+                if (!stream->Read(&msg)) {
+                    return boost::none;
+                }
```
acmorrow, line 183: claude: [question] - why is returning boost::none in both of these cases correct? Also they become indistinguishable. Is that acceptable? What does it mean exactly if `Read` returns false?

src/viam/sdk/components/private/arm_server.cpp
```diff
@@ -92,6 +98,122 @@ ::grpc::Status ArmServer::MoveThroughJointPositions(
...
+        };
+
+        // 4. update_handler: write one update on the wire. Returns false on
+        // cancellation or wire failure so the impl knows to stop.
+        auto update_handler = [stream, context](Arm::trajectory_update) -> bool {
+            if (context->IsCancelled()) {
+                return false;
+            }
+            ResponseT pb;
+            return stream->Write(pb);
```
acmorrow: claude: [change] - Similar to with the client, we should be respecting the returned value and serializing it onto the wire, even if it is currently empty. Otherwise we need to change this code if/when the response contents become meaningful.

src/viam/sdk/components/arm.hpp
```diff
@@ -105,6 +121,137 @@ class Arm : public Component, public Stoppable {
...
+    /// @return How the stream ended.
+    inline stream_outcome move_through_joint_positions_streamed(
+        const std::function<boost::optional<std::vector<trajectory_point>>()>& batch_source,
+        const std::function<bool(trajectory_update)>& update_handler) {
+        return move_through_joint_positions_streamed(batch_source, update_handler, {});
+    }
+
+    /// @brief Execute a stream of trajectory points in order.
+    ///
+    /// `batch_source` returns the next batch of points or `boost::none` on
```
acmorrow: claude: [change] - If these are parameter docs, they should be doxygen parameter docs.

src/viam/sdk/components/arm.hpp
```diff
@@ -105,6 +121,137 @@ class Arm : public Component, public Stoppable {
...
+    /// client EOF or cancellation. Empty batches and protocol-noise messages
```
acmorrow: claude: [change] - The comment on `batch_source` isn't great. This is the abstract method, so claims about implementation are fiction. The comment should tell the user how to use the parameter.

src/viam/sdk/components/arm.hpp
```diff
@@ -105,6 +121,137 @@ class Arm : public Component, public Stoppable {
...
+    /// are filtered out by the dispatcher and never surface here.
+    ///
+    /// `update_handler` consumes each `trajectory_update` the implementation
```
acmorrow: claude: [change] - You don't need to capitalize client and server. And you might want to change the language to say "client implementation" or "server implementation". Also saying "it is" isn't appropriate. Again, this is an abstract interface, so the comment should speak to the contract to which implementations must adhere, not any specific implementation details.

src/viam/sdk/components/arm.hpp
```diff
@@ -105,6 +121,137 @@ class Arm : public Component, public Stoppable {
...
+    /// early. The `false` return is bivalent by side: on the CLIENT it is the
+    /// caller electing to halt (perhaps reacting to what it sees coming back),
+    /// which ends the call with `stream_outcome::k_halted_by_update_handler`;
+    /// on the SERVER it is the framework telling the implementation to stop
+    /// (cancellation, wire failure). If `update_handler` throws, the exception
+    /// is not swallowed -- it propagates out of the call on the client, and
+    /// converts to a terminal gRPC status on the server. It need not be
+    /// `noexcept`.
+    ///
+    /// Both callbacks are borrowed only for the duration of the call: the
```
acmorrow: claude: [change] - This comment about the callbacks is implied by the type system and doesn't add anything.

src/viam/sdk/components/arm.hpp
```diff
@@ -105,6 +121,137 @@ class Arm : public Component, public Stoppable {
...
+    /// implementation invokes them but must not retain them past its return.
+    ///
+    /// Returns `stream_outcome::k_completed` when the trajectory ran to its
```
acmorrow: claude: [change] - As with the parameters, this should be a doxygen comment explaining the return.

src/viam/sdk/components/arm.hpp
```diff
@@ -105,6 +121,137 @@ class Arm : public Component, public Stoppable {
...
+    ///   a cadence independent of the inbound batches. See ERROR SURFACING
+    ///   below for the narrow consequence today and why the obvious fix is not
+    ///   trivial.
+    /// - On the CLIENT (i.e., when invoked through ArmClient on a remote
+    ///   arm), `batch_source` runs on the caller's thread and
+    ///   `update_handler` runs on a separate SDK-managed reader thread. The
+    ///   two callbacks may execute concurrently. Each is called serially
+    ///   with respect to itself (no concurrent calls to the same callback).
+    ///   Implementations sharing mutable state between the two callbacks
+    ///   must provide their own synchronization.
```
acmorrow: claude: [change] - This whole section speaks waaaay too much to the implementation, and not to the contract. This section on threading should instead be framed around what implementations are permitted to do. In particular, that due to the BiDi nature, implementations may use threads, callbacks are not assured to be serialized on the same thread, etc.

src/viam/sdk/components/arm.hpp
```diff
@@ -105,6 +121,137 @@ class Arm : public Component, public Stoppable {
...
+    /// the case that genuinely needs concurrent emit, and it stays gated behind
+    /// `trajectory_update` growing fields.
+    ///
+    /// The obvious fix -- a background reader thread feeding a queue -- is not
+    /// trivial: a thread parked in `stream->Read` only unblocks once the RPC
+    /// finishes, which needs the handler to return, so it cannot be joined
+    /// before returning (the deadlock rdk's recv path hit). `TryCancel` unblocks
+    /// the Read but flattens the client-visible status to CANCELLED, losing the
+    /// specific fault code. A proper fix points at the async CompletionQueue
+    /// gRPC API -- a large change, disproportionate for this landing.
```
acmorrow: claude: [change] - Again, this whole comment block is mis-framed. You need to describe the interaction design, yes, but it should do so in a way to guide implementors of the client and server sides per what they may and may not do, and not the specifics of any particular implementation (like our current client/server stubs).

src/viam/sdk/tests/test_arm.cpp
```diff
@@ -44,6 +53,114 @@ struct CheckVector : boost::static_visitor<> {
...
+            return boost::none;
+        }
+        return (*batches)[(*index)++];
+    };
+}
+
+using Batches = std::vector<std::vector<Arm::trajectory_point>>;
+
+using RawRequest = ::viam::component::arm::v1::MoveThroughJointPositionsStreamedRequest;
+using RawResponse = ::viam::component::arm::v1::MoveThroughJointPositionsStreamedResponse;
```
acmorrow: claude: [change] - snake case.

src/viam/sdk/tests/test_arm.cpp
```diff
@@ -229,6 +346,245 @@ BOOST_AUTO_TEST_CASE(test_get_3d_models) {
...
+        check_points_equal(mock->peek_streamed_batches[0][0], (*batches)[0][0]);
+        check_points_equal(mock->peek_streamed_batches[1][1], (*batches)[1][1]);
+    });
+}
+
+BOOST_AUTO_TEST_CASE(streamed_empty_batches_filtered) {
+    auto mock = MockArm::get_mock_arm();
+    client_to_mock_pipeline<Arm>(mock, [&](Arm& client) {
+        // The empty middle batch is dropped on the wire and never reaches the
+        // impl; the surrounding points still form a valid monotonic stream.
```
acmorrow: claude: [nit] - This seems like not an important thing to test. It isn't required behavior.

src/viam/sdk/components/private/arm_client.cpp
```diff
@@ -104,6 +111,125 @@ void ArmClient::move_through_joint_positions(const std::vector<std::vector<doubl
...
+
+    {
+        const finish_guard guard{stream.get(), finish_status};
+
+        // Send Init on the caller's thread, before the reader starts.
+        ::viam::component::arm::v1::MoveThroughJointPositionsStreamedRequest init_msg;
+        init_msg.set_name(this->name());
+        *init_msg.mutable_init()->mutable_extra() = to_proto(extra);
+        if (!stream->Write(init_msg)) {
+            init_write_failed = true;
```
acmorrow: claude: why isn't this just an immediate throw?

src/viam/sdk/components/private/arm_server.cpp
```diff
@@ -92,6 +98,151 @@ ::grpc::Status ArmServer::MoveThroughJointPositions(
...
+    using response_type = ::viam::component::arm::v1::MoveThroughJointPositionsStreamedResponse;
+
+    // TODO(RSDK-14164): like ArmClient, this hand-rolls the bidi stream, pending
+    // a shared bidi client/server helper (RSDK-14164).
+    //
+    // The point of the bidi stream is that an arm-side fault can reach the client
+    // mid-execution rather than only at the end. This dispatcher runs the
+    // implementation on the gRPC handler thread itself, with batch_source
+    // blocking in stream->Read and update_handler in stream->Write, and no
+    // background reader. That sidesteps the recv-thread deadlock rdk's bidi
```
acmorrow: claude: shouldn't talk about what happened in RDK here.

src/viam/sdk/components/private/arm_server.cpp
```diff
@@ -92,6 +98,151 @@ ::grpc::Status ArmServer::MoveThroughJointPositions(
...
+    // server hit, but it costs us something specific to synchronous grpc++: while
+    // the implementation is parked in a batch_source Read it can't send an
+    // update, so a fault reaches the client only on the next inbound batch, or
+    // promptly once the client half-closes its send side (a client that stops
+    // sending without half-closing, then sits idle, can hold onto an unreported
+    // fault). The fix is the async CompletionQueue API (try_cancel would unblock
+    // a background Read, but it replaces the real status with CANCELLED), and it
+    // waits until trajectory_update carries data that has to flow while the
+    // client is still sending.
```
acmorrow: claude: just say that the correct fix would require an async C++ SDK.

src/viam/sdk/components/private/arm_server.cpp
```diff
@@ -92,6 +98,151 @@ ::grpc::Status ArmServer::MoveThroughJointPositions(
...
+                batch.reserve(msg.batch().points_size());
+                for (const auto& pb_point : msg.batch().points()) {
+                    auto point = from_proto(pb_point);
+                    if (const auto err = validator.check(point)) {
+                        throw ::grpc::Status(::grpc::StatusCode::INVALID_ARGUMENT,
+                                             "MoveThroughJointPositionsStreamed: " + *err);
+                    }
+                    batch.push_back(std::move(point));
+                }
+                if (batch.empty()) {
```
acmorrow: claude: or just `if (!batch.empty()) return;`

src/viam/sdk/components/private/arm_server.cpp
```diff
@@ -92,6 +98,151 @@ ::grpc::Status ArmServer::MoveThroughJointPositions(
...
+                }
+                return batch;
+            }
+        };
+
+        // update_handler serializes one update onto the wire. It converts the
+        // update through to_proto so the wire payload tracks the type as it
+        // grows fields. Returns false on cancellation or wire failure so the
+        // impl knows to stop.
+        auto update_handler = [stream, context](Arm::trajectory_update update) -> bool {
```
acmorrow: claude: do you need the trailing return on the lambda

src/viam/sdk/components/private/arm_server.cpp
```diff
@@ -92,6 +98,151 @@ ::grpc::Status ArmServer::MoveThroughJointPositions(
...
+            if (context->IsCancelled()) {
+                return false;
+            }
+            const response_type pb = to_proto(update);
```
acmorrow: claude: you don't really need `pb`, just `return stream->Write(to_proto(update))` is fine.

src/viam/sdk/components/private/arm_server.cpp
```diff
@@ -92,6 +98,151 @@ ::grpc::Status ArmServer::MoveThroughJointPositions(
...
+                if (context->IsCancelled()) {
+                    return boost::none;
+                }
+                request_type msg;
+                if (!stream->Read(&msg)) {
+                    return boost::none;
+                }
+                const auto mc = msg.message_case();
+                if (mc == request_type::kInit) {
+                    throw ::grpc::Status(::grpc::StatusCode::INVALID_ARGUMENT,
```
acmorrow, line 187: claude: there's nothing wrong, exactly, with throwing a grpc::Status, but it is a little unusual. Are there any alternatives? Do we not, in the C++ SDK, have an exception which carries a grpc::Status already? Does gRPC not have one?

acmorrow:
@lia-viam - This is now ready for review. It is blocked on https://github.com/viamrobotics/api/pull/874 so it won't go green until that lands. Still, I'd like to have it approved and ready to go so as soon as the API changes land, this can merge.

src/viam/sdk/components/arm.hpp
```diff
@@ -105,6 +122,100 @@ class Arm : public Component, public Stoppable {
                                               const MoveOptions& options,
                                               const ProtoStruct& extra) = 0;
 
+    /// @brief A single waypoint in a streamed joint-space trajectory.
+    ///
+    /// The first point of a stream must have @ref time equal to zero; subsequent
+    /// points must be strictly increasing in time across the entire stream.
+    struct trajectory_point {
```
acmorrow, line 129: This should probably document that the API here is degrees.

src/viam/sdk/components/private/arm_server.cpp
```diff
@@ -92,6 +98,150 @@ ::grpc::Status ArmServer::MoveThroughJointPositions(
...
+        // for the non-streaming methods.
+        const GrpcContextObserver::Enable observer_enable{*context};
+
+        // Call the virtual. Its `stream_outcome` is meaningful only to a direct
+        // caller of the impl; over the wire the client reconstructs its own
+        // outcome, so the dispatcher intentionally discards it here.
+        arm->move_through_joint_positions_streamed(batch_source, update_handler, extra);
+    } catch (const ::grpc::Status& s) {
+        return span_guard.commit(s);
+    } catch (const std::exception& e) {
```
acmorrow, line 239: Arguably, we could have special handling for some exceptions, like std::invalid_argument could be mapped to `INVALID_ARGUMENT`. Maybe RSDK-14164 is the right place to do that, or maybe it should just get done explicitly here, and in the service_helper itself now as well, so when we do it for BiDi it comes along for the ride automatically.

src/viam/sdk/components/private/arm_trajectory_validation.hpp
```diff
@@ -0,0 +1,51 @@
...
+//
+// `check()` reports the offending invariant as a string instead of throwing, and
+// the class pulls in no gRPC types. The only caller today is the server
+// dispatcher, which turns that string into a `grpc::Status` (so it could just as
+// well throw), but keeping it free of gRPC means that if we ever want the client
+// to validate before sending, this drops straight in and raises whatever error
+// the client wants from the same message.
+//
+// Invariants:
+//   - the first point of the stream has time zero;
```
acmorrow, line 30: I think this should also validate that if velocities are provided, that the velocities are identically zero at the t=0 sample, since the arm begins at rest.

acmorrow:
Hang back on this one for the moment, there is an update coming due to some small changes in the api repo.

src/viam/sdk/components/arm.hpp
```diff
@@ -105,6 +122,103 @@ class Arm : public Component, public Stoppable {
...
+        std::vector<double> positions;
+
+        /// @brief Optional velocity and acceleration constraints at this waypoint.
+        boost::optional<kinematic_constraints> constraints;
+    };
+
+    /// @brief An update emitted while a streamed trajectory executes.
+    ///
+    /// An arm may emit them at any cadence.
+    struct trajectory_update {};
```
acmorrow, line 160: Note that this type is slightly a lie: the real API `Response` type for streaming has a `oneof message` as the only field. But that `oneof` currently contains only one possible result: a `BatchAck`, which is itself empty. All of this is in furtherance of not breaking compatibility for old clients if and when additional response data types are added, which is anticipated. For now, it seems reasonable to keep the type silent, since the existence of a reply on the stream is implicitly a batch ack (it can't be anything else on the wire).

acmorrow:
> Hang back on this one for the moment, there is an update coming due to some small changes in the api repo.

OK, good to go on this now.

src/viam/sdk/components/arm.hpp
```diff
@@ -105,6 +122,103 @@ class Arm : public Component, public Stoppable {
...
+    struct trajectory_point {
+        /// @brief Target kinematic constraints at a waypoint.
+        ///
+        /// `velocities` is required whenever this struct is present;
+        /// `accelerations` is independently optional. Both are one value per
+        /// joint and, like everything else on this interface, denominated in
+        /// degrees: `velocities` in degrees per second and `accelerations` in
+        /// degrees per second squared.
+        struct kinematic_constraints {
+            std::vector<double> velocities;
```
lia-viam: I don't love putting type info in decls but maybe we should do so here to conform with other parts of the codebase using `_degs` etc suffixes

lia-viam (approved):
minor nit but lgtm, cool stuff here! test failures i think will be fixed by a missing include or two in arm.cpp

src/viam/sdk/components/arm.hpp
```diff
@@ -105,6 +122,103 @@ class Arm : public Component, public Stoppable {
...
+    /// permitted to invoke the two callbacks from different threads and to run
+    /// them concurrently with each other. A caller must therefore not assume the
+    /// callbacks are serialized against one another or pinned to a single
+    /// thread, and must synchronize any mutable state it shares between them. A
+    /// given callback is never invoked concurrently with itself.
+    ///
+    /// @param batch_source Pull-source for the trajectory. Call it to obtain the
+    /// next batch of waypoints; it yields `boost::none` once no more points are
+    /// coming, after which the implementation should finish executing what it
+    /// already has and return. Points arrive grouped into the batches the sender
```
acmorrow, line 205: This comment on `batch_source` should be augmented to note that disengaged return cannot distinguish EOS from cancellation, and that server side implementations are obligated to explicitly check for cancellation.

src/viam/sdk/components/private/arm_server.cpp
acmorrow: Done, implemented special handling for invalid argument.

src/viam/sdk/components/private/arm_trajectory_validation.hpp
acmorrow: Done

src/viam/sdk/components/arm.hpp
acmorrow: Done

src/viam/sdk/components/arm.hpp
acmorrow: Agree, I don't love doing that, but we have nothing better. Hopefully someday we can do this SDK right, with a units library.

Done.

PR #681, opened 2026-07-10
Add armhf to the docker matrix

acmorrow:
There is no armhf LLVM on the apt repos for it, so if we hit the fallback, just don't install there. Probably short-lived if we stop trying to pin the LLVM version to 19.


acmorrow:
Workflow run: https://github.com/viamrobotics/viam-cpp-sdk/actions/runs/29111272783. Sid is red across the board, so that is independent of my change.

lia-viam (approved):

PR #688, opened 2026-07-21
RSDK-14123: Docker conan image

lia-viam:
Last major step of the current docker CI work, adding a conan docker image

One thing of note is that conan.yml CI is a fair bit slower now, but that's because we are doing RelWithDebInfo builds which is expected.

Left unchanged, conan-publish.yml would be doing RelWithDebInfo deps + viam-cpp-sdk as well. I believe this was a thing we wanted to be able to do, along with sanitizer builds down the line, but I think it makes the most sense to have the release packages still be release builds. 

See https://github.com/viamrobotics/viam-cpp-sdk/actions for a bunch of test runs on the PR source branch 

The conan smoke step from the scope doc is omitted but could be added to this PR if we want


.github/workflows/conan.yml
```diff
@@ -89,53 +90,30 @@ jobs:
...
       - name: Create package
-        shell: bash
+        shell: bash -leo pipefail {0}
         run: |
-          conan profile detect
+          conan remote add viamconan "$CONAN_REMOTE_URL"
+          conan remote auth --force -cc core:non_interactive=True viamconan
           conan create . --build=missing -s:a compiler.cppstd=17
+        env:
+          CONAN_REMOTE_URL: ${{ secrets.conanRemote }}
```
acmorrow: Is the URL actually a secret?

.github/workflows/conan.yml
```diff
@@ -89,53 +90,30 @@ jobs:
...
-
+      # Custom shell replaces GH's default `bash --noprofile --norc -eo pipefail {0}`.
+      # -l (login) sources /etc/profile.d/conan-venv.sh, putting the image's conan on
+      # PATH; we re-add -eo pipefail since our string overrides the default's flags.
       - name: Create package
-        shell: bash
+        shell: bash -leo pipefail {0}
         run: |
-          conan profile detect
+          conan remote add viamconan "$CONAN_REMOTE_URL"
```
acmorrow: I had envisioned this `remote add` step, in particular, as being baked into the container, so that users of the container are cache enabled without needing any further action.

If further action is required (like `conan remote auth`), I wonder if that should exist as a re-usable action.

I'd expect there is a way to still inject the secrets as env vars into the container start.

etc/docker/scripts/install-conan.sh
```diff
@@ -0,0 +1,34 @@
...
+python3 -m venv /opt/conan-venv
+# Upgrade pip first: the pip bundled on older distros (e.g. bullseye's 20.3) ships
+# the legacy resolver, which crashes resolving conan's deps (KeyError: 'markupsafe').
+/opt/conan-venv/bin/pip install --no-cache-dir --upgrade pip
+/opt/conan-venv/bin/pip install --no-cache-dir "conan~=2.0"
+
+# Source the venv activate for conan on PATH; profile.d carries it to runtime shells.
+echo '. /opt/conan-venv/bin/activate' > /etc/profile.d/conan-venv.sh
+# shellcheck disable=SC1091
+. /opt/conan-venv/bin/activate
```
acmorrow: This seems like a lot of complexity to work around `pip install --break-system-packages conan`. That's honestly pretty fair to do in a container build. The container is the isolation, we aren't risking breaking things. Or, we could consider `pipx`?

etc/docker/scripts/conan/default.jinja
```diff
@@ -0,0 +1,13 @@
...
+{# (gcc / libstdc++11 / C++17).                                                 #}
+{% set _c = detect_api.detect_default_compiler() %}
+[settings]
+os=Linux
+arch={{ detect_api.detect_arch() }}
+build_type=RelWithDebInfo
+compiler=gcc
+compiler.version={{ detect_api.default_compiler_version(_c[0], _c[1]) }}
+compiler.cppstd=17
+compiler.libcxx=libstdc++11
```
acmorrow: What does the 11 mean here on libcxx?

etc/docker/scripts/conan/default.jinja
```diff
@@ -0,0 +1,13 @@
...
+{# installed gcc version; the rest is the image's pinned toolchain of record    #}
+{# (gcc / libstdc++11 / C++17).                                                 #}
+{% set _c = detect_api.detect_default_compiler() %}
+[settings]
+os=Linux
+arch={{ detect_api.detect_arch() }}
+build_type=RelWithDebInfo
+compiler=gcc
+compiler.version={{ detect_api.default_compiler_version(_c[0], _c[1]) }}
+compiler.cppstd=17
```
acmorrow: Shouldn't this be cppstd=14, for now?

etc/docker/scripts/conan/default.jinja
```diff
@@ -0,0 +1,13 @@
+{# Single default conan profile for the cpp-sdk-conan images, rendered per      #}
```
acmorrow: I had no idea these were templates. How cool.

I think we should give these profiles real names though. I think I would call this one something like `viam-system-gcc-relwithdebinfo`. I think we will end up with a fair number of them.

Then we can set the default profile name to it with `conan config`. Or, maybe even not, so that you need to be explicit about the profile you are requesting.

Also, just noting that profiles support inclusion, so I think it is possible that we start building up "profile layers".

etc/docker/Dockerfile
```diff
@@ -45,3 +46,15 @@ ENV GRPC_SOURCE_VERSION_OVERRIDE=${GRPC_SOURCE_VERSION_OVERRIDE}
...
 
 RUN rm -rf /opt/scripts
+
+
+FROM base AS conan
+
+# Pin CONAN_HOME so the baked default profile is found regardless of runtime HOME.
+# GitHub Actions `container:` overrides HOME to /github/home, which would otherwise
+# send conan looking for its home (and the default profile) in the wrong place.
+ENV CONAN_HOME=/opt/conan-home
```
acmorrow: Or just `CONAN_HOME=/home/conan`? Though, that may be an appealing idea with confusing consequences if there is not a `conan` user.

I'm not sure `/opt` is the right home. Maybe `/usr/local/etc/conan`?

etc/docker/scripts/conan/default.jinja
```diff
@@ -0,0 +1,13 @@
...
+{# machine at conan load time via jinja2. detect_api fills arch and the         #}
+{# installed gcc version; the rest is the image's pinned toolchain of record    #}
+{# (gcc / libstdc++11 / C++17).                                                 #}
+{% set _c = detect_api.detect_default_compiler() %}
+[settings]
+os=Linux
+arch={{ detect_api.detect_arch() }}
+build_type=RelWithDebInfo
```
acmorrow: Defaulting to RelWithDebInfo is going to lead to pretty huge builds: it will mean all of boost, grpc, proto, etc. have symbols. For things like UR and other conan-driven modules, I arranged things so that only the viam pieces ended up with RelWithDebInfo. That way we could at least symbolize our own frames, if not the "system" ones.

.github/workflows/conan-publish.yml
```diff
@@ -179,23 +127,21 @@ jobs:
...
-        --lockfile etc/conan/conan.lock $LOCKFILE_UPDATE_ARGS
-
-        conan upload "viam-cpp-sdk/*" -r viamconan -c
+        # TEMP(test): skip lockfile pin + upload; just build the branch to smoke-test the image.
+        # Pin Release; the image profile defaults to RelWithDebInfo, but releases stay Release.
+        conan create . --build=missing -s:a compiler.cppstd=17 -s:a build_type=Release
+        # conan lock add --requires=viam-cpp-sdk/${{ inputs.ref }}
+        # conan create . --build=missing -s:a compiler.cppstd=17 -s:a build_type=Release \
+        # --lockfile etc/conan/conan.lock $LOCKFILE_UPDATE_ARGS
+        # conan upload "viam-cpp-sdk/*" -r viamconan -c
```
acmorrow: It hadn't registered with me, but of course it makes sense, that you need to explicitly `conan upload` the thing you want to send to the cache. 

We will need to think about how and when that should happen. How will a package using these containers know what to upload to the cache? It should look something like "the things I didn't fetch from the cache, the package I built, and its transitive deps that I did build". I don't know how to express that.

acmorrow (changes requested):

etc/docker/Dockerfile
acmorrow: Wait, even simpler: why not just `$HOME/.conan2` or whatever the default would be? That way we aren't polluting `opt`, and if we ever decided we wanted a real user (rather than `root`) for these containers, it'd just work.

.github/workflows/conan-publish.yml
lia-viam: yeah, i think that is a broader conan usage question we will need to answer. the current approach is to just heuristically choose some slow deps

.github/workflows/conan.yml
lia-viam: we can definitely add that in the container, was sort of going off stuff in the original google doc which seemed to suggest that it was going to happen in build jobs on top of the container rather than in the container itself

.github/workflows/conan.yml
lia-viam: no haha i should change it to a var rather than a secret...it was a secret for a week or so after we signed up for it

etc/docker/scripts/conan/default.jinja
lia-viam: ah in that case i'll make it use Release instead

etc/docker/scripts/conan/default.jinja
lia-viam: we actually just bumped this to 17 because of the xtensor thing i think, c++14 support is an increasingly small increasingly theoretical slice at this point https://github.com/viamrobotics/viam-cpp-sdk/pull/667

etc/docker/scripts/conan/default.jinja
lia-viam: this explicitly requests the LIBCXX C++11 ABI, which some google suggests is default on non-prehistoric gcc/cppstd versions https://www.reddit.com/r/cpp/comments/hr8g9k/comment/fycy8f2/

etc/docker/scripts/install-conan.sh
lia-viam: i'll look into these, i also really don't like creating a container where you have to `source` something before every step in a CI job

etc/docker/Dockerfile
lia-viam: i like this

etc/docker/scripts/install-conan.sh
acmorrow: Oh! They make `.deb` files: https://conan.io/downloads

acmorrow: So close: sadly, they don't make an armhf package

.github/workflows/conan.yml
acmorrow: I think having the viamcache built in and active is the best plan if we can make it happen. That way, users don't need to remember to do it. Things "just work".

Meanwhile: is doing `conan remote auth` required to read from the cache? Or only for writes?

.github/workflows/conan.yml
acmorrow: I think it goes away entirely if we are baking the addition of the viamconan repo into the container. Maybe it becomes something we pull from the environment on container build time though.

etc/docker/scripts/conan/default.jinja
acmorrow: I think that is best to start. The build of the C++ SDK *itself* can opt for RelWithDebInfo, and push to the cache like that.

Eventually, we probably will have versions that build with symbols, for ASAN / UBSAN. But I think starting with Release as the baseline is the right move for now, since that's what most consumers do.

etc/docker/scripts/conan/default.jinja
acmorrow: Ack. Also, we are looking good for a move to C++20, soon!

etc/docker/scripts/conan/default.jinja
```diff
@@ -0,0 +1,13 @@
+{# Single default conan profile for the cpp-sdk-conan images, rendered per      #}
+{# machine at conan load time via jinja2. detect_api fills arch and the         #}
+{# installed gcc version; the rest is the image's pinned toolchain of record    #}
+{# (gcc / libstdc++11 / C++17).                                                 #}
+{% set _c = detect_api.detect_default_compiler() %}
+[settings]
```
acmorrow: Do we need some sort of synthetic setting to disambiguate the cache per distro? For instance, both sid and resolute are on GCC 15. Do we want them to share artifacts? I'd argue no, but I don't see anything that would prevent it.

lia-viam: i think they would be shared yes, here is some discussion on that topic https://docs.conan.io/2/reference/extensions/binary_compatibility.html

.github/workflows/conan.yml
lia-viam: was indeed not required for reads; nice drive-by cleanup

.github/workflows/conan.yml
lia-viam: ended up moving this to the conan setup script

etc/docker/scripts/install-conan.sh
lia-viam: ugh this one was kind of a mess, tried to sketch a solution with claude and there was basically a perfect complement between platforms that have pipx and platforms that have the `break-system-packages` flag. what if we used the deb files for everything except armhf? might be lower surface area bifurcation of all the options

etc/docker/scripts/conan/default.jinja
lia-viam: i like this, let's go with being explicit rather than falling back to implicit default which in practice is often the non-stable output of `conan profile detect`

etc/docker/Dockerfile
lia-viam: some claude-spew about this attempted solution:

```
Two separate things going on.

  $HOME doesn't stay dynamic. Dockerfile ENV expands at build time. ENV CONAN_HOME=$HOME/.conan2 bakes the literal /root/.conan2
  into the image — same hardcoded pin as /opt/conan-home, just spelled indirectly. It does not track the runtime HOME, so it
  fixes nothing about the GHA HOME=/github/home problem the comment describes. The pin is the point; only the path differs.

  /root is the worse path. Verified in debian:bookworm:
  drwxr-xr-x 2 root root /opt
  drwx------ 2 root root /root
  /root is 0700. Any runtime uid that isn't root — GHA container.options: --user, self-hosted runners with non-root uid, docker 
  run -u — can't even traverse into it, so the baked default profile and the viamconan remote go invisible. /opt/conan-home
  inherits 0755 and stays readable by every uid.

  Net: $HOME/.conan2 is equivalent-but-obscure at best, and locks non-root consumers out at worst. Current pin is right.

  Real gap either way: /opt/conan-home is root-owned 0755, so non-root still can't write the conan cache there
```

etc/docker/scripts/install-conan.sh
lia-viam: further claude spew just for details' sake:

```
  armhf: pipx and --break-system-packages both fail

  Marker presence and flag support correlate exactly — the flag is supported precisely where it isn't needed, and absent
  everywhere PEP-668 doesn't apply:

  ┌────────────────┬────────────────────┬────────┬─────────────────────────┬─────────────┐
  │ distro (armhf) │ EXTERNALLY-MANAGED │  pip   │ --break-system-packages │  apt pipx   │
  ├────────────────┼────────────────────┼────────┼─────────────────────────┼─────────────┤
  │ bullseye       │ no                 │ 20.3.4 │ ❌ unsupported          │ unavailable │
  ├────────────────┼────────────────────┼────────┼─────────────────────────┼─────────────┤
  │ bookworm       │ YES                │ 23.0.1 │ ✅                      │ 1.1.0       │
  ├────────────────┼────────────────────┼────────┼─────────────────────────┼─────────────┤
  │ trixie         │ YES                │ 25.1.1 │ ✅                      │ 1.7.1       │
  ├────────────────┼────────────────────┼────────┼─────────────────────────┼─────────────┤
  │ focal          │ no                 │ 20.0.2 │ ❌ unsupported          │ 0.12.3.1    │
  ├────────────────┼────────────────────┼────────┼─────────────────────────┼─────────────┤
  │ jammy          │ no                 │ 22.0.2 │ ❌ unsupported          │ 1.0.0       │
  ├────────────────┼────────────────────┼────────┼─────────────────────────┼─────────────┤
  │ noble          │ YES                │ 24.0   │ ✅                      │ 1.4.3       │
  └────────────────┴────────────────────┴────────┴─────────────────────────┴─────────────┘

  - --break-system-packages needs a marker-conditional branch, adds python3-pip, and — the killer — bullseye still lands on pip
  20.3.4, so the pip install --upgrade pip legacy-resolver workaround (KeyError: 'markupsafe') survives anyway. Buys nothing,
  costs a conditional.
  - pipx has no candidate at all on bullseye, and installs into a venv internally — a dependency wrapping the thing we already
  do.

  Keep the venv for armhf. It sidesteps PEP-668 structurally rather than overriding it, and it's the only mechanism uniform
  across all six.
```

etc/docker/scripts/conan/viam-system-gcc-release.jinja
```diff
@@ -0,0 +1,17 @@
+{# The only conan profile in the cpp-sdk-conan images, rendered per machine at  #}
```
lia-viam: @acmorrow lost our previous comment thread in the rename, but i'm wondering if we should keep a default profile after all. i think when `conan create .` fails for someone using this container they'll probably just do `conan profile detect` to make it go away which would give them a gnu17 build with an unpinned build type, probably leading to cache misses. 

i think installing our own default with stronger constraints than the `conan profile detect` one may be an improvement over the status quo, and we can just use `include` to make `default` be this one

.github/workflows/conan-publish.yml
acmorrow: I think that is a very valuable starting point and we can iterate on exactly how uploads happen later.

etc/docker/scripts/conan/viam-system-gcc-release.jinja
acmorrow: Makes sense to me to create a `default` that includes the named one.

etc/docker/scripts/install-conan.sh
acmorrow: I think I have a workaround. Not sure how you will feel about it, but I think we can make the `pip` install always work without a conditional.

Just:
```
$ mkdir -p ~/.config/pip
$ cat > ~/.config/pip/pip.conf
[global]
break-system-packages = true
```

Then do `python3 -m pip install conan`. It'll work on platforms that both do and do not have support for the `--break-system-packages` flag.

Then delete the file.

etc/docker/Dockerfile
acmorrow: Yeah, sorry, I was too brief. I know $HOME doesn't stay dynamic. I really want this to go to _root_s home at build time.

I'm not convinced by the thing about permissions to root's home. If we ever did want this whole infra running as a non-root user, then this would follow that user to that users home, right?

Does `conan` have any guidance on how to do a multi-user install? The concept is probably alien to it, in which case, our docker container is gonna run as some user, and I don't see why it oughtn'd be root.

In which case that objection doesn't really hold.

etc/conan/conan.lock
```diff
@@ -5,7 +5,6 @@
         "xtl/0.8.0#45f1f2f2e5cbff154b59a57dbd6efb55%1747056138.62",
         "xtensor/0.26.0#929fed688e0b3e850106059bf3782e81%1755248340.339",
         "xsimd/13.2.0#b4819b06b51041b1cd32114700adbb59%1743412135.667",
-        "viam-cpp-sdk/0.35.0#ce07d27f591144d8416f0fe5e32a20e5%1778879440.086464",
```
acmorrow, line 8: Just curious, why does this come out of the lockfile?

acmorrow (commented):
I'm a little confused about what's going on in the workflows right now, should I just ignore it for the most part?

.github/workflows/conan-publish.yml
```diff
@@ -20,182 +20,125 @@ on:
...
-
-        pip install conan
-
-        conan profile detect
-        conan remote add viamconan $CONAN_REMOTE_URL
         conan remote auth --force -cc core:non_interactive=True viamconan
 
     - name: Pre-install b2 (bullseye)
-      if: matrix.image == 'debian:bullseye'
+      if: matrix.image == 'ghcr.io/viamrobotics/cpp-sdk-conan-debian:bullseye'
```
acmorrow, line 121: What is this working around? Can we add a comment?

.github/workflows/conan.yml
```diff
@@ -68,14 +68,12 @@ jobs:
...
       - name: Create package
         run: |
-          # `buf` tries to read a CLI config file that we don't actually use located at
-          # ~/.config/buf/config.yaml. We don't always have permission to access this
-          # directory in CI, so we set the `BUF_CONFIG_DIR` to some other value that we
-          # do have permissions for. See https://github.com/bufbuild/buf/issues/2698 for
-          # more details.
           export BUF_CONFIG_DIR=$(mktemp -d)
           conan profile detect
+          conan remote add viamconan "$CONAN_REMOTE_URL"
```
acmorrow, line 73: Is this still needed if it is in the container image?

.github/workflows/conan.yml
```diff
@@ -149,75 +118,35 @@ jobs:
...
         run: |
-          python3 -m venv ./conan_venv
-          source ./conan_venv/bin/activate
-
-          pip install conan
-          conan profile detect
-
           conan install -r conancenter --update \
-          --tool-requires=b2/5.3.1 --build=b2/5.3.1 -s:a compiler.cppstd=17
+            --tool-requires=b2/5.3.1 --build=b2/5.3.1 -pr:a viam-system-gcc-release
```
acmorrow: Same question here: why the special casing around b2? It probably warrants a comment at minimum.

etc/conan/conan.lock
lia-viam: this was a drive by change of something I noticed while working on another ticket, leftover from before we made it so that `etc/conan/conan.lock` does not include the sdk itself

etc/docker/scripts/install-conan.sh
lia-viam: i can live with that

.github/workflows/conan-publish.yml
```diff
@@ -20,182 +20,125 @@ on:
...
+  #
+  #     - name: Set up python/conan
+  #       env:
+  #         CONAN_REMOTE_URL: ${{ secrets.conanRemote }}
+  #         CONAN_LOGIN_USERNAME_VIAMCONAN: ${{ secrets.conanUsername}}
+  #         CONAN_PASSWORD_VIAMCONAN: ${{ secrets.conanPassword }}
+  #       run: |
+  #         conan profile detect
+  #         conan remote add viamconan $CONAN_REMOTE_URL
+  #         conan remote auth --force -cc core:non_interactive=True viamconan
```
acmorrow: Raises an interesting question. We will also want profile definitions to exist for macos (and, later windows too). For the container setup, the profiles are baked into the images, which is great.

But, for module authors who want to do windows / macos builds, there is going to be a lot of headache getting all the pieces set up right, that the containers do for you: creating the profiles, adding the remote, authing with the secrets, etc.

It sounds to me like that'd be a github action we would author. At which point I start to wonder about this whole scheme. If we need to write the action anyway to make windows and macos setup sane, does that undermine the argument for the conan containers?

lia-viam: yeah this was starting to give me pause seeing all the DRY savings for linux and the same stuff lingering on macos. I think a `setup-viam-conan` github action is pretty doable, which then just leaves the question of what use cases if any are addressed by a conan docker container

.github/workflows/conan.yml
lia-viam: ah this is macos

.github/workflows/conan-publish.yml
acmorrow: I think the big advantage would be for local dev.

acmorrow (approved):
LGTM. I still have some questions but I think there is more value in landing this and iterating from here.

PR #698, opened 2026-07-28
RSDK-14311: Parse hpp files in module generator

lia-viam:
companion PR to https://github.com/viamrobotics/rdk/pull/6280 which has more explanation, and is where most of the action is happening


acmorrow (approved):
LGTM

PR #702, opened 2026-08-14
RSDK-14424: fix macos image and add profile

lia-viam:


.github/workflows/conan-publish.yml
```diff
@@ -21,8 +21,9 @@ on:
 
 jobs:
   build_macos:
-    # Not sure if we actually need x64 macos-latest
-    runs-on: macos-latest
+    # If GitHub bumps this image's default Xcode past apple-clang 21, add a step
```
acmorrow: Can this comment say why we need to do that / what mandates that compiler version?

etc/conan/macos.profile
```diff
@@ -0,0 +1,15 @@
+# Toolchain-of-record for published darwin binaries. compiler.version and
+# os.version are part of the package_id; pin them so consumers get cache hits
+# regardless of which Xcode the runner image defaults to. Keep in sync with
+# the runner/Xcode pin in conan.yml and conan-publish.yml.
+include(default)
+
+[settings]
+os=Macos
+os.version=14.0
```
acmorrow, line 8: Can we include the codename on this as a comment (I believe it is `Sonoma`)

.github/workflows/conan-publish.yml
lia-viam: yeah this was a weird one where claude added a kind of opaque version check block to CI which i felt was probably not necessary, tried to add a comment recapping it

etc/conan/macos.profile
```diff
@@ -0,0 +1,16 @@
...
+# the runner/Xcode pin in conan.yml and conan-publish.yml.
+include(default)
+
+# macOS 14 Sonoma
+[settings]
+os=Macos
+os.version=14.0
+arch=armv8
+compiler=apple-clang
+compiler.version=21
```
acmorrow, line 11: Is this `compiler.version` a minimum, or an exact requirement? If it is a minimum, then maybe the comments above make sense, since the actual compiler might be newer. But if it is exact, and the build would fail if the 21 rev compiler wasn't there, then I don't understand the comments about asserting the match.

.github/workflows/conan-publish.yml
```diff
@@ -21,8 +21,11 @@ on:
...
+    # asserting `clang --version` matches the etc/conan/macos.profile pin. The
+    # profile pins compiler.version without verifying it, so a drifted image would
+    # silently publish binaries built with a newer clang under the pinned package_id.
```
acmorrow: Is it actually a problem in any way if we picked up a newer compiler on macos? We are pinning the deployment target, and macOS does great work maintaining ABI. I doubt there is any reason we can't mix binaries from different toolchain versions.

lia-viam: the more i look at this i think it's just claude misunderstanding how conan profiles work, i'm inclined to just delete the comments and leave as is. if you build with a profile that specifies a compiler.version i'm pretty sure the build will fail if it can't find a binary for which `c++ --version` or similar returns that version number

.github/workflows/conan.yml
```diff
@@ -45,15 +45,7 @@ jobs:
...
     needs: [prepare]
-    runs-on: macos-latest
-    strategy:
-      fail-fast: false
-      matrix:
-        include:
-          - target: aarch64-apple-darwin
-            platform: macosx_arm64
-          - target: x86_64-apple-darwin
-            platform: macosx_x86_64
```
acmorrow, line 56: Is dropping x86 macos deliberate?

acmorrow (approved):
LGTM.

I'm still interested in how we are going to do Windows and macOS conan support with caching and pre-configured profiles, but this isn't the review to solve it.

.github/workflows/conan.yml
lia-viam: yeah claude caught this, it turned out matrix.target and matrix.platform were never being used, and the host machine is an arm64 macos so we had two branches building arm64

